PR #1, opened 2018-06-28
(SERVER-2172) CLI tool to import PKI into CA

justinstoller:
This implements a small CLI tool for importing CRLs, certs, and private key into a CA.


justinstoller:
There's a few small code clean ups I see, some more specs to write, and a lot of git clean up to do. Otherwise I think it would be great to get a review from one or more folks in @puppetlabs/puppetserver-maintainers (iow, I think the code is close enough to warrant reviews).

Also, I let this get a bit out of hand, let me know if I should split this into multiple PRs before asking for review.

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,136 @@
...
+          opts.on('--help', 'This setup specific help output') do |help|
+            parsed['help'] = true
+          end
+          opts.on('--version', 'Output the version') do |v|
+            parsed['version'] = true
+          end
+          opts.on('--config CONF', 'Path to puppet.conf') do |conf|
+            parsed['config'] = conf
+          end
+          opts.on('--private-key KEY', 'Path to PEM encoded key') do |key|
```
amysahli: Defaults might be nice here (I know I'm curious where the private key is usually stored)

amysahli: Do we have some established way/format to document that these are all required?

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,136 @@
...
+            parsed['private-key'] = key
+          end
+          opts.on('--cert-bundle BUNDLE', 'Path to PEM encoded bundle') do |bundle|
+            parsed['cert-bundle'] = bundle
+          end
+          opts.on('--crl-chain [CHAIN]', 'Path to PEM encoded chain') do |chain|
```
amysahli: Is there a place we should say that this is not required and if not included will generate a crl?

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,136 @@
...
+
+        files = [bundle_path, key_path, chain_path, config_path].compact
+
+        errors = validate_file_paths(files)
+        return 1 if log_possible_errors(errors)
+
+        unless chain_path
+          @logger.err 'Warning:'
+          @logger.err '    No CRL chain given'
+          @logger.err '    Full CRL chain checking will not be possible'
```
amysahli: Should we tell them that we will generate a CRL for them based on the cert provided, or is this tmi?  Why wouldnt' full crl chain checking be possible?  It will, right?  But only after some certificates are revoked.

lib/puppetserver/ca/x509_loader.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+      end
+
+      def validate_crl_and_cert(crl, cert)
+        unless crl.issuer == cert.subject
+          @errors << 'Leaf CRL was not issued by leaf certificate'
+        end
+      end
+
+      def validate_full_chain(certs, crls)
```
amysahli: Would you mind adding a brief comment what this is doing?

lib/puppetserver/ca/x509_loader.rb
```diff
@@ -0,0 +1,123 @@
...
+        @crls = chain_path ? load_crls(chain_path) : []
+
+        validate(@certs, @key, @crls)
+      end
+
+      # Only do as much validation as is possible, assume whoever tried to
+      # load the objects wrote errors about any invalid ones, but that bundle
+      # and chain may be empty arrays and pkey may be nil.
+      def validate(bundle, pkey, chain)
+        if !chain.empty? && !bundle.empty?
```
amysahli: I never know if I should put this logic (like if nil ...don't call this) here or in the method itself.   Do you think its better to make the methods "nil safe" or to not call them if the arguments are nil?

lib/puppetserver/ca/setup_command.rb
justinstoller: The agent differentiates between "leaf" and "full" CRL checking, where with leaf it only checks that the most subordinate CA hasn't been revoked by its issuer and full checking checks that none of the issuers farther up the hierarchy have been revoked by their issuers.

The Puppet Server CA can generate its own CRL, but without being given CRLs for its issuer it can't, and the agents its distributes CRLs to won't, be able to check the full chain's CRLs.

lib/puppetserver/ca/setup_command.rb
justinstoller: I didn't see a convention aside from mentioning it in the description, or adding it to the usage banner.

lib/puppetserver/ca/setup_command.rb
justinstoller: I think between this and the last comment I should try to add a bit more description to the help output.

amysahli:
I'm not sure if we care about this case, but if I provide the paths as directories (instead of files) I get this error:
```
 puppetserver-ca setup --private-key=bin --cert-bundle=bin
Warning:
    No CRL chain given
    Full CRL chain checking will not be possible

Traceback (most recent call last):
	8: from /usr/local/bin/puppetserver-ca:23:in `<main>'
	7: from /usr/local/bin/puppetserver-ca:23:in `load'
	6: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/exe/puppetserver-ca:5:in `<top (required)>'
	5: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/cli.rb:39:in `run!'
	4: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/setup_command.rb:31:in `run!'
	3: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/setup_command.rb:31:in `new'
	2: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/x509_loader.rb:13:in `initialize'
	1: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/x509_loader.rb:40:in `load_certs'
/usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/x509_loader.rb:40:in `read': Is a directory @ io_fread - bin (Errno::EISDIR)
```

lib/puppetserver/ca/setup_command.rb
amysahli: Ok.  I was missing the "full" part of the message. Thanks.

lib/puppetserver/ca/x509_loader.rb
justinstoller: I went back and forth on that here myself. I think that having the caller do the nil-checks would essentially inline this method into the caller (the initialize).

The validation seemed sufficiently different to me to warrant being in a different method, but... I'm not sure they're really a different level of abstraction than the loading that currently happens in the initialize method....

amysahli:
Using a bundle with only 1 certificate:
```
puppetserver-ca setup --private-key=bin/privkey.pem --cert-bundle=bin/ca.pem
Warning:
    No CRL chain given
    Full CRL chain checking will not be possible

Error:
    Leaf certificate could not be validated
```
Tracking down error with pry:
```
certs.first
=> #<OpenSSL::X509::Certificate subject=#<OpenSSL::X509::Name CN=localhost>, issuer=#<OpenSSL::X509::Name CN=Puppet CA: localhost>, serial=#<OpenSSL::BN 2>, not_before=2014-09-24 00:26:27 UTC, not_after=2019-09-24 00:26:27 UTC>
[2] pry(#<Puppetserver::Ca::X509Loader>)> store.verify(certs.first)
=> false
[3] pry(#<Puppetserver::Ca::X509Loader>)> store
=> #<OpenSSL::X509::Store:0x00007fcd3f994ad0
 @chain=[#<OpenSSL::X509::Certificate subject=#<OpenSSL::X509::Name CN=localhost>, issuer=#<OpenSSL::X509::Name CN=Puppet CA: localhost>, serial=#<OpenSSL::BN 2>, not_before=2014-09-24 00:26:27 UTC, not_after=2019-09-24 00:26:27 UTC>],
 @error=20,
 @error_string="unable to get local issuer certificate",
 @time=nil,
 @verify_callback=nil>
```
Justin mentioned surfacing this error to a user

lib/puppetserver/ca/cli.rb
```diff
@@ -0,0 +1,76 @@
...
+          logger.warn parser.help
+          return 1
+        end
+      end
+
+      def self.parse_general_inputs(inputs)
+        parsed = {}
+        general_parser = OptionParser.new do |opts|
+          opts.banner = 'Usage: puppetserver ca <command> [options]'
+          opts.on('--help', 'This general help output') do |help|
```
amysahli: This general help output?

amysahli:
Do we want any error messages for unrecognized options?
ex) Below we error out instead of returning something like "--hello is not a valid blah":
```
puppetserver-ca setup --hello
Traceback (most recent call last):
	5: from /usr/local/bin/puppetserver-ca:23:in `<main>'
	4: from /usr/local/bin/puppetserver-ca:23:in `load'
	3: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/exe/puppetserver-ca:5:in `<top (required)>'
	2: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/cli.rb:34:in `run!'
	1: from /usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/setup_command.rb:69:in `parse'
/usr/local/lib/ruby/gems/2.5.0/gems/puppetserver-ca-0.0.1.pre.dev1/lib/puppetserver/ca/setup_command.rb:95:in `parse_inputs': invalid option: --hello (OptionParser::InvalidOption)
```

lib/puppetserver/ca/cli.rb
```diff
@@ -0,0 +1,76 @@
...
+      def self.run!(cli_args = ARGV, out = STDOUT, err = STDERR)
+        logger = Puppetserver::Ca::Logger.new(:info, out, err)
+        parser, general_options, unparsed = parse_general_inputs(cli_args)
+
+        if general_options['version']
+          logger.inform Puppetserver::Ca::VERSION
+          return 0
+        end
+
+        subcommand = VALID_COMMANDS[unparsed.first]
```
amysahli: Do we want to return an error if it's an unrecognized subcommand, or just return help message?

amysahli:
I think we need a more complete output for the initial --help w/out subcommand:
```
 puppetserver-ca --help
Usage: puppetserver ca <command> [options]
        --help                       This general help output
        --version                    Output the version
```

Ticket looks something like:
```
/opt/puppetlabs/bin/puppetserver ca --help
--
Manage the Private Key Infrastructure for Puppet Server's built-in Certificate Authority
 
Usage: puppetserver ca <action> [options]
 
Available subcommands:
puppetserver ca setup         Setup the CA with a private key, cert bundle, and crl chain
 
 
General Options:
-h, --help                Print this help message
-v, --version            Print the version of
--format FORMAT   Output format to display: human (default) or json
--config CONF        Path to the puppet.conf file to read
--verbose                Display verbose logging
 
Actions:
setup:
--private-key KEY    Path to a PEM encoded RSA private key
--crl-chain CRL        Path to a file containing one or more PEM encoded X.509 CRLs
--cert-bundle CERT Path to a file containing one or more PEM encoded X.509 Certificates.
The certificate to be used by this CA should be the first, followed by the
issuing CA's certificate. Additional issuers may follow with the root CA being last.
```

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      def validate_inputs(input, usage)
+        exit_code = nil
+
+        if input['cert-bundle'].nil? || input['private-key'].nil?
+          @logger.err 'Error:'
+          @logger.err 'Missing required argument'
+          @logger.err '    Both --cert-bundle and --private-key are required'
+          @logger.err ''
+          @logger.err usage
```
amysahli: I notice we use @logger a lot.  Would there be any advance/disadvantage to using an ```attr_reader :logger?```

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,136 @@
...
+
+        parser.parse(inputs)
+
+        return parser, parsed
+      end
+
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = 'Usage: puppetserver ca setup [options]'
+          opts.on('--help', 'This setup specific help output') do |help|
```
amysahli: Displays this setup specific help output

lib/puppetserver/ca/x509_loader.rb
```diff
@@ -0,0 +1,123 @@
...
+      def validate_crl_and_cert(crl, cert)
+        unless crl.issuer == cert.subject
+          @errors << 'Leaf CRL was not issued by leaf certificate'
+        end
+      end
+
+      def validate_full_chain(certs, crls)
+        store = OpenSSL::X509::Store.new
+        certs.each {|cert| store.add_cert(cert) }
+        if crls
```
justinstoller: `crls` is an array, but potentially an empty one. We should do an empty check not an existence check before adding crl checking flags to the store.

lib/puppetserver/ca/x509_loader.rb
```diff
@@ -0,0 +1,123 @@
...
+          store.flags = OpenSSL::X509::V_FLAG_CRL_CHECK | OpenSSL::X509::V_FLAG_CRL_CHECK_ALL
+          crls.each {|crl| store.add_crl(crl) }
+        end
+
+        unless store.verify(certs.first)
+          @errors << 'Leaf certificate could not be validated'
```
justinstoller: We should be surfacing any errors from the store as well (perhaps at a higher verbosity level).

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      # Return the correct confdir. We check for being root (user id == 0)
+      # on *nix, else the user path. We do not include a check for running
+      # as Adminstrator since non-development scenarios for Puppet Server
+      # on Windows are unsupported.
+      def user_specific_conf_dir
+        if Gem.win_platform? && Process::UID.eid == 0
+          '/etc/puppetlabs/puppet'
+        else
+          "#{ENV['HOME']}/.puppetlabs/etc/puppet"
```
amysahli: what's the "." before puppetlabs for?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+        end
+      end
+
+      def user_specific_conf_file
+        user_specific_conf_dir + '/puppet.conf'
+      end
+
+      def load
+        unless @using_default_location && !File.exist?(@config_path)
```
amysahli: a little hard to read.  this is **if** !@using_default_location && File.exist?(@config_path) , right?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+
+      # Resolve the cacert, cakey, and cacrl settings from default values,
+      # with any overrides for the specific settings or their dependent
+      # settings (ssldir, cadir) taken into account.
+      def resolve_settings(overrides = {})
+        unresolved_setting = /\$[a-z_]+/
+
+        # Returning the key for unknown keys (rather than nil) is required to
+        # keep unknown settings in the string for later verification.
+        substitutions = Hash.new {|h, k| k }
```
amysahli: i haven't seen this before.  cool!

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        settings[:cacert] = overrides.fetch(:cacert, '$cadir/ca_crt.pem')
+        settings[:cakey] = overrides.fetch(:cakey, '$cadir/ca_key.pem')
+        settings[:cacrl] = overrides.fetch(:cacrl, '$cadir/ca_crl.pem')
+
+        settings.each_pair do |key, value|
+          settings[key] = value.sub(unresolved_setting, substitutions)
+        end
+
+        settings.each_value do |value|
```
amysahli: Could this logic be combined in the loop above when you assign the value?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        cadir = overrides.fetch(:cadir, '$ssldir/ca')
+        settings[:cadir] = substitutions['$cadir'] = cadir.sub(unresolved_setting, substitutions)
+
+        settings[:cacert] = overrides.fetch(:cacert, '$cadir/ca_crt.pem')
+        settings[:cakey] = overrides.fetch(:cakey, '$cadir/ca_key.pem')
+        settings[:cacrl] = overrides.fetch(:cacrl, '$cadir/ca_crl.pem')
+
+        settings.each_pair do |key, value|
+          settings[key] = value.sub(unresolved_setting, substitutions)
```
amysahli: I'm not great with regular expressions.  Could you tell me what's happening here/why? ```value.sub(unresolved_setting, substitutions)```

lib/puppetserver/ca/x509_loader.rb
```diff
@@ -0,0 +1,123 @@
...
+          @errors += errs
+        end
+
+        return certs
+      end
+
+      def load_key(key_path)
+        begin
+          OpenSSL::PKey.read(File.read(key_path))
+        rescue ArgumentError => e
```
amysahli: If key_path file is empty, will this cause an argumentError?

justinstoller: Oh that's weird

What version of Ruby are you using?

```
irb(main):011:0> OpenSSL::PKey.read(File.read('foo-key.pem'))
ArgumentError: Could not parse PKey: no start line
        from (irb):11:in `read'
        from (irb):11
        from bin/console:14:in `<main>'
```

amysahli: ^ ruby 2.5 (I know I shouldnt be so high).

lib/puppetserver/ca/puppet_config.rb
justinstoller: Sorry, if this is overexplaining, this is what the block is doing, what sub is doing, and what the regex does.

The block is going over the so-far-computed values of `cacert`, `cakey`, and cacrl and interpolating any references to `$cadir`, `$ssldir`, or `$confdir` within them to their computed values.

`sub` will take a regex, if it has a match it will look up the match in a hash if that's passed in as a second arguement, replacing the matched substring with the value at that key in the hash.

eg.
```
substitutions = {"$confdir" => '/var/lib'}
regex = /<regex-that-matches-$confdir>/
"$confdir/foo/bar".sub(regex, substitutions)
  => "/var/lib/foo/bar"
```

And the regex itself should match a valid puppet setting that starts with a '$' (including the dollar sign) and ends with '/' (excluding the slash). The regex,
`/\$[a-z_]+/`, will match a dollar sign followed by one or more of lowercase letters or an underscore.

lib/puppetserver/ca/puppet_config.rb
justinstoller: Yeah, that's a good idea

lib/puppetserver/ca/puppet_config.rb
justinstoller: yeah, `unless` is often discouraged because it can be hard to grok. this is a good example of it making total sense when you're in the thick of it, but not afterwards. I'll change this to use `if`.

lib/puppetserver/ca/cli.rb
austb: We should return an exit code >0 for an unrecognized subcommand (which we do below). It could be nice, but not entirely necessary to print an error message before the help text.

lib/puppetserver/ca/puppet_config.rb
justinstoller: That's our [default location](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/run_mode.rb) when running as a non-root/Administrator user (even in Windows).

lib/puppetserver/ca/cli.rb
justinstoller: I think we want to exit 1 (denoting an error) and print the help to stderr. We should probably also print out that we did not recognize a subcommand if one was passed. I'll add that around line 42 where we handle missing subcommands.

lib/puppetserver/ca/cli.rb
justinstoller: I verb poorly, discussed IRL, "Displays this general help output" is what it should be.

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+
+        loader = X509Loader.new(bundle_path, key_path, chain_path)
+        return 1 if log_possible_errors(loader.errors)
+
+        puppet = PuppetConfig.parse(config_path)
+        return 1 if log_possible_errors(puppet.errors)
+
+
+        File.open(puppet.settings[:cacert], 'w') do |f|
```
amysahli: This is just a reminder (Justin and I discussed):
We'll probably want to create a /puppetlabs/etc/puppet directory if there isn't one made yet.
When I blew away my .puppetlabs directory, things started to fail:
```
 rm -rf .puppetlabs
 ~ $🐰  cd .puppetlabs
-bash: cd: .puppetlabs: No such file or directory
 ~ $🐰  cd Development/puppetserver-ca-cli/
 ~/Development/puppetserver-ca-cli (SERVER-2172-wip) $🐰  bundle exec exe/puppetserver-ca setup --private-key=bin/privkey.pem --cert-bundle=bin/ca.pem
Warning:
    No CRL chain given
    Full CRL chain checking will not be possible

bundler: failed to load command: exe/puppetserver-ca (exe/puppetserver-ca)
Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/amy.sahli/.puppetlabs/etc/puppet/ssl/ca/ca_crt.pem
  /Users/amy.sahli/Development/puppetserver-ca-cli/lib/puppetserver/ca/setup_command.rb:38:in `initialize'
  /Users/amy.sahli/Development/puppetserver-ca-cli/lib/puppetserver/ca/setup_command.rb:38:in `open'
  /Users/amy.sahli/Development/puppetserver-ca-cli/lib/puppetserver/ca/setup_command.rb:38:in `run!'
  /Users/amy.sahli/Development/puppetserver-ca-cli/lib/puppetserver/ca/cli.rb:39:in `run!'
  exe/puppetserver-ca:5:in `<top (required)>'
```

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+        @settings = nil
+        @errors = []
+      end
+
+      # Return the correct confdir. We check for being root (user id == 0)
+      # on *nix, else the user path. We do not include a check for running
+      # as Adminstrator since non-development scenarios for Puppet Server
+      # on Windows are unsupported.
+      def user_specific_conf_dir
+        if Gem.win_platform? && Process::UID.eid == 0
```
austb: Should this be an `||`? E.g. "if running on windows or running as root"? Currently this seems like the root user's home directory on *nix would end up as `/var/root/.puppetlabs/etc/puppet`.

lib/puppetserver/ca/setup_command.rb
justinstoller: I used the instance variable directly since there's no external reader of the logger, and we don't mock out the logger in the tests (and if we did that in a test of SetupCommand, we could pass the logger into the instance under test).

Having said, that `attr_reader :logger` might be more idiomatic?

lib/puppetserver/ca/puppet_config.rb
austb: Or maybe `!Gem.win_platform?` was the intended test?

justinstoller: `!Gem.win_platform?`

justinstoller:
Going over this the changes, in summary are:
* Improve help messages:
  * use verbs in the actual help line
  * add information about required/optional flags
  * include list of subcommands in general help output
* x509 loading:
  * add comments explaining store checking to `validate_full_chain`
  * expose store errors in the error output
  * fix bug in store check logic (check for emptiness, rather than existence)
* config parser:
  * remove `unless` usage in favor of `if`
  * comment usage of String#sub()
  * fold setting interpolation failure warning into same loop as interpolation
  * rescue both ArgumentError & PKeyError for forwards compatibility
  * fix windows check (`!Gem.win_platform?`)
* SetupCommand:
  * handle unrecognized flags gracefully
  * explain more about full chain checking on the agent when not providing a CRL
* Misc:
  * check exit status more regularly in tests
  * add error message when invalid subcommand

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -0,0 +1,121 @@
...
+          results = parse_text(File.read(@config_path))
+        end
+
+        results ||= {}
+        results[:main] ||= {}
+        results[:master] ||= {}
+
+        overrides = results[:main].merge(results[:master])
+
+        @settings = resolve_settings(overrides).freeze
```
austb: Might also be nice to log the final resolved settings at the verbose level. Should probably also log at the various stages so they can track their settings back to where they originate.

lib/puppetserver/ca/cli.rb
```diff
@@ -0,0 +1,101 @@
...
+          end
+
+          opts.separator ''
+          opts.separator 'Sub-Command Options:'
+          VALID_COMMANDS.each_pair do |command, cls|
+            opts.separator "  #{command}:"
+            subopts = cls.parser.summarize.
+              drop_while{|l| !(l =~ /Options:/) }.
+              select{|l| l =~ /^\s*--/ }.
+              reject{|l| l =~ /--help|--version/ }
```
justinstoller: This is gross and should either be cleaned up or removed. I pushed up this wip (besides it being the end of the day) to get folks feelings on whether or not a direction _like_ this is worthwhile or whether we should dumb down the help output and have a simpler implementation.

lib/puppetserver/ca/x509_loader.rb
Magisus: I like putting all of this under the `validate` banner. If Amy was referring to whether to move this line down inside the subsequent method, in this case I would keep it outside, since the `validate_crl_and_cert` method cares about something more specific than the whole bundle and chain. In order to give it only what it needs, we need to make sure these things aren't empty first.

lib/puppetserver/ca/cli.rb
```diff
@@ -0,0 +1,101 @@
...
+require 'puppetserver/ca/version'
+require 'puppetserver/ca/setup_command'
+require 'puppetserver/ca/logger'
+
+module Puppetserver
+  module Ca
+    class Cli
+      VALID_COMMANDS = {'setup' => SetupCommand}
+
+      def self.run!(cli_args = ARGV, out = STDOUT, err = STDERR)
```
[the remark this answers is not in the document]
justinstoller: damnit

lib/puppetserver/ca/setup_command.rb
```diff
@@ -0,0 +1,178 @@
...
+
+      def self.description
+        'Set up the CA\'s key, certs, and crls'
+      end
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def run!(input)
```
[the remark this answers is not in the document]
justinstoller: Yeah, I keep using it more in (though not quite) a clojure sense.

spec/puppetserver/ca/cli_spec.rb
```diff
@@ -0,0 +1,463 @@
...
+      Dir.mktmpdir do |tmpdir|
+        with_files_in tmpdir do |bundle, key, chain, conf|
+          Puppetserver::Ca::Cli.run!(['setup',
+                                      '--config', conf,
+                                      '--cert-bundle', bundle,
+                                      '--private-key', key,
+                                      '--crl-chain', chain],
+                                      stdout,
+                                      stderr)
+        end
```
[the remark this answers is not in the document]
Magisus: Could be just verifying we don't throw I suppose

Magisus:
So I tried using this without a CRL on a clean install of Puppetserver, and the server refused to start because if any of its CA files are missing, it doesn't generate them, just fails hard. See https://github.com/puppetlabs/puppetserver/blob/2e271f7c6768cf60f56d396f74b28ce62e3bd677/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L180-L182 So I'm not sure to what extent we can consider the CRL optional...

justinstoller:
I can update this to generate a CRL if one isn't given?

Magisus:
That works for me, so long as it's documented.

justinstoller:
Updated

PR #2, opened 2018-07-19
(maint) Bump to 0.0.1 from prerelease

justinstoller:


justinstoller:
Previously, we used a semver "pre-release" version that contained alpha characters after the standard version number. Our release automation does not handle bumping from pre-release versions, so this bumps us from a pre-release of 0.0.1 to an actual release version of 0.0.1.

PR #3, opened 2018-07-23
(SERVER-2258) Add revoke action

justinstoller:


lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,89 @@
...
+      # class method parser returns an instance of Ruby's OptionParser
+      # (or a custom object that implements summarize() and help().
+      def self.parser(parsed = {})
+        OptionParser.new do |o|
+          o.banner = BANNER
+          o.on('--help', 'Display this revoke specific help output') do |help|
+            parsed['help'] = true
+          end
+          o.on('--version', 'Output the version') do |v|
+            parsed['version'] = true
```
justinstoller: This should probably go away. I don't think `--version` in the actions is necessary nor makes a bunch of sense.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,89 @@
...
+        parser = self.class.parser(results)
+
+        begin
+          parser.order!(args) do |unknown|
+            results[:certname] << unknown
+          end
+        rescue OptionParser::ParseError => e
+          unparsed += e.args
+          unparsed << args.shift unless args.first =~ /^-{1,2}/
+          retry
```
justinstoller: This is a little bit of arcanum from OptParse, we might want to throw it in a util method so action implementations don't have to re-write it every time?

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,89 @@
...
+        unless unparsed.empty?
+          @logger.err 'Error:'
+          @logger.err 'Unknown arguments or flags:'
+          unparsed.each do |arg|
+            @logger.err "    #{arg}"
+          end
+
+          @logger.err ''
+          @logger.err parser.help
+          exit_code = 1
```
justinstoller: This also might be good in a util method along with the unknown option parsing above? Or we could try to refactor it into the top level Cli class so that actions don't need to know about it??

lib/puppetserver/ca/revoke_action.rb
justinstoller: We might need something like https://github.com/puppetlabs/puppetserver-ca-cli/commit/4a2aa6e2f81454338bfa3a6042a425e959557ea6 though (if we don't define `--help` and `--version` on a parser and it runs across them it will by default raise `SystemExit`. Though I think with the `Cli` class now parses any `--help` and `--version` flags and Does the Right Thing™.

justinstoller:
I got this to functionally work locally with a puppetserver running from the repl as localhost.

This might be valuable to others who want to experiment with it locally as well. I have this repo, puppetserver, and puppet all checked out and terminal windows open to each. With puppetserver already running from the repl:

From my `puppet` checkout I can create new certs:
```
09:45:06 sellout:puppet justin (5.5.x<>) ‖ be puppet agent -t --server localhost --certname bazbar
Info: Creating a new SSL key for bazbar
Info: csr_attributes file loading from /Users/justin/.puppetlabs/etc/puppet/csr_attributes.yaml
Info: Creating a new SSL certificate request for bazbar
Info: Certificate Request fingerprint (SHA256): 01:4D:76:CA:15:C7:80:33:7D:B8:C3:B6:40:AD:8E:94:95:D8:04:C3:0C:1B:2E:5B:3B:B2:BC:E7:25:75:DC:25
Exiting; no certificate found and waitforcert is disabled

09:55:53 sellout:puppet justin (5.5.x<>) ‖ be puppet cert sign bazbar
Signing Certificate Request for:
  "bazbar" (SHA256) 01:4D:76:CA:15:C7:80:33:7D:B8:C3:B6:40:AD:8E:94:95:D8:04:C3:0C:1B:2E:5B:3B:B2:BC:E7:25:75:DC:25
Notice: Signed certificate request for bazbar
Notice: Removing file Puppet::SSL::CertificateRequest bazbar at '/Users/justin/.puppetlabs/etc/puppet/ssl/ca/requests/bazbar.pem'

09:56:03 sellout:puppet justin (5.5.x<>) ‖ be puppet agent -t --server localhost --certname bazbar --noop
Info: Caching certificate for bazbar
Info: Caching certificate for bazbar
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Applying configuration version '1532451377'
Notice: Applied catalog in 0.01 seconds
```

Then in my `puppetserver` window I can see it having processed the report for `bazbar` (verifying everything worked correctly on the server side):
```
2018-07-24 09:56:17,542 DEBUG [qtp2082838917-106] [puppetserver] Puppet report supports formats: json pson yaml
2018-07-24 09:56:17,558 DEBUG [qtp2082838917-106] [puppetserver] Puppet Received report to process from bazbar
2018-07-24 09:56:17,559 DEBUG [qtp2082838917-106] [puppetserver] Puppet Processing report from bazbar with processor Puppet::Reports::Store
```

Then from this project's checkout I can revoke the cert:
```
09:55:40 sellout:puppetserver-ca-cli justin (server2258-revoke-action *) ‖ be ./exe/puppetserver-ca revoke bazbar
Revoked certificate for bazbar
```

And I can see puppetserver actually revoke it back in its window:
```
2018-07-24 09:56:46,839 DEBUG [qtp2082838917-116] [p.p.certificate-authority] Revoked bazbar certificate with serial 4
2018-07-24 09:56:47,415 DEBUG [clojure-agent-send-off-pool-4] [p.t.s.w.filesystem-watch-core] Got 1 event(s) in directory /Users/justin/.puppetlabs/etc/puppet/ssl/ca
2018-07-24 09:56:47,421 INFO  [clojure-agent-send-off-pool-4] [o.e.j.u.s.SslContextFactory] x509=X509@5119578f(private key,h=[puppet, localhost],w=[]) for InternalSslContextFactory@3492fcf(null,null)
```

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,123 @@
...
+        elsif results[:certname].length > 1
+          @logger.err 'Only one certname is allowed at a time'
+          @logger.err 'Received certnames:'
+          results[:certname].each do |certname|
+            @logger.err "    #{certname}"
+          end
+          exit_code = 1
+
+        else
+          results[:certname] = results[:certname].first
```
justinstoller: Shouldn't reassign a hash value like this, should maybe store the args in an intermediate place that we do validation on before assigning certname to result.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,123 @@
...
+      # print information to @logger and return a valid exit code
+      def run(args)
+        certname = args[:certname]
+        body = JSON.dump({ desired_state: 'revoked' })
+        headers = {
+          'User-Agent' => 'PuppetserverCaCli',
+          'Content-Type' => 'application/json',
+          'Accept' => 'application/json'
+        }
+        uri = URI('https://localhost:8140/puppet-ca/v1/certificate_status/' +
```
justinstoller: This (host and port) (for now) and the file paths should be pulled from puppet.conf. Host and port maybe should eventually be pulled from puppetserver's configs.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,123 @@
...
+                  certname +
+                  '?environment=production')
+        store = OpenSSL::X509::Store.new
+        store.add_file(File.expand_path('~/.puppetlabs/etc/puppet/ssl/certs/ca.pem'))
+        store.add_file(File.expand_path('~/.puppetlabs/etc/puppet/ssl/certs/localhost.pem'))
+        store.add_file(File.expand_path('~/.puppetlabs/etc/puppet/ssl/crl.pem'))
+
+        result = Net::HTTP.start(uri.host, uri.port, use_ssl: true, cert_store: store) do |connection|
```
justinstoller: I could see us doing two things here, either
a) pulling this http calls some util methods
b) pulling the logic of talking to the CA into a CA class with methods like `revoke` and `sign`.

Not sure which is best.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,123 @@
...
+          request.body = body
+          connection.request(request)
+        end
+
+        if result.is_a? Net::HTTPSuccess
+          @logger.inform "Revoked certificate for #{certname}"
+          return 0
+        else
+          @logger.err "Error - gross debug output:"
+          @logger.err "    #{result.inspect}"
```
justinstoller: Need to review the output from cert on when and how it errs.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,123 @@
...
+                  certname +
+                  '?environment=production')
+        store = OpenSSL::X509::Store.new
+        store.add_file(File.expand_path('~/.puppetlabs/etc/puppet/ssl/certs/ca.pem'))
+        store.add_file(File.expand_path('~/.puppetlabs/etc/puppet/ssl/certs/localhost.pem'))
+        store.add_file(File.expand_path('~/.puppetlabs/etc/puppet/ssl/crl.pem'))
+
+        result = Net::HTTP.start(uri.host, uri.port, use_ssl: true, cert_store: store) do |connection|
+          request = Net::HTTP::Put.new(uri, headers)
+          request.body = body
```
justinstoller: There's not a lot of custom logic with this request so I'm hesitant to write much in the way of tests around it and leave the does-it-actually-talk-to-a-master test to acceptance level. I assume whatever helper we pull this logic into (eg an http util method or a CA class) we can stub it out in higher level specs.

lib/puppetserver/ca/revoke_action.rb
Magisus: I always though it was weird that these we considered flags of the actions in Puppet (at least according to the docs). I would like to see them go away in that context here if we can.

lib/puppetserver/ca/revoke_action.rb
Magisus: I think it makes sense to keep it in the actions, but a utility method definitely seems good.

lib/puppetserver/ca/revoke_action.rb
Magisus: Yeah that seems better. I mostly don't like that the type changes.

lib/puppetserver/ca/revoke_action.rb
Magisus: Are they available in puppetserver's configs, or are you thinking of adding them (presumably as a `ca_server` setting)?

lib/puppetserver/ca/revoke_action.rb
Magisus: Honestly I think I'd prefer (a), at least for now. I'd rather not introduce too many abstractions right now, but having the http calls in util methods gives us more flexibility to swap out the client if we need to.

lib/puppetserver/ca/revoke_action.rb
Magisus: I think acceptance is fine here (though I think we should write a standalone test like we did for the [setup command](https://github.com/puppetlabs/puppetserver/blob/master/acceptance/suites/tests/certificate_authority/intermediate_ca.rb)). Because of how simple this is, I feel like introducing webmock would most just end up testing the HTTP library, which feels unnecessary.

lib/puppetserver/ca/revoke_action.rb
lazartea: Are we going to add the ``--all`` flag?

lib/puppetserver/ca/revoke_action.rb
justinstoller: I was thinking that they'd be mounted as on the master's ssl-host and ssl-port in the webserver.conf

lib/puppetserver/ca/revoke_action.rb
lazartea: So I'm looking at the docs, and I don't think the revoke command uses the ``--all`` flag. Ignore me.

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -72,17 +72,33 @@ def resolve_settings(overrides = {})
...
+        settings[:certname] = substitutions['$certname'] = overrides.fetch(:certname, `hostname`.chomp)
+        server = overrides.fetch(:server, '$certname')
+        settings[:server] = substitutions['$server'] = server.sub(unresolved_setting, substitutions)
+
+        settings[:ca_server] = overrides.fetch(:ca_server, '$server').sub(unresolved_setting, substitutions)
+
+        settings[:masterport] = substitutions['$masterport'] = overrides.fetch(:masterport, '8140')
+        settings[:ca_port] = overrides.fetch(:ca_port, '$masterport').sub(unresolved_setting, substitutions)
+
+        settings[:localcacert] = overrides.fetch(:localcacert, '$certdir/ca.pem')
```
justinstoller: probably time to just do defaults hash and a merge

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -72,17 +72,33 @@ def resolve_settings(overrides = {})
...
         settings[:ssldir] = substitutions['$ssldir'] = ssldir.sub('$confdir', confdir)
 
+        certdir = overrides.fetch(:certdir, '$ssldir/certs')
+        settings[:certdir] = substitutions['$certdir'] = certdir.sub(unresolved_setting, substitutions)
+
         cadir = overrides.fetch(:cadir, '$ssldir/ca')
         settings[:cadir] = substitutions['$cadir'] = cadir.sub(unresolved_setting, substitutions)
 
+        # Note the call to hostname below, this is horrific....
+        settings[:certname] = substitutions['$certname'] = overrides.fetch(:certname, `hostname`.chomp)
```
justinstoller: ....

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -72,17 +72,33 @@ def resolve_settings(overrides = {})
...
+        server = overrides.fetch(:server, '$certname')
+        settings[:server] = substitutions['$server'] = server.sub(unresolved_setting, substitutions)
+
+        settings[:ca_server] = overrides.fetch(:ca_server, '$server').sub(unresolved_setting, substitutions)
+
+        settings[:masterport] = substitutions['$masterport'] = overrides.fetch(:masterport, '8140')
+        settings[:ca_port] = overrides.fetch(:ca_port, '$masterport').sub(unresolved_setting, substitutions)
```
justinstoller: I had to update this class for the hostcert/crl/localcacert so I added `ca_{server,port}` too, but it might be out of scope

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+Given a valid certname, instructs the CA to revoke it over
+HTTPS using the local agent's PKI
+
+Options:
+BANNER
+
+      def self.parser(parsed = {})
+        OptionParser.new do |o|
+          o.banner = BANNER
+          o.on('--config', 'Path to puppet.conf') do |conf|
```
justinstoller: should this be lifted to the main cli class? I think eventually it would then need to reference both a puppet server's conf.d and a puppet.conf??? But maybe the Cli class is a better place for it for the time being regardless?

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+        result = Net::HTTP.start(uri.host, uri.port, use_ssl: true, cert_store: store) do |connection|
+          request = Net::HTTP::Put.new(uri, headers)
+          request.body = body
+          connection.request(request)
+        end
+
+        if result.is_a? Net::HTTPSuccess
+          @logger.inform "Revoked certificate for #{certname}"
+          return 0
+        elsif result.is_a? Net::HTTPNotFound
```
justinstoller: turns out cert doesn't do much for errors, afaict

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+                       settings[:ca_port],
+                       certname)
+
+        result = Net::HTTP.start(uri.host, uri.port, use_ssl: true, cert_store: store) do |connection|
+          request = Net::HTTP::Put.new(uri, headers)
+          request.body = body
+          connection.request(request)
+        end
+
+        if result.is_a? Net::HTTPSuccess
```
justinstoller: both cert and the api will gladly return success if the cert has already been revoked, can we fix that in the api as part of this?

lib/puppetserver/ca/puppet_config.rb
justinstoller: So.. how we get a certname default seems like it could be problematic.

We definitely shouldn't just use backticks in the middle of a giant method. I'll probably put this behind a method that shells out and move it into the load, parse, or initialize methods (I don't like the idea of doing IO in initialize, but it's also something that should only be done once). So something like:
```rb
        @config_path = supplied_config_path || user_specific_conf_file
        @certname = run("hostname").chomp

```
For this PR `certname` is only need to figure out where `hostcert` lives, and in a default case we could skip certname and figure that `hostcert` is the `.pem` file in `certdir` that _isn't_ `ca.pem`. We could even reverse engineer `certname` from that value... That would avoid shelling out. But I'm unsure of the cases that would or wouldn't work...

Set certname, default hostcert (derived from certname) <- works
Set certname, set hostcert <- works
Set hostcert explicity, default certname <- no good

There's also the dev case of having a bunch of random crap in your certdir from testing, ie mine:
```
11:27:58 puppetserver-ca-cli justin (server2258-revoke-action $) ‖ ls ~/.puppetlabs/etc/puppet/ssl/certs/
bazbar.pem
ca.pem
foobar.pem
localhost.pem
```

Note, in our dev env though we use "localhost" which "hostname" wouldn't guess on my box, we, however, also explictly set certname.

I don't know, I'm thinking shelling out might still be the best bet?

Little rubber ducky-ing, but would love your thoughts, @Magisus @amysahli

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+      end
+
+      def make_uri(host, port, certname)
+        URI('https://' + host + ':' + port +
+            '/puppet-ca/v1/certificate_status/' +
+            certname +
+            '?environment=production')
+      end
+
+      def revoke(certname, settings)
```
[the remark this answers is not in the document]
justinstoller: I'm thinking that whatever I do we'll want to change it relatively soon after we add multiple HTTP methods/payloads, so I'm just going to move this wholesale into a Util and we can change it as we see fit in the next PRs. How's that sound?

lib/puppetserver/ca/revoke_action.rb
justinstoller: One thing it does is return the serial number rather than the certname, which the API doesn't do. I don't know if that's super valuable or not (it becomes important when `all_duplicate_certnames` is `true`).

lib/puppetserver/ca/revoke_action.rb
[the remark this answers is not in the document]
lazartea: I'm seeing a lot of overlap with things I'm working on for the sign command, so I agree that these would fit best as Utils (with perhaps some slight modifications)

lib/puppetserver/ca/puppet_config.rb
justinstoller: I think that refactor would screw up what @amysahli is working on? Should I leave these changes as is and then we can refactor after both our PRs land?

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+          @logger.err "    Could not find certificate for #{certname}"
+          return 1
+        else
+          @logger.err 'Error - gross debug output:'
+          @logger.err "    #{result.inspect}"
+          return 1
+        end
+      end
+
+      def validate_file_paths(one_or_more_paths)
```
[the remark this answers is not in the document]
justinstoller: I'll leave this as is because @amysahli is creating a utils with it. I'll update this code once her PR lands.

lib/puppetserver/ca/revoke_action.rb
Magisus: Probably. What response would you expect?

lib/puppetserver/ca/utils.rb
```diff
@@ -0,0 +1,36 @@
+module Puppetserver
+  module Ca
+    module Utils
+      def self.parse_without_raising(parser, args)
```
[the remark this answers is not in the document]
Magisus: It communicates the intent 🤷‍♀️

lib/puppetserver/ca/puppet_config.rb
amysahli: I think we don't need the ```.sub(unresolved_setting, substitutions)``` above for :ca_port since it will go through the .gsub below, right?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -72,17 +72,33 @@ def resolve_settings(overrides = {})
...
+
         cadir = overrides.fetch(:cadir, '$ssldir/ca')
         settings[:cadir] = substitutions['$cadir'] = cadir.sub(unresolved_setting, substitutions)
 
+        # Note the call to hostname below, this is horrific....
+        settings[:certname] = substitutions['$certname'] = overrides.fetch(:certname, `hostname`.chomp)
+        server = overrides.fetch(:server, '$certname')
+        settings[:server] = substitutions['$server'] = server.sub(unresolved_setting, substitutions)
+
+        settings[:ca_server] = overrides.fetch(:ca_server, '$server').sub(unresolved_setting, substitutions)
```
amysahli: Do we need the ```.sub(unresolved_setting, substitutions)``` here too?  maybe it could just get evaluated below with the ```.gsub``` ?

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+require 'json'
+
+module Puppetserver
+  module Ca
+    class RevokeAction
+      SUMMARY = 'Revoke a given certificate'
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca revoke [--help|--version]
+  puppetserver ca revoke CERTNAME
```
amysahli: If we want to let them do --config, let's add it here.

lib/puppetserver/ca/revoke_action.rb
amysahli: We don't need --help here too?

lib/puppetserver/ca/revoke_action.rb
amysahli: Would result.is_a?  be better than result.status?  like ```case response.status```

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,149 @@
...
+        store.add_file(crl)
+
+        store
+      end
+
+      def make_uri(host, port, certname)
+        URI('https://' + host + ':' + port +
+            '/puppet-ca/v1/certificate_status/' +
+            certname +
+            '?environment=production')
```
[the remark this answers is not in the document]
justinstoller: The old Puppet code required every request to include an environment, but in cases like certificates (where environment doesn't make sense) it would ignore the environment param (though still, unfortunately require it). I should double check to see if Puppet Server made the environment optional when it re-implemented the CA endpoints (and we should probably make it optional in Puppet 6 if not)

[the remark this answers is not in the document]
Magisus: Based on the fact that this code https://github.com/puppetlabs/puppet/blob/master/lib/puppet/rest/routes.rb works, I think it's optional now.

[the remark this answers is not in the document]
justinstoller: Double checked, the docs say it's still required, but I couldn't see it in the code and testing manually worked w/o supplying it. Removing.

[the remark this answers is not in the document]
Magisus: We should probably update the docs for Puppet 6 then.

lib/puppetserver/ca/puppet_config.rb
justinstoller: Good catch

lib/puppetserver/ca/revoke_action.rb
justinstoller: Hrm... I hadn't thought about the specifics. We're doing a PUT which is supposed to be idempotent (and we retain the list of revoked certs in the inventory), so theoretically I suppose this should be a success as long as the end state of the http call is that the cert is revoked regardless of its state beforehand.

Maybe instead I should make two calls? One to query the state of the cert being revoked and then one to update it?

justinstoller: This is starting to sound even more like @lazartea 's PR. An `--all` flag doesn't make sense for revoke (it does/did exist in `puppet cert` and it been a feature request to _remove_ that ability for years), but should we accept a `--certname` flag with that potentially contains a list of comma separated certnames?

Magisus: I think it doesn't work now. The docs say you can use it, but if you use it, it errors. At least that's the way it works for `clean`.

I like the idea of accepting a `--certname` flag here. Consistency seems good.

Magisus: I'm also fine returning success so long as the cert ends up revoked, whether or not it started out that way.

lib/puppetserver/ca/revoke_action.rb
justinstoller: Okay, it looks like the current tool (at least according to its documentation) allows for specifying a cert to revoke either by its certname or its serial number (in hex format):
```
be puppet cert --help
..snip..
* revoke:
  Revoke the certificate of a client. The certificate can be specified either
  by its serial number (given as a hexadecimal number prefixed by '0x') or by its
  hostname. The certificate is revoked by adding it to the Certificate Revocation
  List given by the 'cacrl' configuration option. Note that the puppet master
  needs to be restarted after revoking certificates.
```

However the API [doesn't let you do this](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1260-L1263), and doesn't return the serial of the revoked cert.

I think we might want to add this functionality later - I assume its priority would depend on the usage of `allow_duplicate_certnames` in the field? Should we make a ticket for that @Magisus ?

Magisus: Yeah I'll file one.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -1,6 +1,6 @@
 require 'optparse'
 require 'openssl'
-require 'puppetserver/utils/file_utils'
+require 'puppetserver/utils/file_utilities'
```
justinstoller: these are named incorrectly currently, I should probably pull these changes into their own commit.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,150 @@
...
+        revoke_certs(certnames, puppet.settings)
+      end
+
+      def revoke_certs(certnames, settings)
+        url = HttpUtilities.make_ca_url(settings[:ca_server],
+                       settings[:ca_port],
+                       'certificate_status')
+
+        failed = HttpUtilities.with_connection(url, settings) do |connection|
+          certnames.reduce(false) do |has_failures, certname|
```
justinstoller: This block (and the HttpUtilities interface) was fun to write but I wonder if I'm being a bit YAGNI/overwrought?

lib/puppetserver/ca/utils.rb
```diff
@@ -0,0 +1,41 @@
...
+            all << value
+          end
+
+          retry
+        end
+
+        return all, not_options, malformed_options, unknown_options
+      end
+
+      def self.print_invalid_args(logger, args, usage)
```
justinstoller: this is no longer shared code

lib/puppetserver/ca/revoke_action.rb
justinstoller: I'd love your thoughts on the above @pcarlisle

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -143,6 +158,10 @@ def parse_text(text)
       def explicitly_given_config_file_or_default_config_exists?
         !@using_default_location || File.exist?(@config_path)
       end
+
+      def run(command)
+        %x'#{command}'
```
amysahli: What's %x do?

justinstoller: It's like backticks, shells out and returns the commands output

justinstoller: but is a lot more grep-able than backticks

justinstoller: I think they handle errors slightly differently, I should maybe double check that.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,150 @@
...
+        @logger = logger
+      end
+
+      def parse(args)
+        results = {}
+        parser = self.class.parser(results)
+
+        _, non_flags, malformed_flags, unknown_flags = Utils.parse_without_raising(parser, args)
+
+        valid = valid?(results,
```
amysahli: Isn't results always going to be {} here?

lib/puppetserver/ca/revoke_action.rb
pcarlisle: It seems fine. You could improve the readability of the logic a little bit by changing failed to succeeded and inverting all the booleans.

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,150 @@
...
+        url = HttpUtilities.make_ca_url(settings[:ca_server],
+                       settings[:ca_port],
+                       'certificate_status')
+
+        failed = HttpUtilities.with_connection(url, settings) do |connection|
+          certnames.reduce(false) do |has_failures, certname|
+            url.resource_name = certname
+            result = connection.put(REQUEST_BODY, url)
+
+            has_failures ||= !check_result(result, certname)
```
pcarlisle: Because the `||=` operator shortcuts execution `check_result` will stop being called after the first failure.

lib/puppetserver/ca/revoke_action.rb
justinstoller: The parser, which includes a reference to it, mutates it during `Utils.parse_without_raising`

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,150 @@
...
+          return true
+        end
+      end
+
+      def run(args)
+        certnames = args['certnames']
+        config = args['config']
+
+        errors = FileUtilities.validate_file_paths(config)
+        unless errors.empty?
```
amysahli: Did you want to take this out into 
```
def log_possible_errors(maybe_errors)
        errors = Array(maybe_errors).compact
        unless errors.empty?
          @logger.err "Error:"
          errors.each do |message|
            @logger.err "    #{message}"
          end
          return true
        end
      end
```
like you did for import action?

lib/puppetserver/ca/puppet_config.rb
justinstoller: From what I've found they do handle errors the same.

lib/puppetserver/ca/revoke_action.rb
justinstoller: Yeah, good call

lib/puppetserver/ca/revoke_action.rb
```diff
@@ -0,0 +1,150 @@
...
+        end
+
+        puppet = PuppetConfig.parse(config)
+        unless puppet.errors.empty?
+          @logger.err 'Error:'
+          puppet.errors.each {|error| @logger.err "    #{error}" }
+          return 1
+        end
+
+        revoke_certs(certnames, puppet.settings)
```
amysahli: does this need to be return revoke_certs ?

lib/puppetserver/ca/revoke_action.rb
justinstoller: Good catch. No, it's not desired, I think we want to consider the outcome of each revocation independent.

lib/puppetserver/ca/revoke_action.rb
justinstoller: Doesn't need to have an explicit return. Ruby will return the results of the last expression (so whatever `revoke_certs` returns).

However it might make more semantic sense for `revoke_certs` to return some kind of success/failure indicator (like a boolean or a results array) and `run` to return 0 or 1 (exit codes), rather than `revoke_certs` having to know that it should return CLI exit codes...

lib/puppetserver/ca/revoke_action.rb
amysahli: How about instead of !check_result we just switch the true/falses below and rename the method to something like result_unsuccessful? with a question mark implying it will return a boolean

lib/puppetserver/ca/revoke_action.rb
amysahli: ^ oops.  I said the same thing basically

lib/puppetserver/ca/revoke_action.rb
amysahli: I'm wondering how you will solve this.  Maybe 
```
certnames.reduce(Set[false])
```
and then a has_failures << !check_result(result, certname)
and return failed.include?(false)

justinstoller:
I think I've addressed everyone's feedback, let me know if there's anything else!

PR #4, opened 2018-07-24
(SERVER-2255) Generate root and intermediate CA for puppet server

amysahli:


lib/puppetserver/ca/cli.rb
```diff
@@ -22,10 +26,13 @@ class Cli
 
       ACTION_OPTIONS = "\nAction Options:\n" +
         VALID_ACTIONS.map do |action, cls|
-          "  #{action}:\n" +
-          cls.parser.summarize.
+          summary = cls.parser.summarize.
             select{|line| line =~ /^\s*--/ }.
             reject{|line| line =~ /--help|--version/ }.join('')
+
```
amysahli: @justinstoller you did something similar, which is probably better.  I just left this in for now.

spec/shared_examples/cli_parsing.rb
```diff
@@ -6,13 +6,6 @@
     expect(exit_code).to be 0
   end
 
-  it 'prints the help output & returns 1 if no input is given' do
-    args = [action].compact
-    exit_code = Puppetserver::Ca::Cli.run(args, stdout, stderr)
-    expect(stderr.string).to match(usage)
-    expect(exit_code).to be 1
-  end
-
```
amysahli: 'generate' requires no input, so I moved this into the "setup" specs only.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,123 @@
...
+            ef.create_extension("basicConstraints", "CA:TRUE", true)])]))
+
+        csr.add_attribute(ext_req)
+        csr.sign(ca_key, OpenSSL::Digest::SHA256.new)
+
+        directory_name = "tmp"
+        Dir.mkdir(directory_name) unless File.exists?(directory_name)
+
+        File.open("#{directory_name}/req.pem", "w") do |f|
+          f.write(csr.to_s)
```
amysahli: Do we want to save their csr here?  or where would be good?  It also prints to stdout.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,250 @@
...
+        crl.version = 1
+        crl.issuer = ca_cert.subject
+
+        ef = extension_factory_for(ca_cert)
+        crl.add_extension(
+          ef.create_extension(["authorityKeyIdentifier", "keyid:always", false]))
+        crl.add_extension(
+          OpenSSL::X509::Extension.new("crlNumber", OpenSSL::ASN1::Integer(0)))
+
+        crl.last_update = Time.now
```
amysahli: Is there any value in doing something like was done above  "Time.now - (60*60*24)" ? or is Time.now sufficient?

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,250 @@
...
+      end
+
+      def self_signed_ca(key, name, ca_ttl)
+        cert = OpenSSL::X509::Certificate.new
+
+        cert.public_key = key.public_key
+        cert.subject = OpenSSL::X509::Name.new([["CN", name]])
+        cert.issuer = cert.subject
+        cert.version = 2
+        cert.serial = rand(2**128)
```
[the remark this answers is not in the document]
Magisus: This might be a good opportunity for us to do https://tickets.puppetlabs.com/browse/PUP-6824 maybe?

lib/puppetserver/ca/generate_action.rb
Magisus: I know you made some updates around making this more lenient somewhere recently per Josh's feedback, was that on the agent side or the server side? If it was the agent, then `Time.now` is probably fine here.

lib/puppetserver/ca/generate_action.rb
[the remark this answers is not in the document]
Magisus: I take that back, there's a lot more to that ticket than just changing this. That being the case, I think it might be good to bring that logic over.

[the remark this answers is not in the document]
justinstoller: If we bring the serial number bits over we should set the root CA serial to 1, the subordinate CA to 2 and put the number 3 in the serial file.

lib/puppetserver/ca/generate_action.rb
justinstoller: Puppet sets it to be Time.now - 1 second.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,250 @@
...
+        int_key = create_private_key(private_key_length)
+        int_csr = create_csr(int_key, int_ca_name)
+        int_cert = sign(root_key, root_cert, int_csr, ca_ttl)
+        int_crl = create_crl_for(int_cert, int_key, ca_ttl)
+
+        Puppetserver::Utils::FileUtilities.ensure_dir(puppet.settings[:cadir])
+
+        future_files = [ { location: puppet.settings[:cacert], content: [int_cert, root_cert]},
+                         { location: puppet.settings[:cakey], content: root_key },
+                         { location: puppet.settings[:intkey], content: int_key },
```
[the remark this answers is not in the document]
amysahli: Yep.  Thanks!  According to Justin I'll want the root's private key too and figure out a place to put that.  I was thinking $cadir/root_key.pem ?  Thoughts?

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,250 @@
...
+
+module Puppetserver
+  module Ca
+    class GenerateAction
+      CA_EXTENSIONS = [
+        ["basicConstraints", "CA:TRUE", true],
+        ["keyUsage", "keyCertSign, cRLSign", true],
+        ["subjectKeyIdentifier", "hash", false],
+        ["authorityKeyIdentifier", "keyid:always", false]
+      ]
```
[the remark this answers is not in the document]
amysahli: I copied these from @justinstoller https://github.com/puppetlabs/puppetserver/blob/a345d592eeee881bf8d1f4265239eeb65d5fae18/acceptance/lib/puppet_x/acceptance/pki.rb#L9  But I don't really understand then enough to say how correct/tested they are.  I'm assuming since Justin wrote them they are PERFECT :)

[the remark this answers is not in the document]
justinstoller: https://www.ibm.com/support/knowledgecenter/en/SSB23S_1.1.0.13/gtps7/cfgcert.html
and
https://security.stackexchange.com/questions/27797/what-damage-could-be-done-if-a-malicious-certificate-had-an-identical-subject-k
for reference

lib/puppetserver/utils/file_utils.rb
```diff
@@ -0,0 +1,58 @@
...
+          if pe_puppet_exists?
+            return 'pe-puppet', 'pe-puppet'
+          else
+            return 'puppet', 'puppet'
+          end
+        end
+      end
+
+      def self.running_as_root?
+        !Gem.win_platform? && Process.euid == 0
```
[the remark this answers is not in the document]
amysahli: There is.  I thought may be better to do that in the refactoring ticket (when I use a lot of these methods too)

lib/puppetserver/settings/ttl_setting.rb
```diff
@@ -0,0 +1,48 @@
+# A setting that represents a span of time to live, and evaluates to Numeric
```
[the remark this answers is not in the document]
amysahli: @Magisus not sure I understand your comment.  Is this something I can change now?  Yes, its copied (parts) from what we currently do in puppet repo

lib/puppetserver/ca/generate_action.rb
amysahli: ok.  I see that ( @content.last_update = time - 1)  I'll do the same.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,250 @@
...
+          extension = ef.create_extension(*ext)
+          cert.add_extension(extension)
+        end
+
+        cert.sign(ca_key, DEFAULT_SIGNING_DIGEST)
+
+        cert
+      end
+
+      def ensure_file(location, content, mode)
```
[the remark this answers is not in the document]
amysahli: moved.  I had a reason to keep it here...which must have resolved itself ;)  Thanks!

lib/puppetserver/settings/ttl_setting.rb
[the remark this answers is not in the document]
Magisus: No I think this is fine for now. I was just curious if you got it from somewhere, because this kind of parsing can be tricky, and I didn't want us to reimplement from scratch if we can copy from something that has been verified.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,250 @@
...
+
+        # Generate root and intermediate ca and put all the certificates, crls,
+        # and keys where they should go.
+        generate_root_and_intermediate_ca(puppet)
+
+        # Puppet's internal CA expects these file to exist.
+        ensure_file(puppet.settings[:serial], "001", 0640)
+        ensure_file(puppet.settings[:cert_inventory], "", 0640)
+
+        return 0
```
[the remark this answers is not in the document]
amysahli: I can add a @logger.inform  or just a puts.  Not sure about what we prefer.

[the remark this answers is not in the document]
Magisus: Use the logger. It probably provides options for people to redirect it to a file and whatnot.

lib/puppetserver/utils/file_utils.rb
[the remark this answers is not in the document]
Magisus: I'd say anything that is already here as of this PR, you might as well clean up now. No one else is making changes to that part of the settings classes, so moving it probably won't cause too many conflicts.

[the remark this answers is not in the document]
amysahli: OK.  Will do it now :)

spec/puppetserver/ca/cli_spec.rb
```diff
@@ -335,4 +342,43 @@
...
+      expect(stderr.string).to match(usage)
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not print the help output if called correctly' do
+      Dir.mktmpdir do |tmpdir|
+        with_temp_cadir tmpdir do |conf|
+          exit_code = Puppetserver::Ca::Cli.run(['generate', '--config', conf], stdout, stderr)
+          expect(stderr.string).to be_empty
+          expect(stdout.string.strip).to eq("Generation succeeded. Find your files in #{tmpdir}/ca")
```
amysahli: ^^ I had to add .strip to get this to work.  I guess thats expected for stdout.string to add the space at the end??  I looked at the output and didnt see any issues.

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -75,19 +74,29 @@ def resolve_settings(overrides = {})
...
-        settings[:cacert] = overrides.fetch(:cacert, '$cadir/ca_crt.pem')
-        settings[:cakey] = overrides.fetch(:cakey, '$cadir/ca_key.pem')
-        settings[:cacrl] = overrides.fetch(:cacrl, '$cadir/ca_crl.pem')
-        settings[:serial] = overrides.fetch(:serial, '$cadir/serial')
+        # Note the call to hostname below, (tbd from Justin's code)
+        hostname = overrides.fetch(:certname, `hostname`.chomp)
+        settings[:certname] = substitutions['$certname'] = hostname.sub(unresolved_setting, substitutions)
+
+        settings[:ca_name] =        overrides.fetch(:ca_name, 'Puppet CA: $certname')
+        settings[:root_ca_name] =   overrides.fetch(:root_ca_name, 'Puppet Root CA: $certname')
```
amysahli: Thoughts on names?  Not sure if 'Puppet Root CA: $certname' is going to cut it.

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -75,19 +74,29 @@ def resolve_settings(overrides = {})
...
+        settings[:ca_ttl] =         overrides.fetch(:ca_ttl, '15y')
+        settings[:keylength] =      overrides.fetch(:keylength, 4096).to_s
+
+        settings[:cacert] =         overrides.fetch(:cacert, '$cadir/ca_crt.pem')
+        settings[:cakey] =          overrides.fetch(:cakey, '$cadir/ca_key.pem')
+        settings[:rootkey] =        overrides.fetch(:rootkey, '$cadir/root_key.pem')
```
amysahli: Does root_key.pem seem reasonable?

Magisus: Yeah

lib/puppetserver/ca/generate_action.rb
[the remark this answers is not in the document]
amysahli: Updated.  root CA to 1, intermediate CA to 2, and 1 for serial file.

lib/puppetserver/ca/puppet_config.rb
Magisus: Any reason why not? (I don't really have a good feel for any conventions that might exist around this)

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -75,19 +74,29 @@ def resolve_settings(overrides = {})
...
-        settings[:cacert] = overrides.fetch(:cacert, '$cadir/ca_crt.pem')
-        settings[:cakey] = overrides.fetch(:cakey, '$cadir/ca_key.pem')
-        settings[:cacrl] = overrides.fetch(:cacrl, '$cadir/ca_crl.pem')
-        settings[:serial] = overrides.fetch(:serial, '$cadir/serial')
+        # Note the call to hostname below, (tbd from Justin's code)
+        hostname = overrides.fetch(:certname, `hostname`.chomp)
+        settings[:certname] = substitutions['$certname'] = hostname.sub(unresolved_setting, substitutions)
+
+        settings[:ca_name] =        overrides.fetch(:ca_name, 'Puppet RootCA: $certname')
+        settings[:int_ca_name] =    overrides.fetch(:int_ca_name, 'Puppet CA: $certname')
```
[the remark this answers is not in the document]
amysahli: done.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        # Load, resolve, and validate puppet config settings
+        puppet = PuppetConfig.parse(config_path)
+        log_errors_and_return(puppet.errors) if puppet.errors.any?
+
+        # Generate root and intermediate ca and put all the certificates, crls,
+        # and keys where they should go.
+        generate_root_and_intermediate_ca(puppet)
+
+        # Puppet's internal CA expects these file to exist.
+        Puppetserver::Utils::FileUtilities.ensure_file(puppet.settings[:serial], "001", 0640)
```
Magisus: I think we want this to be "003", since the two other certificates have serials 1 and 2.

lib/puppetserver/ca/puppet_config.rb
amysahli: Well i have absolutely no idea ;)

lib/puppetserver/ca/generate_action.rb
amysahli: Talked with Justin, and because this is what the intermediate CA will use to sign certs, it should be 1.  (Since this is what we will look at when the intermediate signs its FIRST certificate).  The root gets a 1 since it signed itself and the intermediate gets a 2 since its the second cert the root signed. Does that sound right?

lib/puppetserver/utils/file_utils.rb
```diff
@@ -0,0 +1,64 @@
...
+      end
+
+      def self.ensure_file(location, content, mode)
+        if !File.exist?(location)
+          Puppetserver::Utils::FileUtilities.write_file(location, content, mode)
+        end
+      end
+
+      def self.validate_file_paths(one_or_more_paths)
+        return unless one_or_more_paths.any?
```
Magisus: You need to return an empty array here, we expect this method to not return nil.

amysahli: Thanks!  Added that and a spec that should have been there to catch that.

spec/puppetserver/ca/cli_spec.rb
Magisus: It probably adds a newline.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def run(input)
+        # Validate config_path provided
+        config_path = input['config']
+        errors = Puppetserver::Utils::FileUtilities.validate_file_paths([config_path])
+        log_errors_and_return(errors) if errors.any?
```
justinstoller: Oh, cool! I thought this would return an `Enumerator`.

lib/puppetserver/utils/file_utils.rb
```diff
@@ -0,0 +1,64 @@
...
+      def self.ensure_file(location, content, mode)
+        if !File.exist?(location)
+          Puppetserver::Utils::FileUtilities.write_file(location, content, mode)
+        end
+      end
+
+      def self.validate_file_paths(one_or_more_paths)
+        return [] unless one_or_more_paths.any?
+        errors = []
+        Array(one_or_more_paths).each do |path|
```
justinstoller: If we're going to require that the input to the method be an array we can change the name of the variable and remove the coercion with Array here.

If we're going to allow a single string path (previous functionality), then we should either remove the call to any? above or wrap it in a coercion as well.

I'm not sure I think the way this changed from the previous implementation - from allowing a single string path or (a potentially empty) array of string paths, to requiring an array of either string paths or falsey values - is an improvement.

I'm fine requiring the caller to provide an array of values, but I'm concerned about requiring it to be either all valid string paths, or all falsey values. I'd prefer to require the caller to do its nil checks as a mix of valid paths and nils will cause issues below.

Or, potentially, allow a single nil value (remove the any? from the unless conditional)?
I guess, calling compact would be a way to allow for the caller to not care about a mix of nils and file paths. But I still think that it would be more supportable long term to require the caller to do its nil checks (like I would expect a nil value passed in here to cause an error).

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+BANNER
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def run(input)
+        # Validate config_path provided
+        config_path = input['config']
+        errors = Puppetserver::Utils::FileUtilities.validate_file_paths([config_path])
```
justinstoller: Sorry, I wrote a novella in the definition of validate_file_paths. tl;dr, I think it would be better to have validate_file_paths no accept nil values at all and wrap this validation in an `if config_path` block.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        return 0
+      end
+
+      def log_errors_and_return(errors)
+        errors = Array(errors).compact
+        @logger.err "Error:"
+        errors.each do |message|
+          @logger.err "    #{message}"
+        end
+        return 1
```
justinstoller: This will return 1 to the caller, the `run` method above, however `run` doesn't do anything with the returned number. ie. this return won't jump out of the enclosing method. I think you probably want to change the name of the method and do something like this in the caller:
```rb
  return log_errors_and_return(errors) if errors.any?
```

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        errors = Puppetserver::Utils::FileUtilities.validate_file_paths([config_path])
+        log_errors_and_return(errors) if errors.any?
+
+        # Load, resolve, and validate puppet config settings
+        puppet = PuppetConfig.parse(config_path)
+        log_errors_and_return(puppet.errors) if puppet.errors.any?
+
+        # Generate root and intermediate ca and put all the certificates, crls,
+        # and keys where they should go.
+        generate_root_and_intermediate_ca(puppet)
```
justinstoller: from what I can tell the `generate_root_and_intermediate_ca` method doesn't need the actual puppet object just the settings from it. It would probably be cleaner to pass `puppet.settings` into this method and then have something like:
```rb
  def generate_root_and_intermediate_ca(settings)
    private_key_length = settings[:keylength]
    # ...
```

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        root_crl = create_crl_for(root_cert, root_key, ca_ttl)
+
+        int_key = create_private_key(private_key_length)
+        int_csr = create_csr(int_key, int_ca_name)
+        int_cert = sign(root_key, root_cert, int_csr, ca_ttl)
+        int_crl = create_crl_for(int_cert, int_key, ca_ttl)
+
+        Puppetserver::Utils::FileUtilities.ensure_dir(puppet.settings[:cadir])
+
+        future_files = [ { location: puppet.settings[:cacert], content: [int_cert, root_cert]},
```
justinstoller: Since we don't use future_files anywhere else we could iterate over the array of them directly, don't know if that's less readable though?

lib/puppetserver/utils/file_utils.rb
```diff
@@ -0,0 +1,64 @@
...
+        !!(Etc.getpwnam('pe-puppet') rescue nil)
+      end
+
+      def self.write_file(path, one_or_more_objects, mode)
+        File.open(path, 'w', mode) do |f|
+          Array(one_or_more_objects).each do |object|
+            f.puts object.to_s
+          end
+        end
+        user, group = find_user_and_group
```
justinstoller: This is a potentially expensive method† that we call every time someone calls write_file. I think it would make sense to either have the caller call find_user_and_group, save those values, and then pass them into {ensure,write}_file, or make this a utility class that saves the user and group as some state within it.

† Well, pe_puppet_exists? raises and rescues if the pe-puppet user doesn't exist, which doesn't take a lot of time but is a pretty bad practice and I don't know a better way to do it considering Etc's interface

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -75,19 +74,28 @@ def resolve_settings(overrides = {})
...
-        settings[:cacert] = overrides.fetch(:cacert, '$cadir/ca_crt.pem')
-        settings[:cakey] = overrides.fetch(:cakey, '$cadir/ca_key.pem')
-        settings[:cacrl] = overrides.fetch(:cacrl, '$cadir/ca_crl.pem')
-        settings[:serial] = overrides.fetch(:serial, '$cadir/serial')
+        # Note the call to hostname below, (tbd from Justin's code)
+        settings[:certname] = substitutions['$certname'] = overrides.fetch(:certname, `hostname`.chomp)
+
+        settings[:ca_name] =        overrides.fetch(:ca_name, 'Puppet CA: $certname')
+        settings[:root_ca_name] =   overrides.fetch(:root_ca_name, 'Puppet Root CA: $certname')
+        settings[:ca_ttl] =         overrides.fetch(:ca_ttl, '15y')
```
justinstoller: I made this comment in my PR as well. But I think that this has gotten big enough that we should look into using a hash of defaults and merging them with the overrides.

lib/puppetserver/ca/generate_action.rb
amysahli: Yup.  Totally didnt catch that.  Thanks!

lib/puppetserver/ca/generate_action.rb
amysahli: Duh :(  Thank you!

lib/puppetserver/ca/generate_action.rb
amysahli: I'll play around with it and see what I can do.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+                         { location: puppet.settings[:cakey], content: int_key },
+                         { location: puppet.settings[:rootkey], content: root_key },
+                         { location: puppet.settings[:cacrl], content: [int_crl, root_crl]}
+                       ]
+
+        future_files.each do | file|
```
justinstoller: Slightly less readable, but more concise would be to do an array of arrays and then destructure in the argument list, eg:
```rb
[
  [settings[:cacert], [int_cert, root_cert]],
  # ...
].each do |(location, content)|
  @logger.warn("#{location} exists, overwriting") if File.exists?(location)
  #...
```

I'm torn on if the brevity is worth the tradeoff in readability though (I might be more used to those kinds of destructurings because of working in clojure).

lib/puppetserver/ca/generate_action.rb
amysahli: done!  Thanks

lib/puppetserver/utils/file_utils.rb
amysahli: fixed! thanks

lib/puppetserver/ca/generate_action.rb
amysahli: Im fine with that.  I think its perfectly readable.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        cert.not_before = Time.now - (60*60*24)
+        cert.not_after  = Time.now + ca_ttl
+
+        ef = extension_factory_for(cert, cert)
+        CA_EXTENSIONS.each do |ext|
+          extension = ef.create_extension(*ext)
+          cert.add_extension(extension)
+        end
+
+        cert.sign(key, DEFAULT_SIGNING_DIGEST)
```
justinstoller: The test has a constant for default signing digest, and for testing we never ended up needing to parameterize that, however Puppet [attempts several different digests](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/certificate_signer.rb#L12-L28), using the most secure one it can load and we should do that here as well.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+          OpenSSL::X509::Extension.new("crlNumber", OpenSSL::ASN1::Integer(0)))
+
+        crl.last_update = Time.now - 1
+        crl.next_update = Time.now + ca_ttl
+        crl.sign(ca_key, DEFAULT_SIGNING_DIGEST)
+
+        crl
+      end
+
+      def create_csr(key, name)
```
justinstoller: @Magisus  I wanted to make sure that on your road map was support for [custom cert extensions](https://puppet.com/docs/puppet/5.5/ssl_attributes_extensions.html#concept-932) (if/when we implement signing certs with SANs). I assume we shouldn't do it now, but should we make a ticket to include custom cert extensions during bootstrapping with this CLI? (does Puppet respect them when bootstrapping currently?)

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+
+        csr.public_key = key.public_key
+        csr.subject = OpenSSL::X509::Name.new([["CN", name]])
+        csr.version = 2
+        csr.sign(key, DEFAULT_SIGNING_DIGEST)
+
+        csr
+      end
+
+      def sign(ca_key, ca_cert, csr, ca_ttl)
```
justinstoller: Since we hard code the serial number to be two it should only be used to sign the intermediate CA. Since it's not a generic sign method we should make that fact more explicit in the method name.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        cert.version = 2
+        cert.serial = 1
+
+        # Make the certificate valid as of yesterday, because so many people's
+        # clocks are out of sync.  This gives one more day of validity than people
+        # might expect, but is better than making every person who has a messed up
+        # clock fail, and better than having every cert we generate expire a day
+        # before the user expected it to when they asked for "one year".
+        cert.not_before = Time.now - (60*60*24)
+        cert.not_after  = Time.now + ca_ttl
```
justinstoller: We do this when creating the root cert but the CRL is set to one second in the past and the intermediate cert is just set to `Time.now`. Can we save these values and then use them everywhere? Do the Int. cert & CRLs need to have a not_before time later than the issuing cert?

lib/puppetserver/ca/import_action.rb
```diff
@@ -1,6 +1,6 @@
 require 'etc'
```
justinstoller: you can remove this too

lib/puppetserver/ca/import_action.rb
```diff
@@ -47,65 +47,22 @@ def run(input)
...
-
-        if !File.exist?(puppet.settings[:cert_inventory])
-          write_file(puppet.settings[:cert_inventory],
-                     "", user, group, 0640)
-        end
+        # Puppet's internal CA expects these file to exist.
+        Puppetserver::Utils::FileUtilities.ensure_file(puppet.settings[:serial], "001", 0640)
+        Puppetserver::Utils::FileUtilities.ensure_file(puppet.settings[:cert_inventory], "", 0640)
 
+        @logger.inform "Import succeeded. Find your files in #{puppet.settings[:cadir]}"
```
justinstoller: Nice addition

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        return parser, parsed, unparsed
+      end
+
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--help', 'Display this generate specific help output') do |help|
+            parsed['help'] = true
+          end
+          opts.on('--version', 'Output the version') do |v|
```
justinstoller: The fact that Import has a version flag is a bug. The main Cli class will parse version and return the gem's version. Individual actions don't have specific versions and we should remove this and the version option declaration in Import as well (I'm not asking for you to remove Import's as part of this PR though).

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -75,19 +74,28 @@ def resolve_settings(overrides = {})
...
 
-        settings[:cacert] = overrides.fetch(:cacert, '$cadir/ca_crt.pem')
-        settings[:cakey] = overrides.fetch(:cakey, '$cadir/ca_key.pem')
-        settings[:cacrl] = overrides.fetch(:cacrl, '$cadir/ca_crl.pem')
-        settings[:serial] = overrides.fetch(:serial, '$cadir/serial')
+        # Note the call to hostname below, (tbd from Justin's code)
+        settings[:certname] = substitutions['$certname'] = overrides.fetch(:certname, `hostname`.chomp)
+
+        settings[:ca_name] =        overrides.fetch(:ca_name, 'Puppet CA: $certname')
+        settings[:root_ca_name] =   overrides.fetch(:root_ca_name, 'Puppet Root CA: $certname')
```
justinstoller: I don't think the root CA should include the certname. The certname will most likely be the hostname and we expect users to be able to extract the root's cert and key and use them elsewhere on different hosts. We might want to put in a random value here - if we were using UUIDs for serials that'd be perfect - since we aren't yet we could just put in a seven digit hex? or something? I don't have a good answer, and could see using certname, I guess... What do you think @Magisus ?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -75,19 +74,28 @@ def resolve_settings(overrides = {})
...
+        settings[:ca_ttl] =         overrides.fetch(:ca_ttl, '15y')
+        settings[:keylength] =      overrides.fetch(:keylength, 4096).to_s
```
justinstoller: Hrm... We make this a string, I assume, because the value from the puppet.conf will be read as a string and the substitution code below will barf if you call it on a number. But this means the consumer of this object needs to coerce it into a number before using it... What if we coerced this to a number here and then skipped non-string values in the substitution code below? (We could also then remove the call `to_i` in `GenerateAction#to_i`)

iow, I think it would be better for this to return usable settings and not have the consumer care about coercing.

lib/puppetserver/ca/puppet_config.rb
justinstoller: What if we called the TTLSetting class here and added any errors that it finds to this classes errors?

lib/puppetserver/ca/generate_action.rb
amysahli: There's an else statement returning  "No FIPS 140-2 compliant digest algorithm in OpenSSL::Digest".  Are we concerned about that?  Would we like to work that error message into the generate action?

lib/puppetserver/ca/generate_action.rb
amysahli: Im remove it from generate and import as well.

lib/puppetserver/ca/generate_action.rb
amysahli: I can set the ca_cert and int_cert to both use "yesterday".  The int_cert is always created after the ca_cert, so it will have a not_before of later than the issuing cert.  With the crl, I don't know.  I know we only do -1 second here: https://github.com/puppetlabs/puppet/blob/09bdfa9eaa368c241e68f01fb270b03aab6fe7b4/lib/puppet/ssl/certificate_revocation_list.rb#L88

lib/puppetserver/ca/generate_action.rb
```diff
@@ -0,0 +1,249 @@
...
+        # might expect, but is better than making every person who has a messed up
+        # clock fail, and better than having every cert we generate expire a day
+        # before the user expected it to when they asked for "one year".
+        cert.not_before = Time.now - (60*60*24)
+        cert.not_after  = Time.now + ca_ttl
+
+        ef = extension_factory_for(cert, cert)
+        CA_EXTENSIONS.each do |ext|
+          extension = ef.create_extension(*ext)
+          cert.add_extension(extension)
```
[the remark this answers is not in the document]
amysahli: Would I need to pull in the dns_alt_names value if its present in the config file?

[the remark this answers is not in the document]
Magisus: Yes I think so.

lib/puppetserver/ca/puppet_config.rb
Magisus: I'm a little curious how hard it would be to make the switch to using UUIDs as the serials. There are open tickets about it, and this could be our chance, but I don't really know what it would entail.

Magisus: In the meantime, a random number seems fine to me, if we think some uniqueness is needed. I don't really know what the CN is used for, in the context of the root.

amysahli: I can just as easily to a random 7 hex as a uuid.  Any preference?

lib/puppetserver/ca/puppet_config.rb
amysahli: OK.  Let's do this in the refactor ticket.

lib/puppetserver/ca/generate_action.rb
amysahli: ^ talked to Justin.  Will include this error if it happens

lib/puppetserver/ca/generate_action.rb
amysahli: ^^ talked to Justin and We'll use the same value for crl and certs.  Checked this and it doesn't seem to cause any problems with them all having the same last_updated time.

lib/puppetserver/ca/puppet_config.rb
amysahli: ^^ Talked to Justin and doing 7 hex random

lib/puppetserver/ca/generate_action.rb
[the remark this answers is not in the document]
amysahli: Made ticket to deal with this: https://tickets.puppetlabs.com/browse/SERVER-2269

lib/puppetserver/utils/file_utils.rb
```diff
@@ -0,0 +1,79 @@
+require 'fileutils'
+require 'etc'
+require 'pry'
```
Magisus: I think this got left in from debugging.

PR #5, opened 2018-07-24
(SERVER-2254) Allow the CLI to read from puppetserver's config

Magisus:
This commit adds a PuppetserverConfig class responsible for parsing
CA-related settings out of Puppetserver's CA config file. This file is
written in HOCON, so this commit pulls in the ruby-hocon gem for parsing
the file.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,3 +7,4 @@ gemspec
 
 gem 'pry'
 gem 'pry-byebug'
+gem 'hocon', '~> 1.2', require: false
--- a/lib/puppetserver/ca/puppet_config.rb
+++ b/lib/puppetserver/ca/puppet_config.rb
@@ -1,10 +1,13 @@
+require 'puppetserver/ca/config_utils'
 
 module Puppetserver
   module Ca
-    # Provides an interface for asking for Puppet[ Server] settings w/o loading
-    # either Puppet or Puppet Server. Includes a simple ini parser that will
-    # ignore Puppet's more complicated conventions.
+    # Provides an interface for asking for Puppet settings w/o loading
+    # Puppet. Includes a simple ini parser that will ignore Puppet's
+    # more complicated conventions.
     class PuppetConfig
+
+      include Puppetserver::Ca::ConfigUtils
 
       def self.parse(config_path = nil)
         instance = new(config_path)
@@ -119,15 +122,10 @@ module Puppetserver
         res
       end
 
-
      private
 
       def explicitly_given_config_file_or_default_config_exists?
         !@using_default_location || File.exist?(@config_path)
-      end
-
-      def running_as_root?
-        !Gem.win_platform? && Process::UID.eid == 0
       end
     end
   end
--- a/lib/puppetserver/ca/puppetserver_config.rb
+++ b/lib/puppetserver/ca/puppetserver_config.rb
@@ -1,0 +1,83 @@
+require 'hocon'
+require 'puppetserver/ca/config_utils'
+
+module Puppetserver
+  module Ca
+    # Provides an interface for querying Puppetserver settings w/o loading
+    # Puppetserver or any TK config service. Uses the ruby-hocon gem for parsing.
+    class PuppetserverConfig
+
+      include Puppetserver::Ca::ConfigUtils
+
+      def self.parse(config_path = nil)
+        instance = new(config_path)
+        instance.load
+
+        return instance
+      end
+
+      attr_reader :errors, :settings
+
+      def initialize(supplied_config_path = nil)
+        @using_default_location = !supplied_config_path
+        @config_path = supplied_config_path || "/etc/puppetlabs/puppetserver/conf.d/ca.conf"
+
+        @settings = nil
+        @errors = []
+      end
+
+      # Populate this config object with the CA-related settings
+      def load
+        if explicitly_given_config_file_or_default_config_exists?
+          begin
+            results = Hocon.load(@config_path)
+          rescue Hocon::ConfigError => e
+            errors << e.message
+          end
+        end
+
+        overrides = results || {}
+        @settings = supply_defaults(overrides).freeze
+      end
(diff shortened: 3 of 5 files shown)


lib/puppetserver/ca/puppet_config.rb
```diff
@@ -1,9 +1,9 @@
 
 module Puppetserver
   module Ca
-    # Provides an interface for asking for Puppet[ Server] settings w/o loading
-    # either Puppet or Puppet Server. Includes a simple ini parser that will
-    # ignore Puppet's more complicated conventions.
+    # Provides an interface for asking for Puppet settings w/o loading
+    # Puppet. Includes a simple ini parser that will ignore Puppet's
+    # more complicated conventions.
```
Magisus: I didn't see a lot of benefit to trying to combine this with the puppetserver version, or to use inheritance to relate them, but I'm open to ideas if someone thinks there'd be a better way to organize the two config classes.

Magisus:
Still need to test this manually with a real puppetserver.

lib/puppetserver/ca/puppetserver_config.rb
```diff
@@ -0,0 +1,84 @@
...
+        instance.load
+
+        return instance
+      end
+
+      attr_reader :errors, :settings
+
+      def initialize(supplied_config_path = nil)
+        @using_default_location = !supplied_config_path
+        @config_path = supplied_config_path || "/etc/puppetlabs/puppetserver/conf.d/ca.conf"
```
Magisus: Is there any use case where puppetserver's conf file might be in a user-local directory, besides running from source? (I think for that case we should specify `--config` manually)

Magisus:
Manually tested this by updating the following:
```
diff --git a/lib/puppetserver/ca/setup_action.rb b/lib/puppetserver/ca/setup_action.rb
index 1ccb91e..c9fd7be 100644
--- a/lib/puppetserver/ca/setup_action.rb
+++ b/lib/puppetserver/ca/setup_action.rb
@@ -2,7 +2,7 @@ require 'etc'
 require 'fileutils'
 require 'optparse'
 require 'puppetserver/ca/x509_loader'
-require 'puppetserver/ca/puppet_config'
+require 'puppetserver/ca/puppetserver_config'

 module Puppetserver
   module Ca
@@ -44,7 +44,7 @@ BANNER
         loader = X509Loader.new(bundle_path, key_path, chain_path)
         return 1 if log_possible_errors(loader.errors)

-        puppet = PuppetConfig.parse(config_path)
+        puppet = PuppetserverConfig.parse(config_path)
         return 1 if log_possible_errors(puppet.errors)

         user, group = find_user_and_group
```

I set up puppetserver running from source from [Amy's branch](https://github.com/puppetlabs/puppetserver/pull/1728) with the new settings and manually pointed the tool at the dev config file. I was able to properly install a CA in the new location using the `setup` command.

lib/puppetserver/ca/puppetserver_config.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      private
+
+      # Return the correct confdir. We check for being root on *nix,
+      # else the user path. We do not include a check for running
+      # as Adminstrator since non-development scenarios for Puppet Server
+      # on Windows are unsupported.
+      # Note that Puppet Server runs as the [pe-]puppet user but to
+      # start/stop it you must be root.
+      def user_specific_root_dir
```
justinstoller: can we return the cadir rather than the root dir?

lib/puppetserver/ca/puppetserver_config.rb
```diff
@@ -0,0 +1,84 @@
...
+        settings[:cert_inventory] = ca_settings.fetch('cert_inventory', "#{cadir}/inventory.txt")
+
+        return settings
+      end
+
+      def explicitly_given_config_file_or_default_config_exists?
+        !@using_default_location || File.exist?(@config_path)
+      end
+
+      def running_as_root?
```
justinstoller: maybe put into a util module and share it between the two config loaders?

lib/puppetserver/ca/puppetserver_config.rb
```diff
@@ -0,0 +1,84 @@
...
+        ca_settings = overrides['certificate-authority'] || {}
+        settings = {}
+
+        cadir = settings[:cadir] = ca_settings.fetch('cadir', "#{user_specific_root_dir}/ca")
+
+        settings[:cacert] = ca_settings.fetch('cacert', "#{cadir}/ca_crt.pem")
+        settings[:cakey] = ca_settings.fetch('cakey', "#{cadir}/ca_key.pem")
+        settings[:cacrl] = ca_settings.fetch('cacrl', "#{cadir}/ca_crl.pem")
+        settings[:serial] = ca_settings.fetch('serial', "#{cadir}/serial")
+        settings[:cert_inventory] = ca_settings.fetch('cert_inventory', "#{cadir}/inventory.txt")
```
justinstoller: we kebab cased this when moving it to hocon, iiuc.

Magisus:
Updated.

Gemfile
```diff
@@ -7,3 +7,4 @@ gemspec
 
 gem 'pry'
 gem 'pry-byebug'
+gem 'hocon', '~> 1.2', require: false
```
lazartea: What does ``require: false`` do?

justinstoller: When you call Bundler.setup it will attempt to autorequire all your gems. We like to manually require them to ensure our classes that depend on something do it explicitly. We don't do it above so when debugging (we only do Bundle.setup in dev environments) we can easily drop pry statements in there.

See https://bundler.io/man/gemfile.5.html#REQUIRE-AS and our usage in
https://github.com/puppetlabs/puppetserver-ca-cli/blob/master/exe/puppetserver-ca#L6

lazartea: thanks!

PR #7, opened 2018-07-30
(SERVER-1714) Add sign action

lazartea:
This is still a work in progress, but I thought it might be nice to get some feedback. The main chunk of work is the logic in ``puppetserver/ca/sign_action.rb`` differentiating between use with and without the ``--all`` flag. I'm not sure all of the parsing logic stands, but I've tested it manually in the irb and I can confirm that it's signing certs. The rest is just getting it pretty/efficient, and I'm starting to write the spec.

### DIFF
--- a/lib/puppetserver/ca/cli.rb
+++ b/lib/puppetserver/ca/cli.rb
@@ -4,6 +4,8 @@ require 'puppetserver/ca/logger'
 require 'puppetserver/ca/import_action'
 require 'puppetserver/ca/generate_action'
 require 'puppetserver/ca/revoke_action'
+require 'puppetserver/ca/sign_action'
+require 'puppetserver/ca/utils'
 
 module Puppetserver
   module Ca
@@ -18,7 +20,8 @@ BANNER
       VALID_ACTIONS = {
         'import'   => ImportAction,
         'generate' => GenerateAction,
-        'revoke'   => RevokeAction
+        'revoke'   => RevokeAction,
+        'sign'     => SignAction
       }
 
       ACTION_LIST = "\nAvailable Actions:\n" +
@@ -94,19 +97,9 @@ BANNER
 
         end
 
-        unparsed, nonopts = [], []
+        all,_,_,_ = Utils.parse_without_raising(general_parser, inputs)
 
-        begin
-          general_parser.order!(inputs) do |nonopt|
-            nonopts << nonopt
-          end
-        rescue OptionParser::InvalidOption => e
-          unparsed += e.args
-          unparsed << inputs.shift unless inputs.first =~ /^-{1,2}/
-          retry
-        end
-
-        return general_parser, parsed, nonopts + unparsed
+        return general_parser, parsed, all
       end
     end
   end
--- a/lib/puppetserver/ca/sign_action.rb
+++ b/lib/puppetserver/ca/sign_action.rb
@@ -1,0 +1,181 @@
+require 'puppetserver/ca/utils'
+require 'puppetserver/utils/http_utilities'
+require 'puppetserver/utils/file_utilities'
+require 'puppetserver/ca/puppet_config'
+require 'optparse'
+require 'openssl'
+require 'net/https'
+require 'json'
+
+module Puppetserver
+  module Ca
+    class SignAction
+
+      include Puppetserver::Utils
+
+      SUMMARY = 'Sign a given certificate'
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca sign [--help|--version]
+  puppetserver ca sign [--config] --certname CERTNAME[,CERTNAME]
+  puppetserver ca sign  --all
+
+Description:
+Given a comma-separated list of valid certnames, instructs the CA to sign each cert.
+
+Options:
+      BANNER
+      BODY = JSON.dump({desired_state: 'signed'})
+
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--certname x,y,z', Array, 'the name(s) of the cert(s) to be signed') do |cert|
+            parsed['certname'] = cert
+          end
+          opts.on('--config PUPPET.CONF', 'Custom path to Puppet\'s config file') do |conf|
+            parsed['config'] = conf
+          end
+          opts.on('--help', 'Display this command specific help output') do |help|
+            parsed['help'] = true
+          end
+          opts.on('--version', 'Output the version') do |v|
+            parsed['version'] = true
+          end
+          opts.on('--all', 'Operate on all certnames') do |a|
+            parsed['all'] = true
+          end
+        end
+      end
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def run(input)
+        config = input['config']
+
+        if config
+          errors = FileUtilities.validate_file_paths(config)
+          return 1 if Utils.handle_errors(@logger, errors)
+        end
+
+        puppet = PuppetConfig.parse(config)
+        return 1 if Utils.handle_errors(@logger, puppet.errors)
+
+        if input['all']
+          requested_certnames = get_all_pending_certs(puppet.settings)
+          if requested_certnames.nil?
+            return 1
+          end
+        else
+          requested_certnames = input['certname']
+        end
+
+        success = sign_requested_certs(requested_certnames, puppet.settings)
+        return success ? 0 : 1
+      end
+
+      def get_certificate_statuses(settings)
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_statuses', 'any_key')
+        HttpUtilities.with_connection(url, settings) do |connection|
+          connection.get(url)
+        end
+      end
+
+      def sign_certs(certnames,settings)
+        results = { }
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_status')
+        HttpUtilities.with_connection(url, settings) do |connection|
+          certnames.each do |certname|
+            url.resource_name = certname
+            results[certname] = connection.put(BODY, url)
+          end
+        end
+        return results
+      end
+
+      def get_all_certs(settings)
+        result = get_certificate_statuses(settings)
+
+        unless result.is_a? Net::HTTPOK
+            @logger.err 'Error:'
+            @logger.err "    #{result.inspect}"
+            return nil
+        end
+        return result
+      end
+
+      def select_pending_certs(get_result)
+        requested_certnames = JSON.parse(get_result).select{|e| e["state"] == "requested"}.map{|e| e["name"]}
+
+        if requested_certnames.empty?
+          @logger.err 'Error:'
+          @logger.err "    No waiting certificate requests to sign"
+          return nil
+        end
+
+        return requested_certnames
+      end
+
+      def get_all_pending_certs(settings)
+        result = get_all_certs(settings)
+        if result
+          select_pending_certs(result.body)
+        end
+      end
+
+      def sign_requested_certs(certnames,settings)
+        success = true
+        results = sign_certs(certnames, settings)
+        results.each do |certname, result|
+          case result.code
+          when '204'
+            @logger.inform "Signed certificate for #{certname}"
+          when '404'
+            @logger.err 'Error:'
+            @logger.err "    Could not find certificate for #{certname}"
+            success = false
+          else
+            @logger.err 'Error:'
+            @logger.err "    When download requested for #{result.inspect}"
+            @logger.err "    code: #{result.code}"
+            @logger.err "    body: #{result.body_to_s}" if result.body
+            success = false
+          end
+        end
+        return success
+      end
+
+      def check_flag_usage(results)
+        if results['certname'] && results['all']
+          '--all and --certname cannot be used together'
+        elsif !results['certname'] && !results['all']
+          'No arguments given'
+        elsif results['certname'] && results['certname'].include?('--all')
+          'Cannot use --all with --certname. If you actually have a certificate request ' +
+                          'for a certifcate named --all, you need to use the HTTP API.'
+        end
+      end
+
+      def parse(args)
+        results = {}
+        parser = self.class.parser(results)
+
+        errors = Utils.parse_with_errors(parser, args)
+
+        if check_flag_usage(results)
+          errors << check_flag_usage(results)
+        end
+
+        errors_were_handled = Utils.handle_errors(@logger, errors, parser.help)
+
+        exit_code = errors_were_handled ? 1 : nil
+
+        return results, exit_code
+      end
+    end
+  end
+end
--- a/spec/puppetserver/ca/sign_action_spec.rb
+++ b/spec/puppetserver/ca/sign_action_spec.rb
@@ -1,0 +1,89 @@
+require 'puppetserver/ca/sign_action'
+require 'puppetserver/ca/cli'
+require 'puppetserver/ca/logger'
+
+RSpec.describe 'Puppetserver::Ca::SignAction' do
+  let(:err)    { StringIO.new }
+  let(:out) { StringIO.new }
+  let(:logger) { Puppetserver::Ca::Logger.new(:info, out, err) }
+  let(:action) { Puppetserver::Ca::SignAction.new(logger) }
+
+  describe 'validation' do
+    it 'does not expect an argument' do
+      _, exit_code = action.parse(['--all', 'foo'])
+      expect(err.string).to match(/Error:.*Unknown input.*foo/m)
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not allow --all with a valid certname call' do
+      _, exit_code = action.parse(['--certname', 'foo', '--all'])
+      expect(err.string).to include('--all and --certname cannot be used together')
+      expect(exit_code).to eq(1)
+    end
+
+    it 'errors without any arguments' do
+      _, exit_code = action.parse([])
+      expect(err.string).to include('No arguments given')
+      expect(exit_code).to eq(1)
+    end
+
+    it 'errors without a certname' do
+      _, exit_code = action.parse(['--certname'])
+      expect(err.string).to match(/Error:.*Missing argument to flag/m)
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not allow --certname and --all' do
+      _, exit_code = action.parse(['--certname', '--all'])
+      expect(err.string).to match(/Cannot use --all.*HTTP API/)
+      expect(exit_code).to eq(1)
+    end
+
+    it 'works with a single cert' do
+      results, exit_code = action.parse(['--certname', 'foo'])
+      expect(results['certname']).to eq(['foo'])
+      expect(exit_code).to eq(nil)
+    end
+
+    it 'works with a comma separated list of certs' do
+      results, exit_code = action.parse(['--certname', 'foo,bar,baz'])
+      expect(results['certname']).to eq(['foo','bar','baz'])
+      expect(exit_code).to eq(nil)
+    end
+  end
+
+  describe 'error handling' do
+    let(:response)  { Struct.new(:code, :body) }
+    let(:success)   { response.new('204', nil) }
+    let(:not_found) { response.new('404', 'Not Found') }
+    let(:empty)     { response.new('404', '[]') }
+
+    it 'logs and exits with zero with successful PUT' do
+      allow(action).to receive(:sign_certs).and_return({'foo' => success})
+      exit_code = action.run({'certname' => ['foo']})
+      expect(exit_code).to eq(0)
+      expect(out.string).to include('Signed certificate for foo')
+    end
+
+    it 'fails when PUT request errors' do
+      allow(action).to receive(:get_certificate_statuses).and_return(not_found)
+      exit_code = action.run({'all' => true})
+      expect(exit_code).to eq(1)
+    end
+
+    it 'fails when no pending certs' do
+      allow(action).to receive(:get_all_certs).and_return(empty)
+      exit_code = action.run({'all' => true})
+      expect(exit_code).to eq(1)
+      expect(err.string).to include('No waiting certificate requests to sign')
+    end
+
+    it 'continues signing certs after failed request' do
+      allow(action).to receive(:sign_certs).and_return({'foo' => success, 'bar' => not_found, 'baz' => success})
+      exit_code = action.run({'certname' => ['foo','bar','baz']})
+      expect(exit_code).to eq(1)
+      expect(out.string).to match(/Signed certificate for foo.*Signed certificate for baz/m)
+      expect(err.string).to include('Could not find certificate for bar')
+    end
+  end
+end
--- a/lib/puppetserver/utils/file_utilities.rb
+++ b/lib/puppetserver/utils/file_utilities.rb
@@ -1,6 +1,5 @@
 require 'fileutils'
 require 'etc'
-require 'pry'
 
 module Puppetserver
   module Utils
--- a/lib/puppetserver/utils/http_utilities.rb
+++ b/lib/puppetserver/utils/http_utilities.rb
@@ -52,6 +52,13 @@ module Puppetserver
 
           Result.new(result.code, result.body)
         end
+
+        def get(url_overide = nil)
+          url = url_overide || @url
+
+          request = Net::HTTP::Get.new(url.to_uri)
+          result = @conn.request(request)
+        end
       end
 
       # Just provide the bits of Net::HTTPResponse we care about
--- a/spec/puppetserver/ca/cli_spec.rb
+++ b/spec/puppetserver/ca/cli_spec.rb
@@ -27,6 +27,16 @@ RSpec.describe Puppetserver::Ca::Cli do
       /.*Usage:.* puppetserver ca revoke.*instructs the CA to revoke.*/m
   end
 
+  describe 'the sign action' do
+    let (:usage) do
+      /.*Usage:.* puppetserver ca sign.*Display this command specific help output.*/m
+    end
+
+    include_examples 'basic cli args',
+      'sign',
+      /.*Usage.* puppetserver ca sign.*Display this command specific help output.*/m
+  end
+
   describe 'the import action' do
     let(:usage) do
       /.*Usage:.* puppetserver ca import.*Display this import specific help output.*/m


lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,174 @@
...
+
+module Puppetserver
+  module Ca
+    class SignAction
+
+      SUMMARY = "Set up the CA's key, certs, and crls"
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca sign [--help|--version]
+  puppetserver ca sign [--all] HOSTNAME
```
lazartea: I don't think this is exactly right. How do I specify that ``HOSTNAME`` is optional (ie, we want it if there's no ``--all`` flag, but we don't want it if there's an ``--all`` flag)?

lazartea:
Travis is failing because I made a mistake when merging up the changes to ImportAction, which I'll fix when I push up my sign action rspec.

lib/puppetserver/ca/cli.rb
```diff
@@ -13,7 +14,8 @@ class Cli
 Puppet Server's built-in Certificate Authority
 BANNER
 
-      VALID_ACTIONS = {'import' => ImportAction}
+      VALID_ACTIONS = {'sign' => SignAction,
```
lazartea: need to fix the alignment here

lib/puppetserver/ca/sign_action.rb
Magisus: I think it might be more consistent with other commands in this gem to have the cert(s) be specified behind a flag, e.g. `puppetserver ca sign --certs certname1,certname2`.

Alternatively, the user can do `puppetserver ca sign --all`.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,174 @@
...
+            end
+            logger_output result
+            end
+          end
+
+        end
+      end
+
+      def logger_output(result)
+        if result.is_a? Net::HTTPNoContent
```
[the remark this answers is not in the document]
Magisus: I think this is something that is built into Puppetserver's UI and changing it is outside the scope of this ticket. I would say expect "NoContent" to indicate success here.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,181 @@
...
+        else
+          @logger.err 'Error - gross debug output:'
+          @logger.err "    #{result.inspect}"
+          return 1
+        end
+      end
+
+      def logger_output_get(result)
+        if result.is_a? Net::HTTPOK
+          return 0
```
lazartea: do we need to log that the get request was successful? does the user care about that?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,181 @@
...
+        elsif result.is_a? Net::HTTPNotFound
+          @logger.err 'Error'
+          @logger.err "   Could not find all certificates."
```
lazartea: this doesn't seem to be a very useful error message. How can I make it more clear?

lib/puppetserver/ca/sign_action.rb
Magisus: No, at least not until we implement a debug mode.

lib/puppetserver/ca/sign_action.rb
Magisus: Does `certificate_statuses` really return 404 sometimes? I would assume it would return some mostly empty JSON. The code seems to indicated that if there are no files on disk, it attempts to render an empty map of JSON and returns that. But you should verify if you can: start up puppetserver, delete all the things in `ca/signed` and `ca/requests` (note that this state is also highly unlikely, since you should at least always have the puppet master's cert in `signed`), and query the endpoint. If it does just return an empty map, I'd do less error handling here and just make sure your later code can handle getting no results.

lazartea: That's a really good point; I'll dig into it further. At the moment, I think things will break if we get back a map with no unsigned certs, since ``requestedCertnames`` would be set to nil. I'll add some logic to handle that.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,181 @@
...
+
+module Puppetserver
+  module Ca
+    class SignAction
+
+      SUMMARY = "Set up the CA's key, certs, and crls"
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca sign [--help|--version]
+  puppetserver ca sign  --certname <comma separated list of hostnames>
```
lazartea: Is this an appropriate way to indicate that it takes a list?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,181 @@
...
+          unparsed << inputs.shift unless inputs.first =~ /^-{1,2}/
+          retry
+        end
+
+        return parser, parsed, unparsed
+      end
+
+      def logger_output_put(result)
+        if result.is_a? Net::HTTPNoContent
+          @logger.inform "Signed Certificate for #{certname}"
```
lazartea: should we be waiting to return 0 when all certs have been signed? or keeping track of all successes based on the number of successes we expect?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+        end
+        return true
+      end
+
+      def valid?(invalid,unparsed,results,parser)
+        unless invalid.empty?
+          log_invalid_certname
+          return false
+        else
+           [check_flag_usage(results,parser), check_unparsed(unparsed,parser)].all?
```
lazartea: this is a little ugly, but the idea is that we want to return false if either ``check_flag_usage`` or ``check_unparsed`` is false, and we want to make sure to call both methods. We can't use ``and`` or ``&&``, as both use short circuit evaluation (``and`` just has lower precedence, it turns out).

justinstoller: The more I think about this the more I think the UX benefit isn't worth the weird code. What do you think?

lazartea: I think I agree... I'm not sure the intent/necessity behind this logic is coming through at all. I could put up a second commit with a simpler if/elsif/elsif check so we can compare the two?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+
+        if valid?(invalid,unparsed,results,parser)
+          exit_code = nil
+        else
+          exit_code = 1
+        end
+        return results, exit_code
+      end
+
+      def parse_inputs(inputs)
```
justinstoller: what's this guy doing?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+        end
+      end
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def run(input)
+        if input['all']
+          requestedCertnames = get_all_pending_certs
```
justinstoller: ruby variables should be `snake_case` not `cameCase`

lib/puppetserver/ca/sign_action.rb
lazartea: nothing, apparently  😂I think it's left over from refactoring/moving things around. Good catch!

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+          result = sign_cert(certname)
+
+          if result.is_a? Net::HTTPNoContent
+            @logger.inform "Signed Certificate for #{certname}"
+          elsif result.is_a? Net::HTTPNotFound
+            @logger.err 'Error'
+            @logger.err "    Could not find certificate for #{certname}"
+            exit_code = 1
+          else
+            @logger.err 'Error - gross debug output:'
```
justinstoller: We probably don't want to actually print this bit about gross debug output in the final code

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+          select_pending_certs(result)
+        end
+      end
+
+      def sign_requested_certs(certnames)
+        exit_code = 0
+        certnames.each do |certname|
+          result = sign_cert(certname)
+
+          if result.is_a? Net::HTTPNoContent
```
justinstoller: @amysahli suggested in my PR, and I think it worked out a little cleaner, to use case instead here: https://github.com/puppetlabs/puppetserver-ca-cli/pull/3/files#diff-12d66b569592783ab1bdd585b3c88581R135
what do you think?

justinstoller: @pcarlisle would you put the result handling here in its own method? I'm wondering how thinly different folks would slice these abstractions.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+        @logger.err 'Unknown arguments or flags:'
+        unparsed.each do |arg|
+          @logger.err "    #{arg}"
+        end
+
+        @logger.err ''
+        @logger.err parser.help
+      end
+
+      def log_invalid_certname()
```
justinstoller: don't need parens after a method w/o args (either defining or calling)

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,252 @@
...
+          end
+        else
+          requestedCertnames = input['certname']
+        end
+
+        exit_code = sign_requested_certs(requestedCertnames)
+        return exit_code
+      end
+
+      def logger_output_put(result,certname)
```
justinstoller: all of the argument lists and arrays should have a space after the commas

justinstoller: Also, I don't think this method is used at all anymore

lazartea: I think you're right. I think ``sign_request_certs`` is covering that error handling now.

lib/puppetserver/ca/sign_action.rb
lazartea: yep, definitely agree. I'll change that.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+        else
+          requested_certnames = input['certname']
+        end
+
+        exit_code = sign_requested_certs(requested_certnames)
+        return exit_code
+      end
+
+      def get_certificate_statuses
+        uri = URI('https://localhost:8140/puppet-ca/v1/certificate_statuses/:any_key?environment=production')
```
amysahli: Is there no way to explicitly request certs with state of "requested" ?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+      end
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def run(input)
+        if input['all']
+          requested_certnames = get_all_pending_certs
+          if requested_certnames.nil?
```
amysahli: Let's add an error message here if they requested ```--all``` but there aren't any cert to sign.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+        end
+        return result
+      end
+
+      def select_pending_certs(get_result)
+        requested_certnames = JSON.parse(get_result.body).select{|e| e["state"] == "requested"}.map{|e| e["name"]}
+
+        if requested_certnames.empty?
+          @logger.err 'Error'
+          @logger.err "No waiting certificate requests to sign"
```
amysahli: sorry, just saw this error here.

lib/puppetserver/ca/sign_action.rb
lazartea: I don't think so, or at least, I didn't see anything about it [here](https://puppet.com/docs/puppet/5.3/http_api/http_certificate_status.html).

justinstoller: Sadly, [no](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1210-L1235).

I'm not finding the ticket but there's a feature request for server side filtering of statuses which would be great as an optimization for this afterwards.

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'puppetserver/ca/sign_action'
+require 'puppetserver/ca/cli'
+require 'puppetserver/ca/logger'
+
+RSpec.describe 'Puppetserver::Ca::SignAction' do
+  let(:err)    { StringIO.new }
+  let(:inform) { StringIO.new }
```
justinstoller: This is a mock stdout, and the above is stderr (or more shortly `err` and `out`). `Logger#inform` goes to stdout, but I don't think we should name this after the logging method.

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    before do
+      allow_any_instance_of(Puppetserver::Ca::SignAction).to receive(:run).
+          and_raise("Validation succeeded when it should have failed")
+    end
+
+    it 'does not expect an argument' do
+      exit_code = Puppetserver::Ca::Cli.run(['sign','--all','foo'],STDOUT, err)
+      expect(err.string).to match(/Error:.*Unknown arguments or flags:.*foo/m)
+      expect(exit_code).to eq(1)
+    end
```
justinstoller: Can we put spaces after these `end`s?

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  describe 'error handling' do
+    let(:response)  { Struct.new(:code, :body) }
+    let(:success)   { response.new('204', nil) }
+    let(:not_found) { response.new('404', 'Not Found') }
+    let(:empty)     { response.new('404', '[]') }
+
+    it 'signs certs with successfull put' do
```
justinstoller: I don't think "signs certs" is what the CLI is doing in response to a successful put. The put signed the certs, the CLI is... logging and exiting with zero, right?

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+    it 'fails when put request errors' do
+      allow(action).to receive(:get_certificate_statuses).and_return(not_found)
+      exit_code = action.run({'all' => true})
+      expect(exit_code).to eq(1)
+    end
+    it 'fails when no pending certs' do
+      allow(action).to receive(:get_all_certs).and_return(empty)
+      exit_code = action.run({'all' => true})
+      expect(exit_code).to eq(1)
```
justinstoller: We should run this by @Magisus when she gets back, but `puppet cert sign` returns `24` when it can't find a/any cert(s) to sign. Should maintain that weirdness or consolidate into just `1`? (We can answer that in a fixup ticket after she gets back)

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+          return check_flag_usage(results, parser)
+        end
+      end
+
+      def parse(args)
+        results = {}
+        unparsed = []
+        invalid = []
+        parser = self.class.parser(results)
+        begin
```
justinstoller: I think you can `Puppetserver::Ca::Utils.parse_with_errors` for this now

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+            @logger.err 'Error:'
+            @logger.err "    #{result.inspect}"
+            exit_code = 1
+          end
+        end
+        return exit_code
+      end
+
+      def store
+        unless @store
```
justinstoller: mmm... We should use `HttpUtilities` here, but it creates a store for each connection. Maybe we should turn `HttpUtilities` into an http client that caches its store? We also don't have to do that in this PR.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+      def sign_requested_certs(certnames)
+        exit_code = 0
+        certnames.each do |certname|
+          result = sign_cert(certname)
+          case result.code
+          when '204'
+            @logger.inform "Signed certificate for #{certname}"
+          when '404'
+            @logger.err 'Error'
+            @logger.err "    Could not find certificate for #{certname}"
```
justinstoller: The error output here varies between in whitespace and colon usage. I think we should circle back around after this and have them all conform to a single format, but for the time being we should make these all the same per action.

I think in this PR we have `Error` with and without a following colon, and then the following line with or without four spaces of indentation.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+        unless result.is_a? Net::HTTPOK
+            @logger.err 'Error:'
+            @logger.err "    #{result.inspect}"
+            return nil
+        end
+        return result
+      end
+
+      def select_pending_certs(get_result)
+        requested_certnames = JSON.parse(get_result.body).select{|e| e["state"] == "requested"}.map{|e| e["name"]}
```
justinstoller: if all we care about is the body, what if we just returned it from `get_all_certs`? That might make testing easier as well.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,213 @@
...
+            @logger.err 'Error'
+            @logger.err "    Could not find certificate for #{certname}"
+            exit_code = 1
+          else
+            @logger.err 'Error:'
+            @logger.err "    #{result.inspect}"
+            exit_code = 1
+          end
+        end
+        return exit_code
```
justinstoller: I ended up doing this in my PR and then changing it. I think that it makes sense for `run` to know about exit codes but for the underlying methods to communicate with success/fail booleans, state symbols, or arrays of them.

https://github.com/puppetlabs/puppetserver-ca-cli/pull/3#discussion_r208422790

lib/puppetserver/ca/cli.rb
```diff
@@ -39,6 +42,7 @@ class Cli
 
       def self.run(cli_args = ARGV, out = STDOUT, err = STDERR)
         logger = Puppetserver::Ca::Logger.new(:info, out, err)
+        #binding.pry
```
justinstoller: whoops

lazartea: I knew there would be at least a few glaringly dumb things

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,175 @@
...
+      BANNER
+      BODY = JSON.dump({desired_state: 'signed'})
+
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--certname x,y,z', Array, 'the name(s) of the cert(s) to be signed') do |cert|
+            parsed['certname'] = cert
+          end
+          opts.on('--config PUPPET.CONF', 'Custom path to Puppet Conf') do |conf|
```
Magisus: I would say "... to Puppet's config file". Puppet Conf is an event in San Francisco :P

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,175 @@
...
+            parsed['config'] = conf
+          end
+          opts.on('--help', 'Display this sign specific help output') do |help|
```
Magisus: "this sign specific" -> "command-specific"

If we ever internationalize this gem (and we probably should), we can use the same string for the `--help` flag of all actions.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,175 @@
...
+
+      def check_flag_usage(results)
+        if results['certname'] && results['all']
+          '--all and --certname cannot be used together'
+        elsif !results['certname'] && !results['all']
+          'No arguments given'
+        elsif results['certname'] && results['certname'].include?('--all')
+          'Cannot use --all with --certname. If you actually have a certificate request ' +
+                          'for a certifcate named --all, you need to use the HTTP API.'
+        end
```
Magisus: I lol'd. Are dashes valid leading characters for a certname?

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end
+
+  describe 'error handling' do
+    let(:response)  { Struct.new(:code, :body) }
+    let(:success)   { response.new('204', nil) }
+    let(:not_found) { response.new('404', 'Not Found') }
+    let(:empty)     { response.new('404', '[]') }
+
+    it 'logs and exits with zero with successfull put' do
```
Magisus: Extra `l` in "successful"

Magisus: Also can we make it `PUT` to make it clearer it's referring to an HTTP action?

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      allow(action).to receive(:sign_cert).and_return(success)
+      exit_code = action.run({'certname' => ['foo']})
+      expect(exit_code).to eq(0)
+      expect(out.string).to include('Signed certificate for foo')
+    end
+
+    it 'fails when put request errors' do
```
Magisus: Same here, can we do `PUT` in all-caps?

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,175 @@
...
+        result = get_all_certs(settings)
+        if result
+          select_pending_certs(result.body)
+        end
+      end
+
+      def sign_requested_certs(certnames,settings)
+        success = true
+        certnames.each do |certname|
+          result = sign_cert(certname,settings)
```
amysahli: So I ran into this as well... Where I was doing a loop of certnames and making the http request w/in that loop.  However, this means that for each cert we open of a new connection and create a new store each time.  (self.with_connection).  Not sure about the efficiency of this.  Thouhts @justinstoller ?  I ended up passing all my certnames together to like ``` sign_cert(certnames, settings)``` and then making just 1 connection for all the requests.

spec/puppetserver/ca/sign_action_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+    end
+
+    it 'does not allow --certname and --all' do
+      exit_code = Puppetserver::Ca::Cli.run(['sign', '--certname', '--all'], out, err)
+      expect(err.string).to match(/Cannot use --all.*HTTP API/)
+      expect(exit_code).to eq(1)
+    end
+
+    it 'works with a single cert' do
+      results, exit_code = action.parse(['--certname', 'foo'])
```
Magisus: Is there a reason we aren't doing the failure cases by calling `parse` directly? I got kinda tripped up figuring out the purpose of the `before` block in this set of tests, and if we did them all but calling parse directly, we wouldn't need it.

lib/puppetserver/ca/sign_action.rb
lazartea: I'm not sure about that, actually. If not, we could lose some of the snark in this error string

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,175 @@
...
+require 'net/https'
+require 'json'
+
+module Puppetserver
+  module Ca
+    class SignAction
+
+      include Puppetserver::Utils
+
+      SUMMARY = "Set up the CA's key, certs, and crls"
```
Magisus: This summary needs to be updated.

spec/puppetserver/ca/sign_action_spec.rb
lazartea: Good point. I went back and fixed all of them and removed the ``before`` block

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,179 @@
...
+          select_pending_certs(result.body)
+        end
+      end
+
+      def sign_requested_certs(certnames,settings)
+        success = true
+        results = sign_cert(certnames, settings)
+        results.each do |certname, result|
+          case result.code
+          when '204'
```
amysahli: are we sure it returns a 204 and not a 200?  I think 204 is standard...but

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,179 @@
...
+        HttpUtilities.with_connection(url, settings) do |connection|
+          connection.get(url)
+        end
+      end
+
+      def sign_cert(certnames,settings)
+        results = { }
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_status')
+        HttpUtilities.with_connection(url, settings) do |connection|
```
lazartea: @amysahli Is this better, do you think? I needed to retain the certnames as well as the results (because we use both for error logging) so I just made it pass back a hash.

lib/puppetserver/ca/sign_action.rb
lazartea: I'm sure. We get 204 back because it's not returning a body, but the request was successful. There are some comments in the PR somewhere (collapsed now because of a lot of squashing) where I talked about this with @Magisus

amysahli: Thanks!  I just noticed that submitting the csr to CA returns a 200 even though it doesnt return content.  So wanted to double check with yours

lib/puppetserver/ca/sign_action.rb
amysahli: yeah, i thinks that looks good

lib/puppetserver/ca/sign_action.rb
```diff
@@ -0,0 +1,179 @@
...
+
+      def get_certificate_statuses(settings)
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_statuses', 'any_key')
+        HttpUtilities.with_connection(url, settings) do |connection|
+          connection.get(url)
+        end
+      end
+
+      def sign_cert(certnames,settings)
```
justinstoller: name should be plural

lazartea:
I think I've addressed everyone's feedback. Anything else or are we ready to merge?

PR #8, opened 2018-08-08
(SERVER-2259) Add clean action

justinstoller:
This provides a clean action. The inputs (and their validation) is the same as the `revoke` action. The actual action is a superset of revocation (send a PUT to revoke and then a DELETE to clean up after ourselves). Consequently a lot of this is directly boilerplate from revoke.

I assume we might want to create a CA object that we can delegate the shared behaviors and validation to? Thought I'd run doing that by folks (in this or another PR).

At a minimum this needs a rebase, but I'ld love to get folks feedback on the approach/see if I'm missing anything.

### DIFF
--- a/lib/puppetserver/ca/clean_action.rb
+++ b/lib/puppetserver/ca/clean_action.rb
@@ -1,0 +1,155 @@
+require 'puppetserver/ca/utils'
+require 'puppetserver/utils/http_utilities'
+require 'puppetserver/utils/file_utilities'
+require 'puppetserver/ca/puppet_config'
+require 'puppetserver/ca/revoke_action'
+
+require 'optparse'
+require 'json'
+
+module Puppetserver
+  module Ca
+    class CleanAction
+
+      include Puppetserver::Utils
+
+      CERTNAME_BLACKLIST = %w{--all --config}
+
+      SUMMARY = 'Clean files from the CA for certificate(s)'
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca clean [--help|--version]
+  puppetserver ca clean [--config] --certname CERTNAME[,ADDLCERTNAME]
+
+Description:
+Given one or more valid certnames, instructs the CA to revoke certificates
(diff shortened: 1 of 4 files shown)


Magisus:
Going to try this out.

justinstoller:
Updated the comment and rebased.

amysahli:
This looks good.  Do we need to do any manual testing?

Magisus:
I'll try this out minus the use of the generate command on the VM I set up earlier.

spec/puppetserver/utils/http_utilities_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+require 'puppetserver/ca/generate_action'
+require 'utils/ssl'
+require 'fileutils'
+
+RSpec.describe Puppetserver::Utils::HttpUtilities do
+  include Utils::SSL
+
+  subject { Puppetserver::Utils::HttpUtilities }
+
+  it 'creates the frickin store correctly' do
```
justinstoller: I should probably change this

justinstoller:
rebased, think this is good to go now!

lib/puppetserver/ca/clean_action.rb
```diff
@@ -0,0 +1,154 @@
...
+      SUMMARY = 'Clean files from the CA for certificate(s)'
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca clean [--help|--version]
+  puppetserver ca clean [--config] --certname CERTNAME[,ADDLCERTNAME]
+
+Description:
+Given one or more valid certnames, instructs the CA to remove files
+pertaining to them (keys, cert, certificate request) over HTTPS using
+the local agent's PKI
```
Magisus: This also revokes the certificate that was issued, I think we should mention that. I found the previous docs we had confusing because they didn't explain how `clean` interacted with `revoke`.

justinstoller: Updated this, let me know if it still needs improvement!

PR #9, opened 2018-08-09
(SERVER-2263) Add create action to generate a new signed cert

amysahli:


lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,253 @@
...
+      # Generate a csr and 2 keys for each certname and return a hash in format
+      # of {certname1 => csr1, certname2 => crs2}
+      def generate_name_csr_hash(certnames, settings, digest)
+        certname_csr_hash = {}
+        host = Puppetserver::Ca::Host.new(digest)
+        certnames.each do |certname|
+          new_private_key = host.create_private_key(settings[:keylength])
+          new_public_key = new_private_key.public_key
+          save(new_private_key, certname, settings[:privatekeydir])
+          save(new_public_key, certname, settings[:publickeydir])
```
amysahli: So, I shouldn't really save the keys until I'm sure we will get certificates made that correspond to those keys.  HOwever, I don't like the idea of holding on to those key strings until all certs have been returned.  Should I?  I guess I could save them in a hash with certname keys and then return all the certnames for the certs generated and then do it in a loop?  Better ideas?

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,253 @@
...
+        url = HttpUtilities.make_ca_url(settings[:ca_server],
+                       settings[:ca_port],
+                       'certificate')
+
+        success = HttpUtilities.with_connection(url, settings) do |connection|
+          certnames.map do |certname|
+            url.resource_name = certname
+            result = connection.get(url)
+            if downloaded = check_download_result(result, certname)
+              save(result.body, certname, settings[:certdir])
```
amysahli: Is it ok to do a save from within a connection?  Could that take too long?  My only other option would be to return a hash of certnames with the cert strings and save them after.

spec/puppetserver/ca/create_action_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      result, code = subject.parse([])
+      expect(code).to eq(1)
+      expect(stderr.string).to include('one certname is required')
+    end
+
+    it 'cannot create certs with the names of flags' do
+      result, code = subject.parse(['--certname', '--config'])
+      expect(code).to eq(1)
+      expect(stderr.string).to include('Cannot manage cert named `--config`')
+      expect(result['certnames']).to eq(['--config'])
```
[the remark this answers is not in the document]
lazartea: @justinstoller  and I both added error logging that handled that. I think he used a blacklist that had all of the commands, and I just specifically disallowed my ``--all`` command. Disallowing certnames that start with ``--`` is a good approach (perhaps the best approach?), and we should probably all come up with a consensus when we refactor.

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,253 @@
...
+            result = connection.put(csr_string, url)
+
+            log_csr_result(result, certname)
+          end
+        end
+      end
+
+      def log_csr_result(result, certname)
+        case result.code
+        when '200', '204'
```
amysahli: Although this actually comes in as 200 with no content, thought we'd accept both in case they change that to a more standard 204 return in the future.

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,253 @@
...
+        end
+      end
+
+      def log_sign_result(result, certname)
+        case result.code
+        when '204'
+          @logger.inform "Signed certificate request for #{certname}"
+        else
+          @logger.err 'Error:'
+          @logger.err "    When download requested for #{certname}:"
```
lazartea: Oh, I like this error message; it's formatted a lot better than mine and gives more valuable context. I might need to steal it.

amysahli: ^ I stole it from Justin :)

amysahli:
Some example output:
```
be ./exe/puppetserver-ca create --certname chum99.net
Error:
    Sorry, looks like a signed certificate already exists for chum99.net
    Can't create a cert for a certname when one already exists.
 ~/Development/puppetserver-ca-cli (SERVER-2263) $🐰  be ./exe/puppetserver-ca create --certname chum100.net
Successfully submitted certificate request for chum100.net
Successfully signed certificate request for chum100.net
Successfully downloaded and saved certificate chum100.net in /Users/amy.sahli/.puppetlabs/etc/puppet/ssl/certs/chum100.net.pem
Successfully saved private key for chum100.net in /Users/amy.sahli/.puppetlabs/etc/puppet/ssl/private_keys
Successfully saved public key for chum100.net in /Users/amy.sahli/.puppetlabs/etc/puppet/ssl/public_keys

```

spec/puppetserver/ca/create_action_spec.rb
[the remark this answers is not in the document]
amysahli: ^^ ok  I added the blacklist option and we can refactor tomorrow

spec/puppetserver/ca/cli_spec.rb
```diff
@@ -363,18 +378,9 @@
...
 
-    it 'does not require the config option' do
-      Dir.mktmpdir do |tmpdir|
-        with_temp_cadir tmpdir do
-          exit_code = Puppetserver::Ca::Cli.run(['generate'], stdout, stderr)
-          expect(exit_code).to eq(0)
-        end
-      end
-    end
-
```
amysahli: I removed this since it was not writing to the tmpdir

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,275 @@
...
+      # Create crls and keys, then submit them to CA, request for the CA to sign
+      # them, download the signed certificates from the CA, and finally save
+      # the signed certs and associated keys.
+      def generate_certs(certnames, settings, digest)
+        certnames.each do |certname|
+          key, csr = generate_key_csr(certname, settings, digest)
+          success1 = submit_certificate_request(certname, csr.to_s, settings)
+          next unless success1
+          success2 = sign_cert(certname, settings)
+          next unless success2
```
amysahli: Is there a nicer way of doing early returns?

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,276 @@
...
+BANNER
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def self.parser(parsed = {})
+        parsed['certnames'] = []
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--certname foo,bar', Array,
```
Magisus: The args should be in ALL CAPS.

lib/puppetserver/ca/host.rb
```diff
@@ -0,0 +1,26 @@
+require 'openssl'
+
+module Puppetserver
+  module Ca
+    class Host
```
Magisus: I guess this is sort of like the client counterpart to the CA object we were discussing. Makes sense.

lib/puppetserver/utils/refactored_client.rb
```diff
@@ -0,0 +1,108 @@
+require 'openssl'
+require 'net/https'
+
+module Puppetserver
+  module Utils
+    # Utilities for doing HTTPS against the CA that wraps Net::HTTP constructs
+    class RefactoredClient
```
Magisus: Are we going to wait to rename this until the refactor ticket?

lib/puppetserver/utils/refactored_client.rb
```diff
@@ -0,0 +1,108 @@
...
+
+          Result.new(result.code, result.body)
+        end
+      end
+
+      # Just provide the bits of Net::HTTPResponse we care about
+      Result = Struct.new(:code, :body)
+
+      # Like URI, but not... maybe of suspicious value
+      URL = Struct.new(:protocol, :host, :port,
```
Magisus: Why did we decide to not just use `URI` here originally?

lib/puppetserver/utils/signing_digest.rb
```diff
@@ -0,0 +1,25 @@
+module Puppetserver
+  module Utils
+    class SigningDigest
+
+      attr_reader :errors, :digest
+
+      def initialize
```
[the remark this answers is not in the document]
Magisus: I think since this is a specific utility object, it sorta makes sense. An alternative might be a module method grouped with other SSL utils (if there are any), just called `digest` perhaps, that returns a (possibly memoized) digest object. Depends on how important it is to us to reuse the object I guess. But I don't mind this.

spec/puppetserver/ca/cli_spec.rb
Magisus: I think we may want to test this at a different level, like via calling the action's `parse` method, if at all.

Magisus:
This needs a rebase.

Magisus:
Tried this out manually on my VM with packages installed, looks good 👍

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,276 @@
...
+  puppetserver ca create [--config] --certname CERTNAME[,ADDLCERTNAME]
+
+Description:
+Creates a new certificate signed by the intermediate CA
+and stores generated keys and certs on disk.
+
+To determine the target location, the default puppet.conf
+is consulted for custom values. If using a custom puppet.conf
+provide it with the --config flag
+Options:
```
Magisus: Can we put a newline above this "Options:"?

spec/puppetserver/ca/cli_spec.rb
amysahli: I'll add that @Magisus

lib/puppetserver/utils/refactored_client.rb
amysahli: @Magisus I'm doing the refactor ticket now.  So either way I'll make sure the work gets done.

lib/puppetserver/utils/refactored_client.rb
amysahli: Not exactly sure.  @justinstoller ?

spec/puppetserver/ca/cli_spec.rb
amysahli: ^ Actually, you moved/added more generate specs in your PR.  So maybe I should add it after you merge yours?

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,288 @@
...
+      # the signed certs and associated keys. Returns true if all certs were
+      # successfully created and saved.
+      def generate_certs(certnames, settings, digest)
+        failed = []
+        certnames.each do |certname|
+          key, csr = generate_key_csr(certname, settings, digest)
+          success1 = submit_certificate_request(certname, csr.to_s, settings)
+          unless success1
+            failed << true
+            return
```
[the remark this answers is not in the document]
amysahli: Nice!  I did that with a map and checked all the values were false.  So looks much better.

lib/puppetserver/utils/refactored_client.rb
justinstoller: I originally just wrapped all of the things that were net/http or it tightly integrated with. I waffled on whether that was excessive after I wrote it, see my comment above this line.  (:

lib/puppetserver/ca/cli.rb
```diff
@@ -21,7 +22,8 @@ class Cli
         'import'   => ImportAction,
         'generate' => GenerateAction,
         'revoke'   => RevokeAction,
-        'sign'     => SignAction
+        'sign'     => SignAction,
+        'create'   => CreateAction,
```
justinstoller: lets put this in alphabetical order

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+        return results, exit_code
+      end
+
+      def run(input)
+        certnames = input['certnames']
+        config_path = input['config']
+
+        # Validate config_path provided
+        if config_path
+          errors = Puppetserver::Utils::FileUtilities.validate_file_paths(config_path)
```
justinstoller: Since you include `Puppetserver::Utils` at the top of this class you should be able to just reference `FileUtilities` here and below.

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+          errors << '    At least one certname is required to create'
+        else
+          results['certnames'].each do |certname|
+            if CERTNAME_BLACKLIST.include?(certname)
+              errors << "    Cannot manage cert named `#{certname}` from " +
+                        "the CLI, if needed use the HTTP API directly"
+            end
+
+            unless certname =~ VALID_CERTNAME
+              errors << '  Certname %{name} must not contain unprintable or non-ASCII characters'
```
justinstoller: Where does `%{name}` come from here?

justinstoller: Is this true?

amysahli: From some old code I forgot to change :(

lib/puppetserver/ca/create_action.rb
amysahli: Thanks!  I'll make sure to change all of them in the file

lib/puppetserver/ca/create_action.rb
amysahli:
```
irb(main):001:0> VALID_CERTNAME = /\A[ -.0-~]+\Z/
=> /\A[ -.0-~]+\Z/
irb(main):002:0> "amy" =~ VALID_CERTNAME
=> 0
```
```
 "¢" =~ VALID_CERTNAME
=> nil
```
So if its valid it returns a 0, which evaluates to 0.  Else, it returns nil

amysahli: What do you mean true?  We currently check this here:
puppet/lib/puppet/ssl/base.rb:31

Magisus: I think that may be leftover from the days when we had worse unicode support. We may be able to allow other characters now. But we'll have to be careful about URL encoding for making requests using the certname. May not be worth updating right now...

justinstoller: Yeah, sorry about that, I didn't mean to be so vauge with the "true" comment. 

As Maggie mentions, I don't think that that regex is actually the limit to valid certnames (according to the spec). Though at one time it may have designated the limit to certnames that Puppet could reliably handle.

Doing a bit more research it looks like any valid unicode can be in a certificate Common Name (this [s.o. answer](https://stackoverflow.com/questions/5136198/what-strings-are-allowed-in-the-common-name-attribute-in-an-x-509-certificate) is the most well written summary I've found), though we often use the CN to match DNS names which it looks like depending on the stack it often gets translated to ASCII ([wiki: internationalized domain names](https://en.wikipedia.org/wiki/Internationalized_domain_name)).

I could have swore we used to have a host `🐎.delivery.puppetlabs.net` but I don't know if our stack was converting it or what it's certname was in Puppet.

Since this is current behavior maybe it isn't worth worrying about now, I was just curious.

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+        end
+      end
+
+      def check_download_result(result, certname)
+        case result.code
+        when '200'
+          return true
+        when '404'
+          @logger.err 'Error:'
+          @logger.err "    Certificate could not be found/created/saved for #{certname}"
```
justinstoller: Maybe something more specific? I'm think something like, "Signed certificate '#{certname}' could not be found on the CA"?

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+        FileUtilities.ensure_dir(puppet.settings[:certdir])
+        FileUtilities.ensure_dir(puppet.settings[:privatekeydir])
+        FileUtilities.ensure_dir(puppet.settings[:publickeydir])
+
+        # Generate and save certs and associated keys
+        all_passed = generate_certs(certnames, puppet.settings, signer.digest)
+        return all_passed ? 0 : 1
+      end
+
+      # Create crls and keys, then submit them to CA, request for the CA to sign
```
justinstoller: Create _certs_ and keys   ... not _crls_

amysahli: thanks.  will fix

lib/puppetserver/ca/create_action.rb
amysahli: ok

lib/puppetserver/ca/create_action.rb
amysahli: ^ Should have been csrs

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+      # @param settings [Hash] a hash of config settings
+      # @return [Boolean] success of all csrs being submitted to CA
+      def submit_certificate_request(certname, csr, settings)
+        client = http_client(settings)
+        url = client.make_ca_url(settings[:ca_server],
+                       settings[:ca_port],
+                       'certificate_request')
+
+        client.with_connection(url) do |connection|
+          url.resource_name = certname
```
justinstoller: We should probably update the calls to `make_ca_url` to include the certname and then when we refactor out the CA update it to cache the connection.

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+          @logger.err 'Error:'
+          @logger.err "    When certificate request submitted for #{certname}:"
+          @logger.err "      code: #{result.code}"
+          @logger.err "      body: #{result.body.to_s}" if result.body
+          return false
+        end
+      end
+
+      # Make an HTTP request to CA to sign the named certificates
+      # @param certname [Array] the name of the certificate to have signed
```
justinstoller: This should be a `String` right?

lib/puppetserver/ca/create_action.rb
```diff
@@ -0,0 +1,268 @@
...
+      end
+
+      def http_client(settings)
+        @client ||= RefactoredClient.new(settings[:localcacert],
+                                         settings[:certificate_revocation],
+                                         settings[:hostcrl])
+      end
+
+      # Make an HTTP request to submit certificate requests to CA
+      # @param certname [Array] the name of the certificate to fetch
```
justinstoller: `String`?

lib/puppetserver/ca/create_action.rb
amysahli: Yes, thank you. Left over from when it was an array

Magisus:
This is ready for a rebase.

lib/puppetserver/utils/http_client.rb
```diff
@@ -45,6 +45,13 @@ def initialize(net_http_connection, url_struct)
           @url = url_struct
         end
 
+        def get(url_overide = nil)
+          url = url_overide || @url
+
+          request = Net::HTTP::Get.new(url.to_uri, HEADERS)
```
amysahli: @Magisus could our auth problem have been that this method lacked "HEADERS" ?

Magisus: I didn't even try this one. I did sign (which is a put) and clean (which is a delete).

amysahli: I don't see how
```
HEADERS = {
        'User-Agent'   => 'PuppetserverCaCli',
        'Content-Type' => 'application/json',
        'Accept'       => 'application/json'
      }
```
would matter but ?

amysahli: Yeah, the get only gets called with the --all flag i guess

PR #10, opened 2018-08-09
(maint) Improve HTTP connections to the CA

justinstoller:


lib/puppetserver/ca/puppet_config.rb
```diff
@@ -113,9 +113,11 @@ def resolve_settings(overrides = {})
         settings[:localcacert] =    overrides.fetch(:localcacert, '$certdir/ca.pem')
         settings[:hostcert] =       overrides.fetch(:hostcert, '$certdir/$certname.pem')
         settings[:hostcrl] =        overrides.fetch(:hostcrl, '$ssldir/crl.pem')
+        settings[:hostcrl] =        overrides.fetch(:hostcrl, '$ssldir/crl.pem')
+        settings[:certificate_revocation] = parse_crl_usage(overrides.fetch(:certificate_revocation, true))
```
justinstoller: might as well be 'true'?

lib/puppetserver/utils/http_utilities.rb
```diff
@@ -23,10 +23,12 @@ def self.make_ca_url(host, port, resource_type = nil, certname = nil)
       # have HTTP verbs defined on it that take a body (and optional
       # overrides). Returns whatever the block given returned.
       def self.with_connection(url, settings, &block)
-        store = make_store(settings[:cacert],
-                           settings[:hostcert],
+        store = make_store(settings[:localcacert],
```
justinstoller: this was wrong before and should definitely be localcacert

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -163,6 +165,17 @@ def explicitly_given_config_file_or_default_config_exists?
       def run(command)
         %x( #{command} )
       end
+
+      def parse_crl_usage(setting)
```
justinstoller: this makes parsing easier on the other end though I don't know how I feel about mixing up the representation of this setting between here and Puppet proper.

lib/puppetserver/utils/http_utilities.rb
```diff
@@ -74,11 +75,24 @@ def to_uri
...
+
+          flags = OpenSSL::X509::V_FLAG_CRL_CHECK
+          if crl_usage == :chain
+            flags |= OpenSSL::X509::V_FLAG_CRL_CHECK_ALL
+          end
+
+          store.flags = flags
+          delimiter = /-----BEGIN X509 CRL-----.*?-----END X509 CRL-----/m
+          File.read(crls).scan(delimiter).each do |crl|
+            store.add_crl(OpenSSL::X509::CRL.new(crl))
```
justinstoller: I'm assuming the PKI for the agent has been set up correctly already, and we've got a lot of logic in the agent for that kind of thing.

I'm assuming we should have some kind of rescue around this, but that we can get away with a "Woah, stuff's broke, go fix your agent's PKI first" kinda thing?

spec/puppetserver/utils/http_utilities_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        ca_name: 'leaf',
+        cadir: cadir,
+        cacert: cacert,
+        cakey: cakey,
+        rootkey: rootkey,
+        cacrl: cacrl
+      }
+
+      digest = generate_action.default_signing_digest
+      generate_action.generate_root_and_intermediate_ca(settings, digest)
```
justinstoller: I think we should probably use this here in testing rather than the ssl utils (I still wasn't able to get it to fail with either of these methods...) But that GenerateAction will eventually become our CA construct?

justinstoller:
This is gross but I wanted to get it out of my head before the end of the day. I'll address my own comments and then rebase this into a single commit and add it to the clean action PR tomorrow morning. @amysahli if you wanted to play around with that RefactorClient tonight I *think* this will address your problems.

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -113,9 +113,11 @@ def resolve_settings(overrides = {})
         settings[:localcacert] =    overrides.fetch(:localcacert, '$certdir/ca.pem')
         settings[:hostcert] =       overrides.fetch(:hostcert, '$certdir/$certname.pem')
         settings[:hostcrl] =        overrides.fetch(:hostcrl, '$ssldir/crl.pem')
+        settings[:hostcrl] =        overrides.fetch(:hostcrl, '$ssldir/crl.pem')
```
justinstoller: dupe

lib/puppetserver/utils/http_utilities.rb
```diff
@@ -23,10 +23,12 @@ def self.make_ca_url(host, port, resource_type = nil, certname = nil)
       # have HTTP verbs defined on it that take a body (and optional
       # overrides). Returns whatever the block given returned.
       def self.with_connection(url, settings, &block)
-        store = make_store(settings[:cacert],
-                           settings[:hostcert],
+        store = make_store(settings[:localcacert],
+                           settings[:certificate_revocation],
                            settings[:hostcrl])
 
+        binding.pry
```
amysahli: mip moop

spec/puppetserver/utils/http_utilities_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+      binding.pry
```
amysahli: mip moop

justinstoller:
@Magisus these look like the correct changes to me, and the test looks right, however "Attempt to write a failing spec" doesn't actually fail, and I'm not seeing why. I'm hoping you might see what I'm missing?

spec/puppetserver/utils/http_utilities_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      hostkey = OpenSSL::PKey::RSA.new(2048)
+      cakey_content = OpenSSL::PKey.read(File.read(settings[:cakey]))
+      cacert_content = OpenSSL::X509::Certificate.new(File.read(settings[:cacert]))
+      hostcert_content = create_cert(hostkey, 'foobar', cakey_content, cacert_content)
+      File.write(hostcert, hostcert_content)
+
+      FileUtils.cp(settings[:cacert], settings[:localcacert])
+      FileUtils.cp(settings[:cacrl], settings[:hostcrl])
+
+      store = subject.make_store(settings[:cacert], :chain, settings[:hostcrl])
```
amysahli: i see above that this is the same as settings[:localcacert], but would be easier to follow if we supply  settings[:localcacert] here instead of settings[:cacert]

justinstoller: yeah, good catch

lib/puppetserver/utils/http_utilities.rb
Magisus: Yeah I think so.

justinstoller: I added something like this when debugging the issue on a VM:
```rb
        begin
          store = make_store(settings[:localcacert],
                             settings[:certificate_revocation],
                             settings[:hostcrl])
          request = ->(conn) { block.call(Connection.new(conn, url)) }
          result = Net::HTTP.start(url.host, url.port,
                          use_ssl: true, cert_store: store,
                          &request)
          return result
        rescue => e
          STDOUT.puts 'Unable to complete request!'
          STDOUT.puts 'Attempting connection to URL:'
          STDOUT.puts pp(url)
          STDOUT.puts 'With SSL Store:'
          STDOUT.puts pp(store)
          STDOUT.puts 'Encountered exception:'
          STDOUT.puts pp(e)
          raise e
        end
```

And I realized that we should pass in a logger here if we're going to do this correctly. What do you think about doing that when we turn it into a class/client?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -62,6 +63,18 @@ def load
...
+        domain = Facter.value(:domain)
+        if domain and domain != ''
+          fqdn = [hostname, domain].join('.')
+        else
+          fqdn = hostname
+        end
+        fqdn.chomp('.')
+      end
+
+
```
amysahli: extra space here :)

spec/puppetserver/utils/http_utilities_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      cadir = tmpdir
+      cacert = File.join(tmpdir, 'ca_crt.pem')
+      cakey = File.join(tmpdir, 'ca_key.pem')
+      rootkey = File.join(tmpdir, 'root_key.pem')
+      cacrl = File.join(tmpdir, 'ca_crl.pem')
+      localcacert = File.join(tmpdir, 'localcacert.pem')
+      hostcrl = File.join(tmpdir, 'hostcrl.pem')
+      hostcert = File.join(tmpdir, 'hostcert.pem')
+
+
```
amysahli: extra space here :)

amysahli:
Great job @Magisus and @justinstoller

lib/puppetserver/utils/http_utilities.rb
```diff
@@ -35,7 +35,6 @@ def self.with_connection(url, settings, &block)
       end
```
amysahli: extra space?

lib/puppetserver/ca/puppet_config.rb
Magisus: I'm fine doing this here. In the PR I put up, I'm renaming at least one setting to make it more accurate. I think we should do the right thing here internally as much as possible.

PR #11, opened 2018-08-10
(SERVER-2269) Allow specifying subject alt names when generating the CA

Magisus:
This commit adds the ability to specify subject alternative names for
the intermediate signing cert when using the `generate` action. These
can be configured via a flag, or read from the puppet.conf file. If none
are provided, Puppet's default set of alt names will be added. This
supports both DNS and IP alt names. If type is not specified, DNS names
are assumed.

### DIFF
--- a/lib/puppetserver/ca/generate_action.rb
+++ b/lib/puppetserver/ca/generate_action.rb
@@ -3,6 +3,8 @@ require 'openssl'
 require 'puppetserver/utils/file_utilities'
 require 'puppetserver/ca/host'
 require 'puppetserver/utils/signing_digest'
+require 'puppetserver/settings/ttl_setting'
+require 'facter'
 
 module Puppetserver
   module Ca
@@ -28,10 +30,17 @@ module Puppetserver
 Usage:
   puppetserver ca generate [--help]
   puppetserver ca generate [--config PATH]
+  puppetserver ca generate [--subject-alt-names ALTNAME[,ADDLALTNAME]]
 
 Description:
 Generate a root and intermediate signing CA for Puppet Server
 and store generated CA keys, certs, and crls on disk.
+
+The `--subject-alt-names` flag can be used to add SANs to the CA
+signing cert. Multiple names can be listed as a comma separated
+string. These can be either DNS names or IP addresses, differentiated
+by prefixes: `DNS:foo.bar.com,IP:123.456.789`. Names with no prefix
+will be treated as DNS names.
 
 To determine the target location, the default puppet.conf
 is consulted for custom values. If using a custom puppet.conf
@@ -60,9 +69,11 @@ BANNER
         signer = SigningDigest.new
         return 1 if Utils.handle_errors(@logger, signer.errors)
 
+        subject_alt_names = choose_alt_names(input['subject_alt_names'], puppet.settings[:subject_alt_names])
+
         # Generate root and intermediate ca and put all the certificates, crls,
         # and keys where they should go.
-        generate_root_and_intermediate_ca(puppet.settings, signer.digest)
+        generate_root_and_intermediate_ca(puppet.settings, signer.digest, subject_alt_names)
 
         # Puppet's internal CA expects these file to exist.
         FileUtilities.ensure_file(puppet.settings[:serial], "001", 0640)
@@ -72,7 +83,7 @@ BANNER
         return 0
       end
 
-      def generate_root_and_intermediate_ca(settings, signing_digest)
+      def generate_root_and_intermediate_ca(settings, signing_digest, subject_alt_names = '')
         valid_until = Time.now + settings[:ca_ttl]
         host = Puppetserver::Ca::Host.new(signing_digest)
 
@@ -82,7 +93,7 @@ BANNER
 
         int_key = host.create_private_key(settings[:keylength])
         int_csr = host.create_csr(settings[:ca_name], int_key)
-        int_cert = sign_intermediate(root_key, root_cert, int_csr, valid_until, signing_digest)
+        int_cert = sign_intermediate(root_key, root_cert, int_csr, valid_until, signing_digest, subject_alt_names)
         int_crl = create_crl_for(int_cert, int_key, valid_until, signing_digest)
 
         FileUtilities.ensure_dir(settings[:cadir])
@@ -149,7 +160,7 @@ BANNER
         crl
       end
 
-      def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest)
+      def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest, subject_alt_names)
         cert = OpenSSL::X509::Certificate.new
 
         cert.public_key = csr.public_key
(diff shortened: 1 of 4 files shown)


lib/puppetserver/ca/generate_action.rb
```diff
@@ -206,11 +217,38 @@ def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest)
...
+        alt_names_ext = ef.create_extension("subjectAltName", subject_alt_names, false)
+        cert.add_extension(alt_names_ext)
+
         cert.sign(ca_key, signing_digest)
 
         cert
       end
 
+      def choose_alt_names(cli_alt_names, settings_alt_names)
+        if !cli_alt_names.empty?
```
amysahli: maybe ```.any``` would be easier to read?

Magisus: It's a string, not an array, does that work?

amysahli: :(

Magisus: Nothing else looked better.

amysahli:
This looks good!  Just need to update specs (like in the refactor we talked about during refactor meeting)

amysahli:
Needs to be rebased :(  @justinstoller You wanna take a look at this so we can merge?

lib/puppetserver/ca/generate_action.rb
```diff
@@ -206,11 +217,40 @@ def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest)
...
+          alt_names_ext = ef.create_extension("subjectAltName", subject_alt_names, false)
+          cert.add_extension(alt_names_ext)
+        end
+
         cert.sign(ca_key, signing_digest)
 
         cert
       end
 
+      def choose_alt_names(cli_alt_names, settings_alt_names)
```
justinstoller: This would be a good thing to have docs on, I was unsure of the inputs and outputs on the first pass through.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -260,6 +301,10 @@ def self.parser(parsed = {})
           opts.on('--config CONF', 'Path to puppet.conf') do |conf|
             parsed['config'] = conf
           end
+          opts.on('--subject-alt-names NAMES',
+                  'Subject alternative names for the CA signing cert') do |sans|
+            parsed['subject_alt_names'] = sans || ''
```
justinstoller: I don't think this block will get called if `sans` would be falsy?

spec/puppetserver/ca/generate_action_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    it 'prepends "DNS" to unprefixed alt names' do
+      expect(subject.munge_alt_names('foo.com,IP:123.456.789')).to eq('DNS:foo.com, IP:123.456.789')
+    end
+
+    it 'adds subject alt names to the intermediate CA cert' do
+      digest = subject.default_signing_digest
+      valid_until = Time.now + 1000
+
+      root_key = subject.create_private_key(4096)
```
justinstoller: Test suite time-wise, it might be nice to have a smaller keylength later on.

spec/puppetserver/ca/generate_action_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'spec_helper'
+require 'utils/ssl'
+require 'shared_examples/cli_parsing'
+
+require 'puppetserver/ca/cli'
+require 'puppetserver/ca/generate_action'
+require 'puppetserver/ca/logger'
+require 'puppetserver/utils/signing_digest'
+require 'puppetserver/ca/host'
```
Magisus: I really don't like pulling in this many things here, but I think it will get significantly cleaner once we have a real CA object. I'm thinking we should either rename `Host` or fold it in elsewhere when we do that refactor; the fact that you can create multiple keys and CSRs with the same `Host` object seems misleading.

lib/puppetserver/ca/generate_action.rb
```diff
@@ -220,6 +267,10 @@ def self.parser(parsed = {})
           opts.on('--config CONF', 'Path to puppet.conf') do |conf|
             parsed['config'] = conf
           end
+          opts.on('--subject-alt-names NAMES',
```
amysahli: Is there a benefit to processing this as a string instead of Array like we do for certnames?

lib/puppetserver/ca/generate_action.rb
```diff
@@ -166,11 +177,46 @@ def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest)
...
+        elsif !settings_alt_names.empty?
+          sans = settings_alt_names
+        else
+          sans = "puppet, #{Facter.value(:fqdn)}, puppet.#{Facter.value(:domain)}"
+        end
+        munge_alt_names(sans)
+      end
+
+      def munge_alt_names(names)
+        raw_names = names.split(/\s*,\s*/).map(&:strip)
```
amysahli: Do we have any problems with altnames having spaces?
ex:
```
"hi, yo good"
=> "hi, yo good"
irb(main):007:0> names.split(/\s*,\s*/).map(&:strip)
=> ["hi", "yo good"]
irb(main):008:0> raw_names = ["hi", "yo good"]
=> ["hi", "yo good"]
irb(main):009:0> munged_names = raw_names.map do |name|
irb(main):010:1*           # Prepend the DNS tag if no tag was specified
irb(main):011:1*           if !name.start_with?("IP:") && !name.start_with?("DNS:")
irb(main):012:2>             "DNS:#{name}"
irb(main):013:2>           else
irb(main):014:2*             name
irb(main):015:2>           end
irb(main):016:1>         end.sort.uniq.join(", ")
=> "DNS:hi, DNS:yo good"
```

amysahli: Also, if this were processed as an array (like certnames) we wouldn't have to do a split and strip, or?

lib/puppetserver/ca/generate_action.rb
amysahli: also, i think we need to change the error handling to catch when we don't provide any names to the flag:
```
be ./exe/puppetserver-ca generate --subject-alt-names
Error:
Unknown arguments or flags:
    --subject-alt-names
```

amysahli: which is my fault:
```
be ./exe/puppetserver-ca generate --config
Error:
Unknown arguments or flags:
    --config
    
```

amysahli: Should be easily fixable with new parsing we've been using (replacing parse and parse_inputs) with:
```
def parse(args)
        results = {}
        parser = self.class.parser(results)

        errors = Utils.parse_with_errors(parser, args)

        if check_flag_usage(results)
          errors << check_flag_usage(results)
        end

        errors_were_handled = Utils.handle_errors(@logger, errors, parser.help)

        exit_code = errors_were_handled ? 1 : nil

        return results, exit_code
      end
```

amysahli: Oh, should "NAMES" here be  comma separated like "FOO,BAR" ?

Magisus: As far as benefit of processing as a string, I only did it that way because the value coming out of Puppet's config is a string. So alternatively I could call `split` just on that one before passing it to munge, but it kind of seemed six-a-one-half-a-dozen-the-other to me. And since they ultimately need to be a single string when the get attached to the cert, I guess I was thinking that the fact they become an array in the middle (for inserting prefixes) is mostly an implementation detail.

PR #12, opened 2018-08-10
(SERVER-2260) Add list option

lazartea:


lib/puppetserver/ca/cli.rb
```diff
@@ -4,6 +4,8 @@
 require 'puppetserver/ca/import_action'
 require 'puppetserver/ca/generate_action'
 require 'puppetserver/ca/revoke_action'
+require 'puppetserver/ca/list_action'
+require 'puppetserver/ca/utils'
```
amysahli: Do we need this?

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
+require 'puppetserver/ca/utils'
+require 'puppetserver/utils/http_utilities'
+require 'puppetserver/utils/file_utilities'
+require 'puppetserver/ca/puppet_config'
+require 'optparse'
+require 'openssl'
+require 'net/https'
```
amysahli: i dont think you need ```openssl``` or ```net/https```

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
...
+require 'json'
+
+module Puppetserver
+  module Ca
+    class ListAction
+
+      include Puppetserver::Utils
+
+      SUMMARY = 'List all certificates'
```
amysahli: maybe 'List all certificate requests' ?

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
...
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--config PUPPET.CONF', 'Custom path to Puppet\'s config file') do |conf|
+            parsed['config'] = conf
+          end
+          opts.on('--help', 'Display this command specific help output') do |help|
+            parsed['help'] = true
+          end
+          opts.on('--version', 'Output the version') do |v|
```
amysahli: I don't think we need version anymore

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
...
+          @logger.inform "Requested Certificates:"
+          print_output(requested)
+        end
+      end
+
+      def print_output(certs)
+        padded = 0
+        certs.each do |cert|
+          cert_size = cert["name"].size
+          padded =  cert_size if cert_size > padded
```
amysahli: small extra space here

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
...
+        return 0
+      end
+
+      def human_readable(requested, signed = nil)
+        if signed && signed.size != 0
+          @logger.inform "Signed Certificates:"
+          print_output(signed)
+        end
+
+        if requested.size != 0
```
amysahli: I think you can do ```requested.any?```

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
...
+      def separate_certs(all_certs)
+        requested = all_certs.find_all { |v| v["state"] == "requested" }
+        signed    = all_certs.find_all { |v| v["state"] == "signed" }
+        return requested, signed
+      end
+
+      def get_all_certs(settings)
+        result = get_certificate_statuses(settings)
+
+        unless result.is_a? Net::HTTPOK
```
amysahli: So if there is an error, we will want to return a 1 from the run method.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,139 @@
...
+
+        if config
+          errors = FileUtilities.validate_file_paths(config)
+          return 1 if Utils.handle_errors(@logger, errors)
+        end
+
+        puppet = PuppetConfig.parse(config)
+        return 1 if Utils.handle_errors(@logger, puppet.errors)
+
+        all_certs = get_all_certs(puppet.settings)
```
amysahli: so maybe:
```return 1 if all_certs.nil? ```

File: lib/puppetserver/ca/list_action.rb [hunk 9]
```diff
@@ -0,0 +1,139 @@
...
+        requested, signed = separate_certs(all_certs)
+        if input['all']
+          human_readable(requested, signed)
+        else
+          human_readable(requested)
+        end
```
lazartea: would ``input['all'] ? human_readable(requested, signed) : human_readable(requested)`` be an improvement on this if/else block or should I leave it as is? Not sure if consolidating this would make it confusing.

lib/puppetserver/ca/list_action.rb
amysahli: Also, maybe we want to specifically catch the NO CONTENT.  So that its not really an error, there just aren't any certificates.  I'm assuming no certificates returns 204?  or?

lib/puppetserver/ca/list_action.rb
amysahli: Yeah, we could add an else statement here just to inform the user that no errors happened, but no certificates were returned.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      SUMMARY = 'List all certificate requests'
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca list [--help|--version]
+  puppetserver ca list [--config]
+  puppetserver ca list [--all]
+
+Description:
+List outstanding certificates. If --all is specified, signed certificates will be listed as well.
```
Magisus: "...outstanding certificate requests."

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,136 @@
...
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--config PUPPET.CONF', 'Custom path to Puppet\'s config file') do |conf|
+            parsed['config'] = conf
+          end
+          opts.on('--help', 'Display this command specific help output') do |help|
+            parsed['help'] = true
+          end
+          opts.on('--all', 'List signed certificates') do |a|
```
Magisus: Is this also going to list revoked certs like the old command?

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,136 @@
...
+                                        'certificate_statuses', 'any_key')
+        HttpUtilities.with_connection(url, settings) do |connection|
+          connection.get(url)
+        end
+      end
+
+      def separate_certs(all_certs)
+        requested = all_certs.find_all { |v| v["state"] == "requested" }
+        signed    = all_certs.find_all { |v| v["state"] == "signed" }
+        return requested, signed
```
Magisus: I think we also want to be finding and listing revoked certs.

spec/puppetserver/ca/list_action_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'puppetserver/ca/list_action'
+require 'puppetserver/ca/cli'
+require 'puppetserver/ca/logger'
+
+RSpec.describe 'Puppetserver::Ca::ListAction' do
+  let(:err)    { StringIO.new }
+  let(:out)    { StringIO.new }
+  let(:logger) { Puppetserver::Ca::Logger.new(:info, out, err) }
+  let(:action) { Puppetserver::Ca::ListAction.new(logger) }
+  let(:result) {[{"name"=>"foo", "state"=>"signed", "dns_alt_names"=>["DNS:foo", "DNS:bar"],
```
Magisus: Oh gross, even the API calls them `dns_alt_names`?

lib/puppetserver/ca/list_action.rb
lazartea: Yep, I just forgot about that. Should I change this to list all certificates, then? And then add an output line ``Revoked Certificates:`` ?

spec/puppetserver/ca/list_action_spec.rb
lazartea: yes :( but we list them as ``alt names`` when we print them out

lib/puppetserver/ca/list_action.rb
justinstoller: I think they're excellent

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,136 @@
...
+      BODY = JSON.dump({desired_state: 'signed'})
+
+      def initialize(logger)
+        @logger = logger
+      end
+
+      def self.parser(parsed = {})
+        OptionParser.new do |opts|
+          opts.banner = BANNER
+          opts.on('--config PUPPET.CONF', 'Custom path to Puppet\'s config file') do |conf|
```
justinstoller: I think we decided this should just be CONF rather than PUPPET.CONF

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,136 @@
...
+        input['all'] ? human_readable(requested, signed) : human_readable(requested)
+
+        return 0
+      end
+
+      def human_readable(requested, signed = nil)
+        if signed && signed.any?
+          @logger.inform "Signed Certificates:"
+          print_output(signed)
+        elsif !requested.any?
```
justinstoller: I think you want `empty?` here though `none?` is the inverse of `any?`.

justinstoller: I think this condition should be part of the following conditional

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,136 @@
...
+        return 1 if all_certs.nil?
+
+        requested, signed = separate_certs(all_certs)
+        input['all'] ? human_readable(requested, signed) : human_readable(requested)
+
+        return 0
+      end
+
+      def human_readable(requested, signed = nil)
+        if signed && signed.any?
```
justinstoller: In `puppet cert` we list pending requests, then signed certs, then revoked certs. I think that's a good order and in that case we'd want to put this signed block after the requested block.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,145 @@
...
+      def get_certificate_statuses(settings)
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_statuses', 'any_key')
+        HttpUtilities.with_connection(url, settings) do |connection|
+          connection.get(url)
+        end
+      end
+
+      def separate_certs(all_certs)
+        requested = all_certs.find_all { |v| v["state"] == "requested" }
```
justinstoller: I think there's also `Enumerable#sort_by` that will do this in one pass.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,145 @@
...
+        signed    = all_certs.find_all { |v| v["state"] == "signed"    }
+        revoked   = all_certs.find_all { |v| v["state"] == "revoked"   }
+        return requested, signed, revoked
+      end
+
+      def get_all_certs(settings)
+        result = get_certificate_statuses(settings)
+
+        unless result.is_a? Net::HTTPOK
```
justinstoller: Can we update `HttpUtilities::Connection#get` to return an `HttpUtilities::Result` and then check that the code is '200' rather than the class is `Net::HTTPOK`?

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,145 @@
...
+          print_output(signed)
+        end
+
+        unless revoked.empty?
+          @logger.inform "Revoked Certificates:"
+          print_output(revoked)
+        end
+      end
+
+      def print_output(certs)
```
justinstoller: super_vague_method_name   :P

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,145 @@
...
+        padded = 0
+        certs.each do |cert|
+          cert_size = cert["name"].size
+          padded = cert_size if cert_size > padded
+        end
+
+        certs.each do |cert|
+          @logger.inform "    #{cert["name"]}".ljust(padded + 6) + " (SHA256) #{cert["fingerprints"]["SHA256"]}" +
+                             (cert["dns_alt_names"].empty? ? "" : "\talt names: #{cert["dns_alt_names"]}")
```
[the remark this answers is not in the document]
Magisus: If it's not hard to do we could just add it here.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,145 @@
...
+        all_certs = get_all_certs(puppet.settings)
+        return 1 if all_certs.nil?
+
+        requested, signed, revoked = separate_certs(all_certs)
+        input['all'] ? human_readable(requested, signed, revoked) : human_readable(requested)
+
+        return 0
+      end
+
+      def human_readable(requested, signed = [], revoked = [])
```
justinstoller: I also don't know about this method name  (;

what's the distinction between this and the method it calls?

lib/puppetserver/ca/list_action.rb
lazartea: Yep, and then I'll modify the sign action to support that.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,145 @@
...
+
+      def print_output(certs)
+        padded = 0
+        certs.each do |cert|
+          cert_size = cert["name"].size
+          padded = cert_size if cert_size > padded
+        end
+
+        certs.each do |cert|
+          @logger.inform "    #{cert["name"]}".ljust(padded + 6) + " (SHA256) #{cert["fingerprints"]["SHA256"]}" +
```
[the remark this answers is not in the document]
lazartea: The + 6 was to account for the tab at the beginning of the line, but there's probably a better way to do it.

lib/puppetserver/ca/list_action.rb
amysahli: I'm fine with either

lib/puppetserver/ca/cli.rb
```diff
@@ -21,7 +22,8 @@ class Cli
         'import'   => ImportAction,
         'generate' => GenerateAction,
         'revoke'   => RevokeAction,
-        'sign'     => SignAction
+        'sign'     => SignAction,
+        'list'     => ListAction
```
justinstoller: Can you put this one in alaphabetical order?

lib/puppetserver/ca/list_action.rb
lazartea: I agree that they're both awful names... Ideas? The reason that I separated them was because I didn't want to repeat code. I think ``human_readable`` is determining which headers to print, and ``print_output`` is actually doing the logging.

lib/puppetserver/ca/list_action.rb
amysahli: What do you mean one pass?

lazartea: so if I do ``something = all_certs.sort_by {|v| v["state"]}`` I know that it will sort ``all_certs``  so that each entry is grouped by ``state``, but how do I then split ``something`` into three separate vars? I'm sure there's a way to do this all in one line but I'm not fluent enough in Ruby to figure it out

amysahli: Yeah, i remember doing something similar to that in clojure, but not ruby

justinstoller: d'oh. I meant `group_by`, thanks for catching that @amysahli !

lib/puppetserver/ca/list_action.rb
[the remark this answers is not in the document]
justinstoller: There's four spaces of white space before the certname though, right? I'm missing where the other two come from.

lib/puppetserver/ca/list_action.rb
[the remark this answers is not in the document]
justinstoller: FWIW, it'd require updating the HTTP API, https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1210-L1226

lib/puppetserver/ca/list_action.rb
justinstoller: yeah, I think those are good descriptions of the methods and we should give them names similar to what you described, like `output_certs_by_state` and `output_certs` or something like that.

lib/puppetserver/ca/list_action.rb
lazartea: So when I do ``requested = all_certs.find_all{|v| v["state"] == "requested"}`` and there aren't any, I get ``requested = []``. But if I do ``certs = all_certs.group_by{|v| v["state"]}`` and then ``requested = certs["requested"]`` if there aren't any, I get ``requested = nil``, which breaks everything. Am I doing something wrong?

lib/puppetserver/ca/list_action.rb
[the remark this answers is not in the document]
justinstoller: It looks like we hardcode 256 as a the default in [puppetserver](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L1210-L1226). Though if we move forward with FIPS enablement we might change that? I suppose the finding the hash name is YAGNI and we should wait until we allow specifying a parameter to this action before making this any more complicated.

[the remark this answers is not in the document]
lazartea: I'm not really sure why we need the two extra spaces, but without them we get:

screen shot 2018-08-13 at 11 23 37 am

which I didn't think looked as nice as 
screen shot 2018-08-13 at 11 23 59 am

I already removed hardcoding the 256 and matching the default, but I can pull that back if you'd rather.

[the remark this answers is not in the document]
lazartea: Also ignore the ``(default)`` that was me playing around with the SHAs.

lib/puppetserver/ca/list_action.rb
justinstoller: No, that's `group_by`'s API, but we could either change the method signature or  check for nil and return an Array in that case.

So we could create a Hash that returns Arrays for unknown keys rather than nil with this:
```rb
certs = Hash.new([]).merge(all_certs.group_by {|c| c['state'] })
```
I don't love that, it might be a little too clever
Implicit or explicit nil checking might be better:
```rb
certs = all_certs.group_by {|c| c['state'] }
requested = certs.fetch('requested', [])
# or
request = cert['requested'] || []
```

One of those options is probably better than changing the method signature. If we did that we'd then probably have to go back to nil checking in the output method, eg
```rb
def print_output(maybe_requested = nil, maybe_signed = nil, maybe_revoked = nil)
  requested = Array(maybe_requested)
  signed = Array(maybe_signed)
  revoked = Array(maybe_revoked)

  ## ... rest of current method
end
```

lazartea: Okay, I'll start playing around with those and see which works best. Thank you!

lib/puppetserver/ca/list_action.rb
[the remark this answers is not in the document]
justinstoller: Okay, looks good, I just wasn't grokking `ljust` correctly.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,144 @@
...
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_statuses', 'any_key')
+        HttpUtilities.with_connection(url, settings) do |connection|
+          connection.get(url)
+        end
+      end
+
+      def separate_certs(all_certs)
+        certs = all_certs.group_by { |v| v["state"]}
+        requested, signed, revoked = certs.fetch("requested", []), certs.fetch("signed", []), certs.fetch("revoked", [])
```
lazartea: @justinstoller does putting this all on one line confuse things, or make it more readable? I couldn't decide.

lib/puppetserver/ca/sign_action.rb
```diff
@@ -100,7 +100,7 @@ def sign_certs(certnames,settings)
       def get_all_certs(settings)
         result = get_certificate_statuses(settings)
 
-        unless result.is_a? Net::HTTPOK
+        unless result.code == 200
```
lazartea: should this be in a separate commit?

Magisus: Seems fine to me as is. It's a minor refactor.

lib/puppetserver/ca/list_action.rb
justinstoller: I think it's understandable, but we usually try to keep lines of code less than 80 characters. I'd probably break it up for that reason.

lazartea: will do

Magisus:
This is ready for a rebase and update to use the new HttpClient.

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,146 @@
+require 'puppetserver/ca/utils'
+require 'puppetserver/utils/http_utilities'
```
amysahli: You'll want to change this to ```puppetserver/utils/http_client```

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,146 @@
...
+        certs.each do |cert|
+          @logger.inform "    #{cert["name"]}".ljust(padded + 6) + " (SHA256) " + " #{cert["fingerprints"]["SHA256"]}" +
+                             (cert["dns_alt_names"].empty? ? "" : "\talt names: #{cert["dns_alt_names"]}")
+          end
+      end
+
+      def get_certificate_statuses(settings)
+        url = HttpUtilities.make_ca_url(settings[:ca_server], settings[:ca_port],
+                                        'certificate_statuses', 'any_key')
+        HttpUtilities.with_connection(url, settings) do |connection|
```
amysahli: You'll want to do something like:
```
client = HttpClient.new(settings[:localcacert],
                                         settings[:certificate_revocation],
                                         settings[:hostcrl])

        url = client.make_ca_url(settings[:ca_server], settings[:ca_port],
                                        'certificate_statuses', 'any_key')

    client.with_connection(url) do ...
```
^ Note that with_connection no longer takes ```settings``` as an argument

lazartea:
@amysahli @Magisus just rebased and added Amy's utils into my code following how she refactored the sign command. Thoughts?

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,155 @@
...
+        return 1 if all_certs.nil?
+
+        requested, signed, revoked = separate_certs(all_certs)
+        input['all'] ? output_certs_by_state(requested, signed, revoked) : output_certs_by_state(requested)
+
+        return 0
+      end
+
+      def output_certs_by_state(requested, signed = [], revoked = [])
+
```
amysahli: maybe no space?

amysahli: but i can't format :(

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,155 @@
...
+        if revoked.empty? && signed.empty? && requested.empty?
+          @logger.inform "No certificates to list"
+          return
+        end
+
+        unless requested.empty?
```
amysahli: i think 
```
if requested.any?
```
is easier to read

lib/puppetserver/utils/http_client.rb
```diff
@@ -60,6 +60,8 @@ def get(url_overide = nil)
 
           request = Net::HTTP::Get.new(url.to_uri)
           result = @conn.request(request)
+
+          Result.new(result.code, result.body)
```
amysahli: I have this in my PR too :) Whoever is second can refactor

lib/puppetserver/ca/list_action.rb
lazartea: so @justinstoller had thoughts about why this should be ``empty`` rather than ``any``, but I can't remember exactly what they were... Willing to do either!

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,155 @@
...
+module Puppetserver
+  module Ca
+    class ListAction
+
+      include Puppetserver::Utils
+
+      SUMMARY = 'List all certificate requests'
+      BANNER = <<-BANNER
+Usage:
+  puppetserver ca list [--help|--version]
```
amysahli: i think we want to get rid of --version here

lib/puppetserver/ca/list_action.rb
```diff
@@ -0,0 +1,155 @@
...
+        revoked = certs.fetch("revoked", [])
+        return requested, signed, revoked
+      end
+
+      def get_all_certs(settings)
+        result = get_certificate_statuses(settings)
+
+        unless result.code == '200'
+          @logger.err 'Error:'
+          @logger.err "    #{result.inspect}"
```
amysahli: maybe like we've done elsewhere?  I'll leave it up to you though
```
      @logger.err "      code: #{result.code}"
      @logger.err "      body: #{result.body.to_s}" if result.body
```

lazartea: yep, will do

lib/puppetserver/ca/list_action.rb
amysahli: dunno?  I think it should be fine:
```
[].any?
=> false
irb(main):002:0> [nil].any?
=> false
irb(main):003:0> ["hey"].any?
=> true
```

Magisus: I could go either way. I know some people really don't like `unless`.

lib/puppetserver/utils/http_client.rb
```diff
@@ -62,6 +62,15 @@ def put(body, url_override = nil)
           Result.new(result.code, result.body)
         end
 
+        def get(url_overide = nil)
```
amysahli: I think we have 2 "gets" now.  (See line 48 above...sorry i did that)  So you should be able to add the Result.new to that.

lib/puppetserver/ca/list_action.rb
lazartea: I think his argument was something about ``if requested.any?`` (with the inverse being  ``unless requested.none?`` and not ``unless requested.empty?``) being syntactically/conceptually different from ``unless empty?`` Is the consensus is that ``if requested.any?`` makes more sense?

lib/puppetserver/utils/http_client.rb
lazartea: Yep, I missed that.

lib/puppetserver/ca/cli.rb
```diff
@@ -24,8 +25,9 @@ class Cli
         'create'   => CreateAction,
         'generate' => GenerateAction,
         'import'   => ImportAction,
+        'list'     => ListAction,
         'revoke'   => RevokeAction,
-        'sign'     => SignAction,
+        'sign'     => SignAction
```
Magisus: I'm a fan of keeping the trailing comma, actually. Makes future diffs cleaner, if you have to add something to the end. The only line changed is the one you added.

PR #13, opened 2018-08-10
(SERVER-2275) Refactor code to make an HTTP client

amysahli:
Refactor to create an actual HTTP client object that
initializes SSL state once and stores it, allowing
multiple connections to be made from it.

### DIFF
--- a/spec/puppetserver/utils/http_client_spec.rb
+++ b/spec/puppetserver/utils/http_client_spec.rb
@@ -1,14 +1,12 @@
 require 'spec_helper'
-require 'puppetserver/utils/http_utilities'
+require 'puppetserver/utils/http_client'
 require 'puppetserver/ca/logger'
 require 'puppetserver/ca/generate_action'
 require 'utils/ssl'
 require 'fileutils'
 
-RSpec.describe Puppetserver::Utils::HttpUtilities do
+RSpec.describe Puppetserver::Utils::HttpClient do
   include Utils::SSL
-
-  subject { Puppetserver::Utils::HttpUtilities }
 
   it 'creates a store that can validate connections to CA' do
     stdout = StringIO.new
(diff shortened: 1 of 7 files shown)


amysahli:
^^ That's weird.  My specs pass locally.
I won't merge this until the clean action gets merged

lib/puppetserver/utils/http_utilities.rb
```diff
@@ -4,36 +4,39 @@
 module Puppetserver
   module Utils
     # Utilities for doing HTTPS against the CA that wraps Net::HTTP constructs
-    module HttpUtilities
+    class HttpUtilities
```
Magisus: I think we should give this a different name if it's going to be an actual class that gets instantiated. Something like `HttpClient`?

spec/puppetserver/utils/http_client_spec.rb
```diff
@@ -53,7 +51,10 @@
       FileUtils.cp(settings[:cacert], settings[:localcacert])
       FileUtils.cp(settings[:cacrl], settings[:hostcrl])
 
-      store = subject.make_store(settings[:localcacert], :chain, settings[:hostcrl])
+      client = Puppetserver::Utils::HttpClient.new(settings[:localcacert],
+                                                      :chain,
```
justinstoller: indentation is off here, should be in vertical line with the start of `settings` above.

Magisus:
One bug here: https://github.com/puppetlabs/puppetserver-ca-cli/blob/1008a86f769d2eda1ee9a042b0e6efa3a57d4878/lib/puppetserver/ca/sign_action.rb#L156 `body_to_s` is apparently not a valid method on result. Changing it to `body.to_s` fixes it.

Magisus:
One bug here: https://github.com/puppetlabs/puppetserver-ca-cli/blob/1008a86f769d2eda1ee9a042b0e6efa3a57d4878/lib/puppetserver/ca/sign_action.rb#L156 `body_to_s` is apparently not a valid method on result. Changing it to `body.to_s` fixes it.

Magisus:
Manually tested this with packages, and it seems to work well. I did discover that we probably need to update the auth rules around the `certificate_status` endpoint, it's currently set to deny all out of the box.

PR #14, opened 2018-08-14
(SERVER-2272) Refactor action specs out of cli_spec

amysahli:


lib/puppetserver/ca/revoke_action.rb
```diff
@@ -40,7 +40,7 @@ def self.parser(parsed = {})
             parsed['config'] = conf
           end
           o.on('--help', 'Displays this revoke specific help output') do |help|
-            parsed['help'] = help
+            parsed['help'] = true
```
amysahli: @justinstoller This should be true, right?  Since it doesn't accept any arguments

justinstoller: If this block is called `OptionParser` will pass true in as the argument. These two are equivalent, though I'm unsure which is more idiomatic.

justinstoller: I think if we accept the argument `|help|` above it makes sense to set it to that value. However we can leave the argument off and OptionParser will be fine with it, eg:
```rb
irb(main):007:0> parser = OptionParser.new {|o| o.on('--help') {|h| results['help'] = h } }
irb(main):008:0> parser.parse(['--help'])
=> []
irb(main):009:0> results
=> {"help"=>true}

# or
irb(main):013:0> results.delete('help')
=> true
irb(main):014:0> results
=> {}

irb(main):011:0> parser = OptionParser.new {|o| o.on('--help') { results['help'] = true } }
irb(main):015:0> parser.parse(['--help'])
=> []
irb(main):016:0> results
=> {"help"=>true}

```

amysahli: ok.  Half our code was "true" and 1/2 was "help".  So I wanted to be consistent.

PR #15, opened 2018-08-14
(SERVER-2273) Refactor puppet_config to use a defaults hash

amysahli:


lib/puppetserver/ca/puppet_config.rb
```diff
@@ -84,60 +84,68 @@ def resolve_settings(overrides = {})
...
+        defaults = {
+          # Order for base settings here matters!
+          # These need to be evaluated before we can construct their dependent
+          # defaults below
+          :base => {
+            :confdir => user_specific_conf_dir,
+            :ssldir =>'$confdir/ssl',
+            :cadir =>'$ssldir/ca',
+            :certdir =>'$ssldir/certs',
+            :certname => @certname,
```
justinstoller: While you're cleaning up this class can we memoize the value computed in `user_specific_conf*` and wrap `@certname` in a similar reader method?

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -84,60 +84,68 @@ def resolve_settings(overrides = {})
...
-        settings[:certname] = substitutions['$certname'] = overrides.fetch(:certname, @certname)
-
-        server = overrides.fetch(:server, '$certname')
-        settings[:server] = substitutions['$server'] = server.sub(unresolved_setting, substitutions)
-
-        settings[:masterport] = substitutions['$masterport'] = overrides.fetch(:masterport, '8140')
-
-        settings[:ca_name] =  overrides.fetch(:ca_name, 'Puppet CA: $certname')
-        settings[:root_ca_name] = overrides.fetch(:root_ca_name, "Puppet Root CA: #{SecureRandom.hex(7)}")
+        defaults = {
```
justinstoller: Since we never operate on `defaults` as a single collection and the two keys contain collections that are importantly different structurally I could see splitting this into two variables like an Array for `base` and a Hash for `dependent`.

But on the other hand its nice to call them all out as defaults and have the variable names and structure provide clear readability.

I could go either way.

lib/puppetserver/ca/puppet_config.rb
```diff
@@ -179,6 +187,14 @@ def run(command)
         %x( #{command} )
       end
 
+      def munge_ttl_setting(ca_ttl_setting)
```
justinstoller: Can we move the constants `TTLSetting::UNITMAP` and `TTLSetting::FORMAT` into this class and then make this method have the same body as `TTLSetting#munge` and then delete the `TTLSetting` class?

lib/puppetserver/ca/puppet_config.rb
amysahli: Yeah, I tried both in my iterations.  I'll try to see if I can improve it.

PR #16, opened 2018-08-15
(maint) Configure HTTPS connection with agent identity

justinstoller:


justinstoller:
/cc @Magisus This is what I got working on your VM. I think I've functionally tested it locally (updated puppetserver.conf with a repl session), I'll look into writing more tests.

I haven't been able to replicate the signeddir and csrdir being owned by root that I saw on your VM (yet). The generate action doesn't create either of those directories, maybe agent created them accidentally? Should we create those during the `generate` and `import` actions? I can add that to this PR.

Magisus:
On a clean install on a new VM, where I installed puppetserver and this branch of the gem, updated `auth.conf`, then ran `generate`, then started puppetserver, the `ca/requests` and `ca/signed` dirs are both owned by root. However, that didn't seem to impede any of the tasks I tried: `list`, `sign`, and `clean` all work as expected.

Then I deleted the CA dir and recreated it the old way, by starting puppetserver using `puppetserver start`. In this case, ALL the ca files were owned by root. All the CLI tasks still worked as expected.

So even though I wasn't able to replicate bad server responses, I agree that we should be consistent one way or another, whether that means updating puppetserver or the gem, to create all the dirs with the same permissions. Do you think it's expected that the permissions be for `puppet` or for `root`?

Magisus:
So I guess this needs to be accompanied by a puppetserver PR updating the default `auth.conf`. Do we want to try restricting the `certificate_status` endpoint to `localhost`?

PR #17, opened 2018-08-15
(SERVER-2265) Refactor utils, configs, actions into separate dirs

lazartea:
This commit separates the utils, configs, and actions into their own directories.
It also renames files to be more consistant.


lib/puppetserver/ca/utils/parse.rb
```diff
@@ -0,0 +1,82 @@
+module Puppetserver
+  module Ca
+    module Utils
+      module Parse
```
lazartea: How do we feel about this name change? I figured ``Puppetserver::Ca::Utils::Utils`` didn't make much sense.

lib/puppetserver/ca/actions/clean.rb
```diff
@@ -0,0 +1,159 @@
...
+require 'puppetserver/ca/utils/file_system'
+require 'puppetserver/ca/config/puppet'
+require 'puppetserver/ca/actions/revoke'
+
+require 'optparse'
+require 'json'
+
+module Puppetserver
+  module Ca
+    module Actions
```
Magisus: For namespacing like this, should it be `Action` or `Actions`?

lib/puppetserver/ca/utils/config.rb
```diff
@@ -0,0 +1,13 @@
+module Puppetserver
+  module Ca
+    module Utils
```
Magisus: How do people feel about putting all parent modules on one line to reduce indentation, e.g. 
```
module Puppetserver::Ca::Utils
  module Config
  end
end
```

lib/puppetserver/ca/utils/parse.rb
Magisus: I would prefer something like `Cli` maybe.

lib/puppetserver/ca/actions/clean.rb
```diff
@@ -0,0 +1,159 @@
...
+      class Clean
+
+        include Puppetserver::Ca::Utils::Parse
```
justinstoller: by including the whole module you have access to the methods defined in it w/o having to namespace them.

I think we should just include the `Puppetserver::Ca::Utils` module, we can then directly refer to constants defined within it (eg we can just use `Parse.parse_with_errors` below)

lib/puppetserver/ca/actions/clean.rb
justinstoller: I think the coherence of the object within the namespace helps. Like I think of `Utils::FileSystem` as a collection of utility methods for interacting with the filesystem. That kinda breaks down with `HttpClient` as a singular entity vs a collection of methods. With `Actions::Clean` I would expect to be a collection of methods for cleaning vs `Action::Clean` as a singular clean action. I guess its really that I think of the parent module as providing some description of the child object vs describing itself...

tl;dr: I'd go with `Action`, and I could go either way on `Utils`.

lib/puppetserver/ca/utils/config.rb
justinstoller: no feelings either way

lib/puppetserver/ca/utils/parse.rb
justinstoller: I wouldn't go with just `Cli` since there's a class named that in the higher level `Puppetserver::Ca` module. But maybe `CliParsing` or something?

lazartea:
@justinstoller is this still broken? When you look at individual commits, you can see that each file was renamed, and then changed. But when we look at the total changes, it's showing deleted files again... Will we still retain the git history?

justinstoller:
It looks great, though we should have `(SERVER-2265)` at the start of each commit

lib/puppetserver/ca/utils/config.rb
justinstoller: We played with this and combining the names messed with Ruby's constant lookup so we went with the old way.

lazartea:
@Magisus @justinstoller fixed rebase issue, and tested locally. Looks like it works!

PR #19, opened 2018-08-17
(maint) Allow 200 response to process correctly

amysahli:
When --all flag is included we ask for all csrs from
the CA.  Code is returned in string format so we must
read it as such.


Magisus:
We usually use the `(maint)` header for things like this.

PR #20, opened 2018-08-17
(SERVER-2268) Update error messaging

amysahli:


lib/puppetserver/ca/create_action.rb
```diff
@@ -169,7 +169,7 @@ def check_submit_result(result, certname)
           return true
         else
           @logger.err 'Error:'
-          @logger.err "    When certificate request submitted for #{certname}:"
+          @logger.err "    When certificate request submitted for #{certname} received:"
```
justinstoller: This is missing single quotes around the certname.

justinstoller:
Adding to the commit message why we're updating the error messaging would be helpful. We also hit this endpoint when revoking. You might want to wait until after we land https://github.com/puppetlabs/puppetserver-ca-cli/pull/17 and rebase.

lib/puppetserver/ca/action/sign.rb
```diff
@@ -147,7 +147,7 @@ def sign_requested_certs(certnames,settings)
               success = false
             else
               @logger.err 'Error:'
-              @logger.err "    When download requested for #{result.inspect}"
+              @logger.err "    When attempting to sign certificate request '#{certname}' received:"
```
amysahli: My main reason for this PR was this typo...but while I was here I thought I would standardize all our sign error messaging.

lib/puppetserver/ca/action/create.rb
```diff
@@ -203,7 +203,7 @@ def check_sign_result(result, certname)
             return true
           else
             @logger.err 'Error:'
-            @logger.err "    When signing request submitted for #{certname}:"
+            @logger.err "    When attempting to sign certificate request '#{certname}' received:"
             @logger.err "      code: #{result.code}"
```
Magisus: Can we phrase all of them like this one? "When attempting to do X, received:"

lib/puppetserver/ca/action/sign.rb
justinstoller: Gotcha, I was thinking (hence my comment about revoke) that the updates to puppetserver to support SANs affect all PUTs to the certificate_status endpoint similarly and that this PR re-evaluate all those error messages.

Breaking it up along user facing lines, sign, revoke, create, etc makes sense (vs implementation lines like http methods to an endpoint).

lib/puppetserver/ca/action/create.rb
amysahli: sure

amysahli:
Closing this as the work will be contained in other upcoming tickets.

PR #21, opened 2018-08-20
(SERVER-2286) Create a master cert as part of `generate` action

Magisus:
This commit updates the `generate` action to also create a master
certificate signed by the intermediate CA. This cert support subect
alternative names that can be customized via Puppet's dns_alt_names
setting or the `--subject-alt-names` flag.

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -4,6 +4,7 @@ require 'puppetserver/ca/utils/file_system'
 require 'puppetserver/ca/host'
 require 'puppetserver/ca/utils/cli_parsing'
 require 'puppetserver/ca/utils/signing_digest'
+require 'facter'
 
 module Puppetserver
   module Ca
@@ -15,7 +16,20 @@ module Puppetserver
           ["basicConstraints", "CA:TRUE", true],
           ["keyUsage", "keyCertSign, cRLSign", true],
           ["subjectKeyIdentifier", "hash", false],
+          ["nsComment", "Puppet Server Internal Certificate", false],
           ["authorityKeyIdentifier", "keyid:always", false]
+        ].freeze
+
+        SSL_SERVER_CERT = "1.3.6.1.5.5.7.3.1"
+        SSL_CLIENT_CERT = "1.3.6.1.5.5.7.3.2"
+
+        MASTER_EXTENSIONS = [
+          ["basicConstraints", "CA:FALSE", true],
+          ["nsComment", "Puppet Server Internal Certificate", false],
+          ["authorityKeyIdentifier", "keyid:always", false],
+          ["extendedKeyUsage", "#{SSL_SERVER_CERT}, #{SSL_CLIENT_CERT}", true],
+          ["keyUsage", "keyEncipherment, digitalSignature", true],
+          ["subjectKeyIdentifier", "hash", false]
         ].freeze
 
         # Make the certificate valid as of yesterday, because so many people's
@@ -30,10 +44,17 @@ module Puppetserver
 Usage:
   puppetserver ca generate [--help]
   puppetserver ca generate [--config PATH]
+  puppetserver ca generate [--subject-alt-names ALTNAME1[,ALTNAME2...]]
 
 Description:
 Generate a root and intermediate signing CA for Puppet Server
 and store generated CA keys, certs, and crls on disk.
+
+The `--subject-alt-names` flag can be used to add SANs to the
+certificate generated for the Puppet master. Multiple names can be
+listed as a comma separated string. These can be either DNS names or
+IP addresses, differentiated by prefixes: `DNS:foo.bar.com,IP:123.456.789`.
+Names with no prefix will be treated as DNS names.
 
 To determine the target location, the default puppet.conf
 is consulted for custom values. If using a custom puppet.conf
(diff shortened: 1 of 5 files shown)


lib/puppetserver/ca/action/generate.rb
```diff
@@ -87,16 +112,35 @@ def generate_root_and_intermediate_ca(settings, signing_digest)
...
           int_crl = create_crl_for(int_cert, int_key, valid_until, signing_digest)
 
+          master_key = host.create_private_key(settings[:keylength])
+          master_csr = host.create_csr(settings[:certname], master_key)
+          master_cert = sign_master_cert(int_key, int_cert, master_csr,
+                                       valid_until, signing_digest, subject_alt_names)
+
           FileSystem.ensure_dir(settings[:cadir])
+          FileSystem.ensure_dir(settings[:certdir])
+          FileSystem.ensure_dir(settings[:privatekeydir])
```
justinstoller: Should we create the publickeydir/pubkeys too?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -168,25 +212,77 @@ def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest)
...
+        # 1) Specified on the CLI
+        # 2) Specified in settings
+        # 3) Defaults
+        def choose_alt_names(cli_alt_names, settings_alt_names, certname)
+          if !cli_alt_names.empty?
+            sans = cli_alt_names
+          elsif !settings_alt_names.empty?
+            sans = settings_alt_names
+          else
+            sans = "puppet, #{certname}"
```
justinstoller: Should this default live in PuppetConfig?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -168,25 +212,77 @@ def sign_intermediate(ca_key, ca_cert, csr, valid_until, signing_digest)
...
-        def parse(args)
-          results = {}
-          parser = self.class.parser(results)
+        def sign_master_cert(int_key, int_cert, csr, valid_until, signing_digest, subject_alt_names)
+          cert = OpenSSL::X509::Certificate.new
+          cert.public_key = csr.public_key
+          cert.subject = csr.subject
+          cert.issuer = int_cert.subject
+          cert.version = 2
+          cert.serial = 1
```
justinstoller: We should update the serials and inventory files now

lib/puppetserver/ca/action/generate.rb
Magisus: Oh yeah, meant to look into that.

lib/puppetserver/ca/action/generate.rb
Magisus: Oh yeah, that would make sense.

lib/puppetserver/ca/action/generate.rb
Magisus: Added.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -120,13 +121,22 @@ def generate_pki(settings, signing_digest, subject_alt_names = '')
           FileSystem.ensure_dir(settings[:cadir])
           FileSystem.ensure_dir(settings[:certdir])
           FileSystem.ensure_dir(settings[:privatekeydir])
+          FileSystem.ensure_dir(settings[:publickeydir])
+
+          FileSystem.write_file(settings[:cert_inventory],
+                                "0x0001 #{format_time(master_cert.not_before)} " +
```
justinstoller: nit: this is a weird bit of domain knowledge that would be good to put into a method

Magisus: okay

lib/puppetserver/ca/action/generate.rb
```diff
@@ -62,19 +83,24 @@ def run(input)
           signer = SigningDigest.new
           return 1 if CliParsing.handle_errors(@logger, signer.errors)
 
+          if input['subject_alt_names'].empty?
+            subject_alt_names = munge_alt_names(puppet.settings[:subject_alt_names])
+          else
+            subject_alt_names = munge_alt_names(input['subject_alt_names'])
+          end
+
```
amysahli: I think it would be easier to read something like:
```
alt_names_to_munge = input['subject_alt_names'].empty? puppet.settings[:subject_alt_names] : input['subject_alt_names']
subject_alt_names = munge_alt_names(alt_names_to_munge)
```

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -151,6 +155,9 @@ def resolve_settings(overrides = {})
           settings[:ca_ttl] = munge_ttl_setting(settings[:ca_ttl])
           settings[:certificate_revocation] = parse_crl_usage(settings[:certificate_revocation])
 
+          # rename dns_alt_names to subject_alt_names now that we support IP alt names
+          settings[:subject_alt_names] = overrides.fetch(:dns_alt_names, "puppet,$certname")
```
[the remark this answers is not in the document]
Magisus: I don't think it works because the name of the setting I'm pulling out of Puppet is different than the name I'm saving it under. But maybe I'm missing something?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -62,19 +83,24 @@ def run(input)
...
           # Generate root and intermediate ca and put all the certificates, crls,
           # and keys where they should go.
-          generate_root_and_intermediate_ca(puppet.settings, signer.digest)
+          generate_pki(puppet.settings, signer.digest, subject_alt_names)
 
           # Puppet's internal CA expects these file to exist.
-          FileSystem.ensure_file(puppet.settings[:serial], "001", 0640)
-          FileSystem.ensure_file(puppet.settings[:cert_inventory], "", 0640)
+          FileSystem.ensure_file(puppet.settings[:serial], "002", 0640)
```
amysahli: This is 002 because we created Master with 001?

Magisus: Yes.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -62,19 +83,24 @@ def run(input)
...
+          end
+
           # Generate root and intermediate ca and put all the certificates, crls,
           # and keys where they should go.
-          generate_root_and_intermediate_ca(puppet.settings, signer.digest)
+          generate_pki(puppet.settings, signer.digest, subject_alt_names)
 
           # Puppet's internal CA expects these file to exist.
-          FileSystem.ensure_file(puppet.settings[:serial], "001", 0640)
-          FileSystem.ensure_file(puppet.settings[:cert_inventory], "", 0640)
```
amysahli: Why dont we need to ensure this file anymore? (cert_inventory)?  Was it never needed?

lib/puppetserver/ca/action/generate.rb
Magisus: I went back and forth on this. I didn't like doing the munge in-line, but I'm also not sure about creating the other throwaway variable. But I'll change it if that seems better.

lib/puppetserver/ca/action/generate.rb
Magisus: Because we're writing it explicitly as part of generating the certs. It needs data from the cert object, so I moved it into the method where those are available.

lib/puppetserver/ca/config/puppet.rb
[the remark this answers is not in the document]
amysahli: You are right.  I didnt even notice they were different

lib/puppetserver/ca/action/generate.rb
amysahli: Gotcha

lib/puppetserver/ca/action/generate.rb
justinstoller: I don't mind the temp variable, though I don't think we should have a 120 character ternary.

Magisus: The length was one of the main reasons I went this way instead. Even though I also don't love variables that are only assigned inside if blocks 🤷‍♀️

PR #18, opened 2018-08-15
(maint) Prevent the `generate` command from overwriting the CA

Magisus:
Previously, the `generate` command would overwrite any existing CA files
in the target locations. This meant that if a user called
`generate` more than once, they could accidentally nuke thier CA,
thereby making all certs issued by it invalid. This commit makes the
command exit if any of the CA files already exist, and prompts the user to
remove them manually if they really want to regenerate their CA.

PR #22, opened 2018-08-21
(maint) Prevent `generate` and `import` from overwriting existing CA files

Magisus:
Supersedes https://github.com/puppetlabs/puppetserver-ca-cli/pull/18

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -131,7 +131,7 @@ BANNER
             [settings[:hostpubkey], master_key.public_key],
             [settings[:capub], int_key.public_key],
             [settings[:cert_inventory], inventory_entry(master_cert)],
-            [settings[:serial], "002"],
+            [settings[:serial], "0x0002"],
           ]
 
           private_files = [
@@ -140,13 +140,16 @@ BANNER
             [settings[:cakey], int_key],
           ]
 
-          errors = FileSystem.check_for_existing_files(public_files.map { |f| f.first })
-          errors += FileSystem.check_for_existing_files(private_files.map { |f| f.first })
+          errors = FileSystem.check_for_existing_files(public_files.map(&:first))
+          errors += FileSystem.check_for_existing_files(private_files.map(&:first))
 
-          if errors.any?
-            errors << "If you would really like to replace your CA, please delete the existing files first.
-  Note that any certificates that were issued by this CA will become invalid if you
-  replace it!"
+          if !errors.empty?
+            instructions = <<-ERR
+If you would really like to replace your CA, please delete the existing files first.
+Note that any certificates that were issued by this CA will become invalid if you
+replace it!
+ERR
+            errors << instructions
             return errors
           end
 
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -4,6 +4,7 @@ require 'puppetserver/ca/utils/file_system'
 require 'puppetserver/ca/host'
 require 'puppetserver/ca/utils/cli_parsing'
 require 'puppetserver/ca/utils/signing_digest'
+require 'puppetserver/ca/config/puppet'
 require 'facter'
 
 module Puppetserver
@@ -91,10 +92,8 @@ BANNER
 
           # Generate root and intermediate ca and put all the certificates, crls,
           # and keys where they should go.
-          generate_pki(puppet.settings, signer.digest, subject_alt_names)
-
-          # Puppet's internal CA expects these file to exist.
-          FileSystem.ensure_file(puppet.settings[:serial], "002", 0640)
+          errors = generate_pki(puppet.settings, signer.digest, subject_alt_names)
+          return 1 if CliParsing.handle_errors(@logger, errors)
 
           @logger.inform "Generation succeeded. Find your files in #{puppet.settings[:cadir]}"
           return 0
@@ -123,18 +122,16 @@ BANNER
           FileSystem.ensure_dir(settings[:privatekeydir])
           FileSystem.ensure_dir(settings[:publickeydir])
 
-          FileSystem.write_file(settings[:cert_inventory],
-                                inventory_entry(master_cert),
-                                0644)
-
           public_files = [
             [settings[:cacert], [int_cert, root_cert]],
             [settings[:cacrl], [int_crl, root_crl]],
             [settings[:hostcert], master_cert],
             [settings[:localcacert], [int_cert, root_cert]],
             [settings[:localcacrl], [int_crl, root_crl]],
-            [settings[:hostpubkey], [master_key.public_key]],
-            [settings[:capub], [int_key.public_key]],
+            [settings[:hostpubkey], master_key.public_key],
+            [settings[:capub], int_key.public_key],
+            [settings[:cert_inventory], inventory_entry(master_cert)],
+            [settings[:serial], "002"],
           ]
 
           private_files = [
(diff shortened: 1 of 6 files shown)


lib/puppetserver/ca/action/generate.rb
```diff
@@ -143,15 +140,25 @@ def generate_pki(settings, signing_digest, subject_alt_names = '')
             [settings[:cakey], int_key],
           ]
 
+          errors = FileSystem.check_for_existing_files(public_files.map { |f| f.first })
```
justinstoller: little golf-y but `(&:first)` works too.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -143,15 +140,25 @@ def generate_pki(settings, signing_digest, subject_alt_names = '')
...
+          errors += FileSystem.check_for_existing_files(private_files.map { |f| f.first })
+
+          if errors.any?
```
justinstoller: Definitely more readable but I'm still don't love using `any?` like this

lib/puppetserver/ca/action/import.rb
```diff
@@ -65,40 +79,29 @@ def run(input)
...
+          results = {}
+          parser = self.class.parser(results)
 
-        errors = CliParsing.parse_with_errors(parser, args)
+          errors = CliParsing.parse_with_errors(parser, args)
 
-        if check_flag_usage(results)
-          errors << check_flag_usage(results)
-        end
+          if check_flag_usage(results)
```
justinstoller: I think elsewhere we updated this to be
```rb
if error = check_flag_usage(results)
  errrors << error
end
```
if we did that we probably want to do the same here for consistency, though I could see us preferring
```rb
error = check_flag_usage(results)
errors << error if error
```

lib/puppetserver/ca/action/generate.rb
```diff
@@ -123,18 +122,16 @@ def generate_pki(settings, signing_digest, subject_alt_names = '')
...
             [settings[:cacrl], [int_crl, root_crl]],
             [settings[:hostcert], master_cert],
             [settings[:localcacert], [int_cert, root_cert]],
             [settings[:localcacrl], [int_crl, root_crl]],
-            [settings[:hostpubkey], [master_key.public_key]],
-            [settings[:capub], [int_key.public_key]],
+            [settings[:hostpubkey], master_key.public_key],
+            [settings[:capub], int_key.public_key],
+            [settings[:cert_inventory], inventory_entry(master_cert)],
+            [settings[:serial], "002"],
```
justinstoller: nit: looking at the code we actually do the inventory serial number format here as well in the puppetserver code. We might want to do that here or as a separate ticket (kind of an unrelated bug fix).

lib/puppetserver/ca/action/generate.rb
```diff
@@ -143,15 +140,25 @@ def generate_pki(settings, signing_digest, subject_alt_names = '')
             [settings[:cakey], int_key],
           ]
 
+          errors = FileSystem.check_for_existing_files(public_files.map { |f| f.first })
+          errors += FileSystem.check_for_existing_files(private_files.map { |f| f.first })
+
+          if errors.any?
+            errors << "If you would really like to replace your CA, please delete the existing files first.
```
justinstoller: This is the preferred way to do it in clojure, but for some reason I thought HEREDOCs were the preferred way to do multiline strings with newlines in them in Ruby. I can't find any consensus on the Internet or even recommendations in Rubocop. So I'm probably wayy off on that. Do you know if there's any preference within the company?

lib/puppetserver/ca/action/generate.rb
Magisus: Oh yeah, always forget about that.

lib/puppetserver/ca/action/generate.rb
Magisus: I missed the conversation on why that is last time. What don't you like about it? (I'm happy to change it, just would like to know what the misgivings are.)

lib/puppetserver/ca/action/import.rb
Magisus: I like the explicit block better, I kind of don't like the post-condition syntax most of the time. I'll update this to be like it is elsewhere.

lib/puppetserver/ca/action/generate.rb
Magisus: Do you mean we just shouldn't duplicate the logic? Or is there some formatting that is missing here? I think we might need it in both places because most signing will go through the API...

lib/puppetserver/ca/action/generate.rb
Magisus: I think a HEREDOC would be more  idiomatic. I'll update.

lib/puppetserver/ca/action/generate.rb
justinstoller: I think the format for the serial number is the same in the serial file as it is in inventory file: https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/puppetserver/certificate_authority.clj#L373

lib/puppetserver/ca/action/generate.rb
justinstoller: Because it's only the same as `!empty?` if the collection can contain _only_ truthy values).

When it's unambiguous that the collection can only contain truthy values (like here) then I'm fine with it.

My not-love for it comes from a concern that using it too freely will encourage us in the future to use it as a synonym for not-empty even in collections that can contain falsey values.

lib/puppetserver/ca/action/generate.rb
Magisus: Ah, gotcha. I'll fix that.

lib/puppetserver/ca/action/generate.rb
amysahli: ^ @justinstoller why don't you like any here?

spec/puppetserver/ca/action/import_spec.rb
```diff
@@ -274,13 +215,9 @@
...
-        exit_code = Puppetserver::Ca::Cli.run(['import',
-                                                '--private-key', key_file,
-                                                '--cert-bundle', bundle_file,
-                                                '--crl-chain', chain_file],
-                                                stdout,
-                                                stderr)
-
+        exit_code = subject.run({ 'cert-bundle' => bundle_file,
+                                    'private-key'=> key_file,
+                                    'crl-chain' => chain_file })
```
amysahli: not important...jsut small spacing issue.

spec/puppetserver/ca/action/generate_spec.rb
```diff
@@ -21,7 +19,7 @@
   subject { Puppetserver::Ca::Action::Generate.new(logger) }
 
   it 'prints the help output & returns 1 if invalid flags are given' do
-    exit_code = Puppetserver::Ca::Cli.run(['generate', '--hello'], stdout, stderr)
+    _, exit_code = subject.parse({ '--hello' => ''})
```
amysahli: I notice elsewhere we use parse with an array like: ['--hello', '']

lib/puppetserver/ca/action/import.rb
```diff
@@ -58,47 +75,36 @@ def run(input)
           FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0640)
 
           # Puppet's internal CA expects these file to exist.
-          FileSystem.ensure_file(puppet.settings[:serial], "001", 0640)
+          FileSystem.ensure_file(puppet.settings[:serial], "0x0001", 0640)
```
amysahli: what's this  "0x0001" about?

Magisus: The serials are supposed to be in hex, to maintain what puppet is doing now.

spec/puppetserver/ca/action/generate_spec.rb
Magisus: Oh right, I should change this. I remember there being a comment before about being surprised this works :P

Magisus:
Updated.

PR #23, opened 2018-08-21
(SERVER-2276) Make a CA object to wrap HTTP calls

amysahli:


lib/puppetserver/ca/ca.rb
```diff
@@ -0,0 +1,210 @@
...
+
+      def sign_certs(certnames)
+        put('certificate_status', certnames, SIGN_BODY, :sign)
+      end
+
+      def revoke_certs(certnames)
+        put('certificate_status', certnames, REVOKE_BODY, :revoke)
+      end
+
+      # Should this csr be in JSON? instead of string?
```
amysahli: I'm sending the csr as a string, which works.  Should it be  JSON.dump like the other bodies?

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -0,0 +1,210 @@
...
+
+      def sign_certs(certnames)
+        put('certificate_status', certnames, SIGN_BODY, :sign)
+      end
+
+      def revoke_certs(certnames)
+        put('certificate_status', certnames, REVOKE_BODY, :revoke)
+      end
+
+      # Should this csr be in JSON? instead of string?
```
amysahli: I'm sending the csr as a string, which works. Should it be JSON.dump like the other bodies?

lib/puppetserver/ca/action/sign.rb
```diff
@@ -74,45 +73,12 @@ def run(input)
...
         def get_all_certs(settings)
-          result = get_certificate_statuses(settings)
-
-          unless result.code == '200'
-              @logger.err 'Error:'
-              @logger.err "    #{result.inspect}"
-              return nil
-          end
-          return result
+          ca(settings).get_certificate_statuses
```
justinstoller: Since there's not much in here any more should we just fold these up into `#run` or a method that run calls?

lib/puppetserver/ca/certificate_authority.rb
justinstoller: Should it be JSON? it looks like it csr is an OpenSSL object? I think then `to_s` is alaised to  `to_pem`? The docs say that it should be sent as a `Content-Type` `plain/text` but contain just the pem encoded CSR: https://puppet.com/docs/puppet/5.5/http_api/http_certificate_request.html#save

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -0,0 +1,210 @@
...
+
+      # logs the action and returns true/false for success
+      def process_results(type, certname, result)
+        case type
+        when :sign
+          case result.code
+          when '204'
+            @logger.inform "Successfully signed certificate request for #{certname}"
+            return true
+          else
```
justinstoller: Is there any result that can't return 404? We'd also want it to be an error in most cases as well, right?

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -0,0 +1,210 @@
...
+            @logger.err 'Error:'
+            @logger.err "    When signing request submitted for #{certname}:"
```
justinstoller: It seems like almost all of these are variations on:
success:
```
@logger.inform "Was successful in #{verb} #{object} for #{certname}"
```
error:
```
@logger.err 'Error:'
@logger.err "    When #{verb} #{object} for #{certname} received:"
# ...
```

Would it make it easier to read to refactor that out? Or would do you think it'd be a bit too obscure?

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -0,0 +1,210 @@
...
+        url = make_ca_url(resource_type)
+        results = @client.with_connection(url) do |connection|
+          certnames.map do |certname|
+            url.resource_name = certname
+            result = connection.put(body, url)
+            process_results(type, certname, result)
+          end
+        end
+
+        results.all?
```
justinstoller: if we changed this to `results.all? {|result| result == :success }` could we combine the check methods that return a symbol and the process methods?

lib/puppetserver/ca/certificate_authority.rb
Magisus: We definitely use `text/plain` in the agent.

lib/puppetserver/ca/certificate_authority.rb
Magisus: If we do it like this, it becomes impossible to translate. Since we might still want that for this gem one day, I think we should leave it more explicit.

Magisus: See https://confluence.puppetlabs.com/display/enterprise/Translation-friendly+strings for guidelines on writing translatable strings.

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -0,0 +1,210 @@
...
+            unless revoked == :error
+              clean_result = connection.delete(url)
+              cleaned = check_clean(certname, clean_result)
+            end
+
+            cleaned == :success && [:success, :not_found].include?(revoked)
+          end
+        end
+
+        return results.all?
```
Magisus: Agree with Justin, I think we should move the verification into a block passed to `all`. I find this dangling call here at the end hard to read.

lib/puppetserver/ca/certificate_authority.rb
justinstoller: Good catch, and thanks for the link!

I imagine that making a giant table of strings and turning this into a bunch of selectors would make this impenetrable?

lib/puppetserver/ca/certificate_authority.rb
amysahli: ^ Sorry Justin.  I don't undertand.  If its not 204, it will log the error.

justinstoller: My bad, I was being obtuse. I meant that if 404 was a valid error state for this endpoint (which I think it's a valid response from the server for all endpoints), then the existing 404 handling would give the user a better error message then this more general one, and potentially allow us to fold this in with other result processing.

justinstoller: Though if we have to duplicate all of our success and error messages combining them may not be much of a win.

amysahli: Lemme take a look about adding it back.

lib/puppetserver/ca/certificate_authority.rb
amysahli: I'm updated it to be .to_pem for clarity and have this call use custom headers with content-type text/plain

lib/puppetserver/ca/action/sign.rb
```diff
@@ -70,49 +69,18 @@ def run(input)
...
-          client = http_client(settings)
-          url = client.make_ca_url(settings[:ca_server],
-                                   settings[:ca_port],
-                                   'certificate_statuses',
-                                   'any_key')
-          client.with_connection(url) do |connection|
-            connection.get(url)
-          end
+        def ca(settings)
+          @ca ||= Puppetserver::Ca::CertificateAuthority.new(@logger, settings)
```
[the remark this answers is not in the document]
Magisus: I kind of like creating it in `run` and then passing it, the way we do with `settings`. I don't think we would have to pass `settings` to `get_all_pending_certs` at all if we did that, and we could call `sign_certs` on it directly up above.

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -59,29 +54,32 @@ def initialize(net_http_connection, url_struct)
             @url = url_struct
           end
 
-          def get(url_overide = nil)
+          def get(url_overide = nil, headers = {})
             url = url_overide || @url
+            headers = DEFAULT_HEADERS.merge(headers)
```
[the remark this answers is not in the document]
Magisus: Since this isn't a destructive operation, seems fine to me. You aren't really manipulating it, you're creating a new copy.

lib/puppetserver/ca/certificate_authority.rb
amysahli: ok.  I remember now
it was:
```
  when '404'
              @logger.err 'Error:'
              @logger.err "    Could not find certificate for #{certname}"
              success = false
```
But if there are SANs we also get a 404, and it that case we want the custom error message that the endpoint sends us and not something like "could not find certificate".  Thats why i changed it.

lib/puppetserver/ca/utils/http_client.rb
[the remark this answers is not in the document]
amysahli: ^ Yeah.  Justin also mentiioned I'm probably thinking about merge!  NOT merge.  So all is good.

lib/puppetserver/ca/action/sign.rb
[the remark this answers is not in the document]
amysahli: ^ Yep.  That looks way better.  Updated

lib/puppetserver/ca/certificate_authority.rb
amysahli: added back.

Magisus:
I'll try this out in packages tomorrow.

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -0,0 +1,224 @@
...
+            type: :submit)
+      end
+
+      # Make an HTTP PUT request to CA
+      # @param resource_type [String] the resource type of url
+      # @param certnames [Array] array of certnames
+      # @param body [JSON/String] body of the put request
+      # @param type [Symbol] type of error processing to perform on result
+      # @return [Boolean] whether all requests were successful
+      def put(certnames, resource_type:, body:, type:, headers: nil)
```
Magisus: `headers` needs to default to the empty hash. If it's nil, when it gets passed to `DEFAULT_HEADERS.merge(headers)` inside the HTTPClient, it will crash.

PR #25, opened 2018-08-24
(SERVER-2288) Generate a master certificate as part of the import action

amysahli:


amysahli:
This also adds subject-alt-names to Import action while i'm in there.

lib/puppetserver/ca/action/import.rb
```diff
@@ -48,41 +51,71 @@ def run(input)
...
+            [settings[:cacert], loader.certs],
+            [settings[:cacrl], loader.crls],
+            [settings[:localcacert], loader.certs],
+            [settings[:hostcrl], loader.crls],
+            [settings[:hostpubkey], master_key.public_key],
+            [settings[:hostcert], master_cert],
+            [settings[:cert_inventory], ca.inventory_entry(master_cert)],
+            [settings[:serial], "0x0002"],
+            [File.join(settings[:signeddir], "#{settings[:certname]}.pem"), master_cert]
+          ]
```
amysahli: So added in here localcacert and hostcrl (same as cacert and cacrl) along with the Master cert stuff we need (hostcert, hostpubkey) Also, instead of ensuring the serial and cert inventory, I just create them since we have custom stuff we want in there 
(instead of
```
 FileSystem.ensure_file(puppet.settings[:serial], "001", 0640)
 FileSystem.ensure_file(puppet.settings[:cert_inventory], "", 0640)
```
)

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -126,9 +126,8 @@ def resolve_settings(overrides = {})
             :ca_server => '$server',
             :ca_port => '$masterport',
             :localcacert => '$certdir/ca.pem',
-            :localcacrl => '$ssldir/crl.pem',
```
amysahli: removing localcacrl as this is the same as hostcrl.

lib/puppetserver/ca/utils/file_system.rb
```diff
@@ -14,13 +14,9 @@ def self.write_file(*args)
...
         end
 
-        def self.ensure_dir(setting)
-          instance.ensure_dir(setting)
-        end
-
-        def self.ensure_file(location, content, mode)
-          if !File.exist?(location)
-            instance.write_file(location, content, mode)
+        def self.ensure_dirs(one_or_more_settings)
```
amysahli: ^^ removed ensure_file as I removed the only place we ever called it in the import.rb

ensure_dirs now takes multiple settings.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -311,11 +150,11 @@ def self.parser(parsed = {})
             end
             opts.on('--subject-alt-names NAME1[,NAME2]',
                     'Subject alternative names for the master cert') do |sans|
-              parsed['subject_alt_names'] = sans
+              parsed['subject-alt-names'] = sans
```
amysahli: This is just me changing underscores to hyphens for consistency.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -0,0 +1,177 @@
...
+
+      def extension_factory_for(ca, cert = nil)
+        ef = OpenSSL::X509::ExtensionFactory.new
+        ef.issuer_certificate  = ca
+        ef.subject_certificate = cert if cert
+
+        ef
+      end
+
+      def inventory_entry(cert)
```
amysahli: I didn't know where to put this method.  We need it in both import and generate actions.  Any better ideas would be welcome.

spec/puppetserver/ca/action/import_spec.rb
```diff
@@ -239,53 +239,113 @@
...
-  it 'actually, honest to god, moves files' do
+  it 'actually, honest to god, moves/creates files' do
     Dir.mktmpdir do |tmpdir|
       with_files_in tmpdir do |bundle, key, chain, conf|
-        File.open conf, 'w' do |f|
-          f.puts(<<-INI)
-            [master]
-              cadir = #{tmpdir}/ca
-          INI
-        end
```
amysahli: Didn't actually need this, since its in the ```with_files_in``` method itself.

lib/puppetserver/ca/action/import.rb
```diff
@@ -131,6 +166,10 @@ def self.parser(parsed = {})
                     'Common name to use for the master cert') do |name|
               parsed['certname'] = name
             end
+            opts.on('--subject-alt-names NAME1[,NAME2]',
+                    'Subject alternative names for the master cert') do |sans|
+              parsed['subject-alt-names'] = sans
```
justinstoller: should `subject-alt-names` in this file be `subject_alt_names`?

lib/puppetserver/ca/action/import.rb
```diff
@@ -48,41 +51,71 @@ def run(input)
...
+
+          public_files = [
+            [settings[:cacert], loader.certs],
+            [settings[:cacrl], loader.crls],
+            [settings[:localcacert], loader.certs],
+            [settings[:hostcrl], loader.crls],
+            [settings[:hostpubkey], master_key.public_key],
+            [settings[:hostcert], master_cert],
+            [settings[:cert_inventory], ca.inventory_entry(master_cert)],
+            [settings[:serial], "0x0002"],
```
justinstoller: We should not in our docs, or help, that the cert/crl provided to import for the leaf CA must not have already issued or revoked any certs. Otherwise we won't know what the content of the serial and inventory files should be.

lib/puppetserver/ca/config/puppet.rb
justinstoller: good catch, it looks like it's not even a setting https://puppet.com/docs/puppet/5.5/configuration.html

lib/puppetserver/ca/local_certificate_authority.rb
justinstoller: This makes sense. This object creates the content for most of those files.

spec/puppetserver/ca/action/import_spec.rb
```diff
@@ -239,53 +239,113 @@
     end
   end
 
-  it 'actually, honest to god, moves files' do
+  it 'actually, honest to god, moves/creates files' do
```
justinstoller: while your changing it, a less glib test name might be appropriate

lib/puppetserver/ca/action/import.rb
justinstoller: nvm

lib/puppetserver/ca/action/import.rb
```diff
@@ -110,6 +143,8 @@ def parse(args)
         end
 
         def self.parser(parsed = {})
+          parsed['certname'] = ''
```
Magisus: Oh good, this is causing a failure in CI right now, that this wasn't set.

lib/puppetserver/ca/config/puppet.rb
Magisus: My b. Shoulda known better than to assume a naming pattern in Puppet settings :P

lib/puppetserver/ca/utils/file_system.rb
Magisus: Can we call them "dirs" instead of "settings"? Even though we always pass in paths from settings, this method could definitely work on any file path.

lib/puppetserver/ca/utils/file_system.rb
```diff
@@ -6,6 +6,15 @@ module Ca
...
 
+        DIR_MODES = {
+          :ssldir => 0771,
+          :cadir => 0755,
+          :certdir => 0755,
+          :privatekeydir => 0750,
+          :publickeydir => 0755,
+          :signeddir => 0755
+        }
+
```
amysahli: I've added this so that our directories are now created with the correct permissions (as found in the puppet defaults.rb file)

lib/puppetserver/ca/action/import.rb
amysahli: ok!  Added it to the help text above.

PR #26, opened 2018-08-27
(SERVER-2295) Add subject alternate names to create action

amysahli:


lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -82,11 +82,8 @@ def sign_master_cert(int_key, int_cert, csr)
           extension = ef.create_extension(*ext)
           cert.add_extension(extension)
         end
-
-        if !@settings[:subject_alt_names].empty?
-          alt_names_ext = ef.create_extension("subjectAltName", @settings[:subject_alt_names], false)
-          cert.add_extension(alt_names_ext)
-        end
+        alt_names_ext = ef.create_extension("subjectAltName", @settings[:ca_subject_alt_names], false)
+        cert.add_extension(alt_names_ext)
```
amysahli: I removed the ```!@settings[:subject_alt_names].empty?```, since this will ALWAYS have a value as a master cert.

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -152,12 +152,20 @@ def resolve_settings(overrides = {})
...
           # rename dns_alt_names to subject_alt_names now that we support IP alt names
-          settings[:subject_alt_names] = overrides.fetch(:dns_alt_names, "puppet,$certname")
+          settings[:subject_alt_names] = overrides.fetch(:dns_alt_names, "")
+          # If the cert is a master cert, it requires different defaults:
+          settings[:ca_subject_alt_names] = overrides.fetch(:dns_alt_names, "puppet,$certname")
 
           # Some special cases where we need to manipulate config settings:
           settings[:ca_ttl] = munge_ttl_setting(settings[:ca_ttl])
           settings[:certificate_revocation] = parse_crl_usage(settings[:certificate_revocation])
+          settings[:ca_subject_alt_names] = munge_alt_names(settings[:ca_subject_alt_names])
```
amysahli: So, my options were to either provide overrides from the actions, or to do the work here and create a separate SANs for master certs.  I called it ```ca_subject_alt_names```, but I could call it ```master_subject_alt_names``` too.    Ideas to make this better are welcomed.

lib/puppetserver/ca/action/create.rb
```diff
@@ -96,9 +102,15 @@ def run(input)
...
-          puppet = Config::Puppet.parse(config_path)
+          settings_overrides = {}
+          # Since puppet expects the key to be called 'dns_alt_names', we need to use that here
+          # to ensure that the overriding works correctly.
+          settings_overrides[:dns_alt_names] = input['subject-alt-names'] unless input['subject-alt-names'].empty?
+          puppet = Config::Puppet.new(config_path)
+          puppet.load(settings_overrides)
           return 1 if CliParsing.handle_errors(@logger, puppet.errors)
 
+
```
Magisus: extra line here

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -152,12 +152,20 @@ def resolve_settings(overrides = {})
...
+          # If user provides subject-alt-names, we need to add the certname in addition
+          # to these.
+          if overrides[:dns_alt_names]
+            overrides[:dns_alt_names] = overrides[:dns_alt_names] + ',' + '$certname'
+          end
           # rename dns_alt_names to subject_alt_names now that we support IP alt names
-          settings[:subject_alt_names] = overrides.fetch(:dns_alt_names, "puppet,$certname")
+          settings[:subject_alt_names] = overrides.fetch(:dns_alt_names, "")
+          # If the cert is a master cert, it requires different defaults:
+          settings[:ca_subject_alt_names] = overrides.fetch(:dns_alt_names, "puppet,$certname")
```
Magisus: Instead of doing this here with a second setting, can we add the extra `$certname` and supply these defaults if necessary in the `local_ca` object when generating the master cert?

spec/puppetserver/ca/utils/http_client_spec.rb
```diff
@@ -34,6 +34,7 @@
       serial = File.join(tmpdir, 'serial')
 
       settings = {
+        ca_subject_alt_names: 'DNS: puppet, DNS:foo',
```
Magisus: This can probably be removed if we can avoid the extra setting.

lib/puppetserver/ca/config/puppet.rb
amysahli: Great idea.  Thanks!

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -152,8 +152,13 @@ def resolve_settings(overrides = {})
             settings[setting_name] = setting_value
           end
 
+          # If user provides subject-alt-names, we need to add the certname in addition
+          # to these.
+          if overrides[:dns_alt_names]
+            overrides[:dns_alt_names] = overrides[:dns_alt_names] << ',$certname'
+          end
```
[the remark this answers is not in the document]
amysahli: ^ Maggie helped me make this better.

lib/puppetserver/ca/action/create.rb
```diff
@@ -141,6 +152,10 @@ def generate_key_csr(certname, settings, digest)
           private_key = host.create_private_key(settings[:keylength])
           csr = host.create_csr(certname, private_key)
 
+          if !settings[:subject_alt_names].empty?
+            host.add_csr_extension(csr, "subjectAltName", settings[:subject_alt_names])
+          end
+
```
Magisus: So I tried this out, and it doesn't work, I think you need to add the extension before you sign the CSR. So maybe pass the names into `create_csr` and add the extension in there?

amysahli: ^ Yep, that would make sense.  Thanks Maggie

Magisus:
Could you add one other bug fix to this (showed up in CI): we figured out Justin was wrong about the serial here https://github.com/puppetlabs/puppetserver-ca-cli/blob/master/lib/puppetserver/ca/action/import.rb#L93 needing to be in hex. It should just be "002" instead, or else it fails to parse when puppetserver tries to increment it:
```
Internal Server Error: java.lang.NumberFormatException: For input string: "0x0002"   
```

joshcooper:
Can we use a uuid instead? I think there’s a ticket about that.

Magisus:
I'm a little leery of trying to make this change right now, because it will also entail changes in Puppetserver and we're a little pressed for time. The ticket is https://tickets.puppetlabs.com/browse/SERVER-1595, if we end up with any runway (hah) we can think about doing this, unless you think it's high priority from a security perspective?

PR #27, opened 2018-08-30
WIP (SERVER-2297) Read settings from both config files

Magisus:
Note that this is incompatible with 5.x Puppet. We may need to branch the gem.

### DIFF
--- a/lib/puppetserver/ca/action/create.rb
+++ b/lib/puppetserver/ca/action/create.rb
@@ -14,21 +14,23 @@ module Puppetserver
 
         # Only allow printing ascii characters, excluding /
         VALID_CERTNAME = /\A[ -.0-~]+\Z/
-        CERTNAME_BLACKLIST = %w{--all --config}
+        CERTNAME_BLACKLIST = %w{--all --puppet-config --server-config}
 
         SUMMARY = "Create a new certificate signed by the CA"
         BANNER = <<-BANNER
 Usage:
   puppetserver ca create [--help]
-  puppetserver ca create [--config PATH] [--certname CERTNAME[,ADDLCERTNAME]]
+  puppetserver ca create [--puppet-config PATH] [--server-config PATH]
+                         [--certname CERTNAME[,ADDLCERTNAME]]
 
 Description:
 Creates a new certificate signed by the intermediate CA
 and stores generated keys and certs on disk.
 
-To determine the target location, the default puppet.conf
-is consulted for custom values. If using a custom puppet.conf
-provide it with the --config flag
+To determine the target location, the default puppet.conf and
+puppetserver.conf are consulted for custom values. If using custom
+conf files, provide them with the --puppet-config and --server-config
+flags.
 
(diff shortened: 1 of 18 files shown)


Magisus:
This appears to work for basic use cases in conjunction with https://github.com/puppetlabs/puppetserver/pull/1780

Magisus:
Still need to add more split config flags to the other commands.

lib/puppetserver/ca/action/create.rb
```diff
@@ -14,21 +14,23 @@ class Create
...
-        CERTNAME_BLACKLIST = %w{--all --config}
+        CERTNAME_BLACKLIST = %w{--all --puppet-config --server-config}
 
         SUMMARY = "Create a new certificate signed by the CA"
         BANNER = <<-BANNER
 Usage:
   puppetserver ca create [--help]
-  puppetserver ca create [--config PATH] [--certname CERTNAME[,ADDLCERTNAME]]
+  puppetserver ca create [--puppet-config PATH] [--server-config PATH]
+                         [--certname CERTNAME[,ADDLCERTNAME]]
```
amysahli: I know this is a WIP.  But we'll need to update the run function below to support these changes.

lib/puppetserver/ca/config/combined.rb
```diff
@@ -0,0 +1,46 @@
...
+                       settings_overrides: {})
+          @errors = []
+          puppet_settings = Puppet.new(puppet_config_path)
+          puppet_overrides, server_overrides = split_overrides(settings_overrides, puppet_settings)
+
+          puppet_settings.load(puppet_overrides)
+          @errors << puppet_settings.errors
+          server_settings = PuppetServer.parse(puppet_settings.settings, server_config_path)
+          @errors << server_settings.errors
+          @errors.flatten!
```
amysahli: Remembering our conversation yesterday.  What's better, @errors += or ``` << then flatten! ``` ?

lib/puppetserver/ca/action/create.rb
Magisus: Oh yeah, guess I missed this one.

lib/puppetserver/ca/config/combined.rb
```diff
@@ -0,0 +1,46 @@
...
+
+          # Override things in puppet with things in puppetserver
+          @settings = puppet_settings.settings.merge(server_settings.settings)
+          # Override the result with remaining CLI overrides
+          @settings.merge!(server_overrides)
```
amysahli: I think it would make my brain happier if we merged these server_overrides into the server_settings like we do for puppet_settings.

Magisus:
If we're going to support Puppetserver settings falling back to Puppet settings, this needs to be updated.

lib/puppetserver/ca/config/combined.rb
justinstoller: `+=`

PR #28, opened 2018-08-30
(SERVER-2287) Add custom auth extension to master cert

Magisus:
This commit adds a new custom extension to the master cert when
generating it. This extension is checked by tk-auth when allowing access
to the certificate_status and certificate_statuses endpoints. This is to
help ensure that only requests from the puppet master's cert can access
those endpoints.

### DIFF
--- a/lib/puppetserver/ca/local_certificate_authority.rb
+++ b/lib/puppetserver/ca/local_certificate_authority.rb
@@ -16,6 +16,8 @@ module Puppetserver
       SSL_SERVER_CERT = "1.3.6.1.5.5.7.3.1"
       SSL_CLIENT_CERT = "1.3.6.1.5.5.7.3.2"
 
+      CLI_AUTH_EXT_OID = "1.3.6.1.4.1.34380.1.3.39"
+
       MASTER_EXTENSIONS = [
         ["basicConstraints", "CA:FALSE", true],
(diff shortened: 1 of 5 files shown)


Magisus:
This auth extension is checked by puppetserver in this PR: https://github.com/puppetlabs/puppetserver/pull/1781

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -82,6 +84,11 @@ def sign_master_cert(int_key, int_cert, csr)
           extension = ef.create_extension(*ext)
           cert.add_extension(extension)
         end
+
+        # Status API access for the CA CLI
+        cli_auth_ext = OpenSSL::X509::Extension.new(CLI_AUTH_EXT_OID, OpenSSL::ASN1::UTF8String.new("true").to_der, false)
```
amysahli: could just do this I think?
```
ef.create_extension(CLI_AUTH_EXT_OID, "true", false)
```

amysahli: Not that its really any better, but its what we do above and below

Magisus: I had run into issues with the factory, but that was probably because I was trying to use an invalid short name. I can try that with the full OID.

PR #30, opened 2018-08-30
(SERVER-2292) Make cert generation respect custom CSR attributes

amysahli:


lib/puppetserver/ca/host.rb
```diff
@@ -1,36 +1,153 @@
...
       end
 
       def create_private_key(keylength)
         OpenSSL::PKey::RSA.new(keylength)
       end
 
-      def create_csr(name, key, extensions = [])
+      def create_extension(extension_name, extension_value, critical = false)
+        value = OpenSSL::ASN1::UTF8String.new(extension_value.to_s)
+        # OpenSSL::X509::ExtensionFactory.new.create_extension(extension_name, value.to_der, critical)
```
Magisus: Should this line go away?

lib/puppetserver/ca/host.rb
```diff
@@ -1,36 +1,153 @@
...
+      ]
+
+      PRIVATE_EXTENSIONS = [
+        'subjectAltName', '2.5.29.17',
+      ]
+
+      # A mapping of Puppet extension short names to their OIDs. These appear in
+      # csr_attributes.yaml.
+      PUPPET_SHORT_NAMES =
+      {:pp_uuid =>             "1.3.6.1.4.1.34380.1.1.1",
```
Magisus: I'm not sure what we're doing about this in Puppet, but in my testing it looks like the keys parsed from the YAML is coming in as strings. Maybe we need to update these keys to be strings.

amysahli: Yep, that was a bug.  Just fixed it and added a spec.  Thank you!

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -130,7 +131,6 @@ def resolve_settings(overrides = {})
             :hostcert => '$certdir/$certname.pem',
             :hostprivkey => '$privatekeydir/$certname.pem',
             :hostpubkey => '$publickeydir/$certname.pem',
-            :publickeydir => '$ssldir/public_keys',
```
amysahli: This isn't related to my PR.  Just noticed it was a duplicate (it's in base-defaults above)

Magisus: Good catch.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -39,6 +39,10 @@ def initialize(digest, settings)
         @settings = settings
       end
 
+      def errors
+        @host.errors
```
Magisus: This seems good. It'd be easy to add any additional errors to the ones from host if this class ever needed to return any of its own.

lib/puppetserver/ca/action/create.rb
```diff
@@ -152,9 +153,14 @@ def generate_key_csr(certname, settings, digest)
           private_key = host.create_private_key(settings[:keylength])
           extensions = []
           if !settings[:subject_alt_names].empty?
-            extensions << host.create_extension("subjectAltName", settings[:subject_alt_names])
+            s_value = OpenSSL::ASN1::UTF8String.new(settings[:subject_alt_names].to_s)
+            extensions << OpenSSL::X509::Extension.new("subjectAltName", s_value.to_der, false)
```
Magisus: So turns out for whatever reason you don't want to process subject alt names this way. The server throws an error down in the Java signing code if you convert this to a UTF8 string DER. Putting this back to what you had before (I didn't add the function back, just put its contents here) makes it work again.

PR #32, opened 2018-09-08
(maint) Correctly encode SAN extReqs in CSRs

justinstoller:
Previously we passed a bare string into Extension.new, which would not
do the special pre-processing that an ExtensionFactory will do for us to
ensure that subjectAltNames were properly encoded. Because what was
encoded was a string, the printing and manual inspection lead us to
believe that everything was encoded properly, however the
ExtensionFactory will create a sequence with elements typed by their
kind of name (eg DNS or IP). More rigorous CA code was not allowing the
encoded string in the place of what should have been an encoded
sequence.

This moves us to use the ExtensionFactory so we do not need to care
about the convoluted nature of CSR encoding. It also adds a test to
prove that what is encoded in our CSRs are sequences that can be decoded
into their correct individual elements.

### DIFF
--- a/spec/puppetserver/ca/action/create_spec.rb
+++ b/spec/puppetserver/ca/action/create_spec.rb
@@ -242,6 +242,21 @@ RSpec.describe Puppetserver::Ca::Action::Create do
         _, csr = subject.generate_key_csr('foo', settings, OpenSSL::Digest::SHA256.new)
         expect(csr.attributes.count).to eq(1)
       end
+
+      it 'correctly encodes subject alt names' do
+        settings = { :subject_alt_names => 'DNS:foo, DNS:puppet',
+                     :keylength => 512,
+                     :csr_attributes => '$confdir/csr_attributes.yaml'}
+        _, csr = subject.generate_key_csr('foo', settings, OpenSSL::Digest::SHA256.new)
+
+        # If the subject alt names are correctly encoded then we should be able
+        # to decode just their context dependent values (ie just the names,
+        # not their type labels)
+        alt_request = csr.attributes[0].value.value[0].value[0].value[1].value
+        alt_names = OpenSSL::ASN1.decode(alt_request)
+        alt_names = alt_names.value.map {|name| name.value }
+        expect(alt_names).to include('foo', 'puppet')
+      end
     end
   end
 end
--- a/lib/puppetserver/ca/action/create.rb
+++ b/lib/puppetserver/ca/action/create.rb
@@ -153,7 +153,10 @@ BANNER
           private_key = host.create_private_key(settings[:keylength])
           extensions = []
           if !settings[:subject_alt_names].empty?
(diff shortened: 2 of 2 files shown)


justinstoller:
I'm working on a hacked up pre-suite to run this code through the server's acceptance tests with the additional changes in https://github.com/puppetlabs/puppetserver/pull/1783 .

justinstoller:
Ironically, I found this little tidbit on our docs page super helpful for visually checking whether or not we were encoding field correctly: https://puppet.com/docs/puppet/5.1/ssl_attributes_extensions.html#manually-checking-for-extensions-in-csrs-and-certificates
(of course I didn't find it until I had dropped into how different fields were supposed to be DER encoded)

> Note that every extension is preceded by any combination of two characters (.$ and .. in the above example) that contain ASN.1 encoding information. Since OpenSSL is unaware of Puppet’s custom extensions OIDs, it’s unable to properly display the values.

With that info the difference between these two CSRs become much more obvious:
```
puppetserver-ca-cli justin (maint-derp-der $%) ‖ openssl req -noout -text -in ~/.puppetlabs/etc/puppet/ssl/ca/requests/flarc.pem
...snip...
        Attributes:
        Requested Extensions:
            X509v3 Subject Alternative Name:
                ..DNS:localhost, DNS:fooberry

puppetserver-ca-cli justin (maint-derp-der $%) ‖ openssl req -noout -text -in ~/.puppetlabs/etc/puppet/ssl/ca/requests/flard.pem
...snip...
        Attributes:
        Requested Extensions:
            X509v3 Subject Alternative Name:
                DNS:localhost, DNS:fooberry
```

justinstoller:
Validated this works in puppetlabs/puppetserver#1783 
```
      Test Suite: tests @ 2018-09-07 22:15:11 -0700
      - Host Configuration Summary -
              - Test Case Summary for suite 'tests' -
       Total Suite Time: 1603.23 seconds
      Average Test Time: 106.88 seconds
              Attempted: 15
                 Passed: 15
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 15
      - Specific Test Case Status -
Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```

spec/puppetserver/ca/action/create_spec.rb
```diff
@@ -242,6 +242,21 @@
...
+      it 'correctly encodes subject alt names' do
+        settings = { :subject_alt_names => 'DNS:foo, DNS:puppet',
+                     :keylength => 512,
+                     :csr_attributes => '$confdir/csr_attributes.yaml'}
+        _, csr = subject.generate_key_csr('foo', settings, OpenSSL::Digest::SHA256.new)
+
+        # If the subject alt names are correctly encoded then we should be able
+        # to decode just their context dependent values (ie just the names,
+        # not their type labels)
+        alt_request = csr.attributes[0].value.value[0].value[0].value[1].value
```
Magisus: whheeee xD

PR #33, opened 2018-09-11
(SERVER-2305) Generate files needed for infra_crl

justinstoller:
Previously we did not create the infra_inventory.txt, infra_serials, or
infra_crl.pem that is needed by the infra-crl feature of Puppet Sever.

This patch adds the creation of the first two files (as empty files) and
the last as a duplicate of the cacrl (they shouldn't diverge until nodes
have actually been revoked).

This has multiple caveats:
 - if someone in importing a "full" crl that will become the base of
   their "infra" crl
 - the locations of the files are hardcoded within specific file names
   to be rooted in $cadir (the current Puppet Server code allows them
   to changed, but we've discussed only allowing moving the cadir in
   Puppet Server).

---

This PR also includes meaningful maint commits that should be viewed as well (we discussed earlier and decided to put them all in this PR, happy to break them out if folks prefer). They change the exit code of the sign action when there are no pending requests (for more compatibility) and backfill tests for the permissions of files created during import/generate.


spec/puppetserver/ca/action/generate_spec.rb
```diff
@@ -39,21 +39,34 @@
...
-        expect(File.exist?(File.join(tmpdir, 'ssl', 'public_keys', 'foocert.pem'))).to be true
+
+        created_correctly = ->(*args) do
+          perms = args.pop
+          file = File.join(tmpdir, *args)
+          File.exists?(file) &&
+            File.stat(file).mode.to_s(8)[-3..-1] == perms
+        end
+
+        expect(created_correctly.('ca', 'ca_crt.pem', '644')).to be true
```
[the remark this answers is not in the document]
Magisus: So I've never seen this syntax outside of a book before. More standard I think would be to just create a helper method at the top of the test with `def` syntax. That being said, I don't hate keeping this local like this. It'll expand people's horizons when they read it :)

spec/puppetserver/ca/action/import_spec.rb
```diff
@@ -249,12 +249,21 @@
...
-        expect(File.exist?(File.join(tmpdir, 'ca', 'ca_crt.pem'))).to be true
-        expect(File.exist?(File.join(tmpdir, 'ssl', 'certs', 'foocert.pem'))).to be true
-        expect(File.exist?(File.join(tmpdir, 'ssl', 'private_keys', 'foocert.pem'))).to be true
-        expect(File.exist?(File.join(tmpdir, 'ssl', 'public_keys', 'foocert.pem'))).to be true
+        created_correctly = ->(*args) do
+          perms = args.pop
+          file = File.join(tmpdir, *args)
+          File.exists?(file) &&
+            File.stat(file).mode.to_s(8)[-3..-1] == perms
+        end
```
Magisus: However, because this is in two places, maybe a better option would be to define it as a spec helper in another file. We already have a directory for test utilities here https://github.com/puppetlabs/puppetserver-ca-cli/tree/master/spec/utils. Is there a reason you didn't go this route?

justinstoller: Yeah, that's a good point. I had originally written lambdas to wrap the tmpdir and directory, ie
```rb
cadir = ->(file) { File.join(tmpdir, 'ca', file) }
File.exists?(cadir['ca_crt.pem'])
```
And then kept the tmpdir wrapping when I turned it into more of a test assertion helper. I'll pull all these assertions into their own test helper.

Magisus: Ah yeah, that makes sense. Sounds good, probably worth it, especially if you can remove even more duplication than just the lambda.

justinstoller:
agh, updated but I need to reword the log messages

justinstoller:
better

PR #34, opened 2018-09-11
(maint) Generate do-see-do

justinstoller:


justinstoller:
I attempted to manage the git history with a commit of just `git mv` and updating the references to that file in external requires (no constant updates), then a commit that doesn't move any files but updates all the constants/references in docs, then a third commit that moves specs that should be named similar to the files they test.

However that doesn't seem to have preserved the git history like I had wanted, not sure if there's a better way to do this or if this is just as good as it gets with Github's blame.

justinstoller:
ping @Magisus when you get a minute

Magisus:
I can do a little poking, but off the top of my head I don't know of a way.

lib/puppetserver/ca/cli.rb
```diff
@@ -23,8 +23,8 @@ class Cli
 
       VALID_ACTIONS = {
         'clean'    => Action::Clean,
-        'create'   => Action::Create,
         'generate' => Action::Generate,
+        'setup'    => Action::Setup,
```
Magisus: This is no longer alphabetized.

Magisus:
I'm fine leaving the history as is.

PR #35, opened 2018-09-13
(maint) Return 24 when revoking or cleaning unsigned CSRs

justinstoller:
I don't know, seems like a weird behavior to keep.

### DIFF
--- a/lib/puppetserver/ca/certificate_authority.rb
+++ b/lib/puppetserver/ca/certificate_authority.rb
@@ -152,11 +152,24 @@ module Puppetserver
               cleaned = check_clean(certname, clean_result)
             end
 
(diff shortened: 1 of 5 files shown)


justinstoller:
ping @Magisus

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -18,31 +18,49 @@ def initialize(logger, settings)
...
       end
 
+      def worst_result(previous_result, current_result)
+        %i{success invalid not_found error}.each do |state|
+          if previous_result == state
+            return current_result
+          elsif current_result == state
+            return previous_result
+          else
+            next
```
Magisus: Oh boy, that's fun. Makes sense.

PR #36, opened 2018-09-18
(maint) Update readme with some usage details

Magisus:


amysahli:
Looks good Maggie.  Are we in flux regarding creating cerfificates "ca create?"

README.md
```diff
@@ -13,8 +13,42 @@ You may install it yourself with:
...
+puppetserver ca import --cert-bundle certs.pem --crl-chain crls.pem --private-key ca_key.pem
+```
+
+The remaining actions provided by this gem require a running Puppet Server, since
+it primarily uses the CA's API endpoints to do its work. The following examples
+assume that you are using the gem packaged within Puppet Server.
+
+To sign a pending certificate request:
+```
+puppetserver ca sign --certname mycsr.pem
```
justinstoller: This is actually wrong, it should `--certname foo.example.com` or something that doesn't look like a filepath.

README.md
```diff
@@ -13,8 +13,42 @@ You may install it yourself with:
...
+```
+
+To list certificates and CSRs:
+```
+puppetserver ca list --all
+```
+
+To revoke a signed certificate:
+```
+puppetserver ca revoke --certname mycert.pem
```
justinstoller: ditto

README.md
Magisus: Oh, derp.

Magisus:
Yes we've renamed "generate" -> "setup" and "create" -> "generate" to match `puppet cert`.

Magisus:
Do we want me to add the other commands as well?

justinstoller:
Good catch @amysahli , I totally glossed over the specifics when I first looked at this. I'd probably either include all the commands or call out that the ones you're using are showing usage conventions eg instead of saying `to sign a cert`, saying something like `all commands take a comma separated list of one or more certnames to act on` and then using `sign` as an example.

Magisus:
Updated to list all the commands.

README.md
```diff
@@ -13,8 +13,52 @@ You may install it yourself with:
...
+```
+puppetserver ca list --all
+```
+
+To revoke a signed certificate:
+```
+puppetserver ca revoke --certname foo.example.com
+```
+
+To clean up all SSL files for a given certname:
```
justinstoller: This revokes and cleans up the files

Magisus: Updated

PR #37, opened 2018-09-21
(SERVER-2267) Output all SANs including IP names

amysahli:


lib/puppetserver/ca/action/list.rb
```diff
@@ -97,7 +97,7 @@ def output_certs(certs)
 
           certs.each do |cert|
             @logger.inform "    #{cert["name"]}".ljust(padded + 6) + " (SHA256) " + " #{cert["fingerprints"]["SHA256"]}" +
-                               (cert["dns_alt_names"].empty? ? "" : "\talt names: #{cert["dns_alt_names"]}")
+                               (cert["subject_alt_names"].empty? ? "" : "\talt names: #{cert["subject_alt_names"]}")
```
justinstoller: We should see if subject_alt_names exists and if so use it, otherwise use dns_alt_names.

justinstoller: We can do that later

Magisus:
I verified this against the server changes in https://github.com/puppetlabs/puppetserver/pull/1810 and everything seems to be in order.

lib/puppetserver/ca/action/list.rb
```diff
@@ -96,8 +96,12 @@ def output_certs(certs)
           end
 
           certs.each do |cert|
+            # In newer versions of the CA api we return subjcet_alt_names
+            # in addition to dns_alt_names, this field includes DNS alt
+            # names but also IP alt names.
+            alt_names = cert["subject_alt_names"] || cert["dns_alt_names"]
             @logger.inform "    #{cert["name"]}".ljust(padded + 6) + " (SHA256) " + " #{cert["fingerprints"]["SHA256"]}" +
-                               (cert["dns_alt_names"].empty? ? "" : "\talt names: #{cert["dns_alt_names"]}")
+                               (alt_names.empty? ? "" : "\talt names: #{alt_names}")
```
justinstoller: @Magisus 	I updated this wrt my prior comment. Let me know what you think. Otherwise I'm +1

Magisus: Good, this seems more robust.

PR #38, opened 2018-09-24
(SERVER-2320) Allow generation of an authorized cert offline

Magisus:
The cert used by the CA CLI needs to have a special auth extension that
is disallowed by the Puppet Server CA, to prevent any node from
requesting a cert with the special permissions. Previously, this cert
was always created along with the rest of the CA files.

This commit adds a way to generate an authorized cert with the special
extension, from an existing CA. The special cert is signed offline, not
through the main Puppet CA. It can be generated with values for the
master, or with values for other nodes. In the latter case the cert and
keys will need to be distributed manually.

This also updates `generate` to not use the `subject_alt_names`
setting. This setting is supposed to be specific to the master node, but
`generate` can be used to create certs for any node, and those shouldn't
necessarily have the master's alt names. When using `generate`, the user
needs to specify both certname and any alt names manually.

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -1,9 +1,12 @@
 require 'puppetserver/ca/utils/cli_parsing'
 require 'puppetserver/ca/host'
 require 'puppetserver/ca/certificate_authority'
+require 'puppetserver/ca/local_certificate_authority'
+require 'puppetserver/ca/x509_loader'
 require 'puppetserver/ca/config/puppet'
 require 'puppetserver/ca/utils/file_system'
 require 'puppetserver/ca/utils/signing_digest'
+require 'puppetserver/ca/utils/config'
 
 module Puppetserver
   module Ca
@@ -20,12 +23,26 @@ module Puppetserver
         BANNER = <<-BANNER
 Usage:
   puppetserver ca generate [--help]
-  puppetserver ca generate [--config PATH] [--certname NAME[,NAME]]
+  puppetserver ca generate --certname NAME[,NAME] [--config PATH]
                            [--subject-alt-names NAME[,NAME]]
+                           [--ca-client]
 
 Description:
 Generates a new certificate signed by the intermediate CA
 and stores generated keys and certs on disk.
+
+If the `--ca-client` flag is passed, the cert will be generated
+offline, without using Puppet Server's signing code, and will add
+a special extension authorizing it to talk to the CA API. This can
+be used for regenerating the master's host cert, or for manually
+setting up other nodes to be CA clients. Do not distribute certs
+generated this way to any node that you do not intend to have
+administrative access to the CA (e.g. the ability to sign a cert).
+
+Since the `--ca-client` causes a cert to be generated offline, it
+should ONLY be used when Puppet Server is NOT running, to avoid
+conflicting with the actions of the CA service. This will be
+mandatory in a future release.
 
 To determine the target location, the default puppet.conf
 is consulted for custom values. If using a custom puppet.conf
@@ -55,6 +72,11 @@ BANNER
             opts.on('--subject-alt-names NAME[,NAME]',
                     'Subject alternative names for the generated cert') do |sans|
               parsed['subject-alt-names'] = sans
+            end
+            opts.on('--ca-client',
+                    'Whether this cert will be used to request CA actions.\
+                    Causes the cert to be generated offline.') do |ca_client|
+              parsed['ca-client'] = true
             end
           end
         end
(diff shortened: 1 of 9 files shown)


Magisus:
Some questions here:
1) Do we only want to allow creating CA client certs when the server isn't running?
2) Should it overwrite existing files?
3) Do we like this UX, or should this be its own command/a different flag?
4) Do we want to call out that we're not going through the CA API in logging?

Magisus:
Manually verified that this works. Created a CA using `pupeptserver ca setup`, deleted the master cert, ran `puppetserver ca generate ... --ca-client`, it overwrote all the other files that were still hanging around (signed dir, keys), then I could successfully talk to the server again. Didn't even have to HUP it.

Magisus:
If we like this approach I'll add a couple more tests to `generate`.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -103,27 +109,60 @@ def run(input)
...
-          settings_overrides[:dns_alt_names] = input['subject-alt-names'] unless input['subject-alt-names'].empty?
           puppet = Config::Puppet.new(config_path)
           puppet.load(settings_overrides)
           return 1 if CliParsing.handle_errors(@logger, puppet.errors)
 
+          # We don't want generate to respect the alt names setting, since it is usually
+          # used to generate certs for other nodes
+          alt_names = Puppetserver::Ca::Utils::Config.munge_alt_names(input['subject-alt-names'])
+          if !alt_names.empty?
+            alt_names += ",DNS:#{puppet.settings[:certname]}"
```
justinstoller: If someone gave us an explicit list of alt names, then also add the certname as an alt name? I'm not grokking the purpose here.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -73,33 +82,68 @@ def create_master_cert(ca_key, ca_cert)
...
 
         return master_key, master_cert
       end
 
-      def sign_master_cert(int_key, int_cert, csr)
+      # Create a cert that is authorized to talk to the CA API. This may
+      # be used to regenerate the master cert, or to create certs for other nodes
+      # that need CA access, like the PE console.
+      def create_authorized_cert(ca_key, ca_cert, csr, alt_names = '')
+        cert = nil
```
justinstoller: old code?

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -73,33 +82,68 @@ def create_master_cert(ca_key, ca_cert)
                                               @settings[:hostpubkey])
         if master_key
           master_csr = @host.create_csr(name: @settings[:certname], key: master_key)
-          master_cert = sign_master_cert(ca_key, ca_cert, master_csr)
+          alt_names =
+            if @settings[:subject_alt_names].empty?
```
justinstoller: I thought we decided not to do assignment of from non-one-line-ternary-conditionals somewhere else?

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -73,33 +82,68 @@ def create_master_cert(ca_key, ca_cert)
...
+        cert = nil
+        cert = sign_authorized_cert(ca_key, ca_cert, csr, alt_names)
+
+        return cert
+      end
+
+      # Used when generating certificates offline.
+      def load_ca
+        if File.exist?(@settings[:cacert]) && File.exist?(@settings[:cakey]) && File.exist?(@settings[:cacrl])
+          ca = Puppetserver::Ca::X509Loader.new(@settings[:cacert], @settings[:cakey], @settings[:cacrl])
```
justinstoller: I don't' think I'd name this `ca`. We're the CA, it's our PKI loader.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -103,27 +109,60 @@ def run(input)
...
+
+          ca = Puppetserver::Ca::LocalCertificateAuthority.new(digest, settings)
+          ca_cert, ca_key = ca.load_ca
+
+          passed = certnames.map do |certname|
+            # For certs signed offline, any alt names are added directly to the cert,
+            # rather than to the CSR.
+            key, csr = generate_key_csr(certname, settings, digest)
+            return false unless csr
+            cert = ca.create_authorized_cert(ca_key, ca_cert, csr, alt_names)
```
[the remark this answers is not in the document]
Magisus: Yeah this really threw me for a loop, realizing this class isn't set up that way at all. It was too large of a refactor while I was still trying to figure out what else needed to happen, but maybe when I've address the other comments here, I'll give this another shot (can always file a separate ticket for it if we want to unblock this PR).

lib/puppetserver/ca/local_certificate_authority.rb
Magisus: Yeah I personally prefer variable assignment in both branches, this was just ripped directly from another part of the code. I'll update.

lib/puppetserver/ca/action/generate.rb
Magisus: Apparently if SANs are present, the CN is ignored, see https://stackoverflow.com/questions/5935369/ssl-how-do-common-names-cn-and-subject-alternative-names-san-work-together. My testing bore this out.

I'll switch the order here so that the certname is added before the munge, so it is deduplicated.

Magisus:
Updated with extra/better tests.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -103,27 +120,68 @@ def run(input)
...
+        def generate_authorized_certs(certnames, alt_names, settings, digest)
+          # Make sure we have all the directories where we will be writing files
+          FileSystem.ensure_dirs([settings[:ssldir],
+                                  settings[:certdir],
+                                  settings[:privatekeydir],
+                                  settings[:publickeydir]])
+
+          ca = Puppetserver::Ca::LocalCertificateAuthority.new(digest, settings)
+          ca_cert, ca_key = ca.load_ca
+          return 1 if CliParsing.handle_errors(@logger, ca.errors)
```
justinstoller: I think here we want to return `false`?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -103,27 +120,68 @@ def run(input)
...
+              # altnames are present, so we need to ensure that the certname is
+              # also listed among the alt names.
+              alt_names += ",DNS:#{certname}"
+              alt_names = Puppetserver::Ca::Utils::Config.munge_alt_names(alt_names)
+            end
+
+            # For certs signed offline, any alt names are added directly to the cert,
+            # rather than to the CSR.
+            key, csr = generate_key_csr(certname, settings, digest)
+            return false unless csr
```
justinstoller: I think this will break out of the certnames loop and directly return false from `generate_authorized_certs`?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -133,7 +191,15 @@ def generate_certs(certnames, settings, digest)
           ca = Puppetserver::Ca::CertificateAuthority.new(@logger, settings)
 
           passed = certnames.map do |certname|
-            key, csr = generate_key_csr(certname, settings, digest)
+            if !alt_names.empty?
+              # When validating the cert, OpenSSL will ignore the CN field if
+              # altnames are present, so we need to ensure that the certname is
+              # also listed among the alt names.
+              alt_names += ",DNS:#{certname}"
```
justinstoller: I would assume that each cert in the certnames list would have the `DNS:#{certname}` value of the preceding certs? It'd be nice to set something like
```rb
this_certs_alt_names = alt_names.dup
if !this_certs_alt_names.empty?
  this_certs_alt_names += ",DNS:#{certname}"
```

Magisus: Oh, derp

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -73,33 +82,60 @@ def create_master_cert(ca_key, ca_cert)
...
         end
 
         return master_key, master_cert
       end
 
-      def sign_master_cert(int_key, int_cert, csr)
+      # Used when generating certificates offline.
+      def load_ca
+        if File.exist?(@settings[:cacert]) && File.exist?(@settings[:cakey]) && File.exist?(@settings[:cacrl])
+          signing_cert, key = nil
```
justinstoller: This kinda works, but not explicitly, iirc. I think that any additional variables in a multi-assignment line get `nil`, so like
```
a, b, c = 1, 2
```
will leave `a == 1`, `b == 2`, and `c == nil`. iow, you're explicitly setting `signing_cert` to `nil` and _implicitly_ setting `key` to `nil` because the rhs lacks enough objects to match the lhs.

justinstoller: Also I think we should lift this to above the first `if`, incase the files don't exist we'll try to return uninitialized variables.

justinstoller:
Sorry, missed those comments:
>Some questions here:
>
> 1. Do we only want to allow creating CA client certs when the server isn't running?

Yeah, I think we want to be able to assume that when the CA is running it is canonical and doesn't need to coordinate with other processes (for updating the serial number for instance).

> 2. Should it overwrite existing files?

I'd rather not, we don't do that elsewhere so it seems good to not start now. This matches the the-master-must-be-offline-when-regenerating-its-cert that is assumed above. Don't know how @jpartlow feels about that though?

> 3. Do we like this UX, or should this be its own command/a different flag?

The help around the flag assuages my concerns about it only being a flag.

> 4. Do we want to call out that we're not going through the CA API in logging?

I don't know? Nothing jumped out at me wrt logging.

lib/puppetserver/ca/action/generate.rb
Magisus: I think since this is in a `map` call, it just ends up populating `passed` with trues and falses. But I'll try it out.

justinstoller:
I think this close, most of my comments are wrt issues that might arise with multiple certnames passed in, I don't know how valuable that functionality is, but it might be good to have tests around it.

justinstoller:
>> 2. Should it overwrite existing files?

> I'd rather not, we don't do that elsewhere so it seems good to not start now. This matches the the-master-must-be-offline-when-regenerating-its-cert that is assumed above. Don't know how jpartlow feels about that though?

So, I'm not sure how we got the current implementation of `save_file` in the generate action as it goes against the behavior I thought we had elsewhere (probably my own capriciousness). I could see not changing it since it's the existing behavior :shrug:

jpartlow:
I think the question is should the `puppetserver ca generate ... --ca-client` overwrite an existing set of cert/keys?

I'm not sure? What's the behavior of puppetserver ca setup? Probably we should match that?

Magisus:
Yeah I've updated this so that it no longer overwrites certs and keys.

Magisus:
Updated the docs to emphasize that the flag should not be used while the server is running. I'm going to file a ticket for making that mandatory, so we can figure out how to do it right, and maybe figure out a way to error more gracefully in general when the server isn't running.

Manually tested this with single and multiple certnames, things look good.

PR #41, opened 2018-10-08
(SERVER-2345) Support autosigning when generating certs

Magisus:
Previously if the user tried to generate a cert that was autosigned by
the CA, the command would error, because it would request that the cert
be signed after it was already signed by autosign. This commit updates
the generate logic to take into account autosigning.

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -199,28 +199,36 @@ BANNER
 
             current_alt_names = process_alt_names(alt_names, certname)
 
-            key, csr = generate_key_csr(certname, settings, digest, current_alt_names)
-            next false unless csr
-            next false unless ca.submit_certificate_request(certname, csr)
+            next false unless submit_csr(certname, ca, settings, digest, current_alt_names)
 
             # Check if the CA autosigned the cert
-            if result = ca.get_certificate(certname)
+            if download_cert(ca, certname, settings)
               @logger.inform "Certificate for #{certname} was autosigned."
-              next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
-              next false unless save_keys(certname, settings, key)
               true
             else
               next false unless ca.sign_certs([certname])
-              if result = ca.get_certificate(certname)
-                next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
-                next false unless save_keys(certname, settings, key)
-                true
-              else
-                false
-              end
+              download_cert(ca, certname, settings)
             end
           end
           passed.all?
+        end
+
+        def submit_csr(certname, ca, settings, digest, alt_names)
+          key, csr = generate_key_csr(certname, settings, digest, alt_names)
+          return false unless csr
+          # Always save the keys, since soemtimes the server saves the CSR
+          # even when it returns a 400 (e.g. when the CSR contains alt names
+          # but the server isn't configured to sign such certs)
+          return false unless save_keys(certname, settings, key)
+          return false unless ca.submit_certificate_request(certname, csr)
+          true
+        end
+
+        def download_cert(ca, certname, settings)
+          if result = ca.get_certificate(certname)
+            return false unless save_file(result.body, certname, settings[:certdir], "Certificate")
+            true
+          end
         end
 
         # For certs signed offline, any alt names are added directly to the cert,
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -202,13 +202,22 @@ BANNER
             key, csr = generate_key_csr(certname, settings, digest, current_alt_names)
             next false unless csr
             next false unless ca.submit_certificate_request(certname, csr)
-            next false unless ca.sign_certs([certname])
+
+            # Check if the CA autosigned the cert
             if result = ca.get_certificate(certname)
+              @logger.inform "Certificate for #{certname} was autosigned."
               next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
               next false unless save_keys(certname, settings, key)
               true
             else
-              false
+              next false unless ca.sign_certs([certname])
+              if result = ca.get_certificate(certname)
+                next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
+                next false unless save_keys(certname, settings, key)
+                true
+              else
+                false
+              end
             end
           end
           passed.all?
--- a/spec/puppetserver/ca/action/generate_spec.rb
+++ b/spec/puppetserver/ca/action/generate_spec.rb
@@ -99,7 +99,9 @@ RSpec.describe Puppetserver::Ca::Action::Generate do
     it 'logs an error if any could not be downloaded' do
       not_found = Utils::Http::Result.new('404', 'Not Found')
       allow(connection).to receive(:put).and_return(success)
-      allow(connection).to receive(:get).and_return(not_found, success_with_content)
+      # Look for each cert twice, before and after signing
+      allow(connection).to receive(:get).and_return(not_found, not_found,
+                                                    not_found, success_with_content)
       Dir.mktmpdir do |tmpdir|
         with_temp_dirs tmpdir do |config|
           code = subject.run({'certnames' => ['foo', 'bar'],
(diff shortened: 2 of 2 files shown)


Magisus:
Also tested this out manually with a local puppetserver configured with `autosign = true`.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -202,13 +202,22 @@ def generate_certs(certnames, alt_names, settings, digest)
...
+              @logger.inform "Certificate for #{certname} was autosigned."
               next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
               next false unless save_keys(certname, settings, key)
               true
             else
-              false
+              next false unless ca.sign_certs([certname])
+              if result = ca.get_certificate(certname)
+                next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
+                next false unless save_keys(certname, settings, key)
```
justinstoller: Did we want to save keys after a successful cert request?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -202,13 +202,22 @@ def generate_certs(certnames, alt_names, settings, digest)
...
             if result = ca.get_certificate(certname)
+              @logger.inform "Certificate for #{certname} was autosigned."
               next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
               next false unless save_keys(certname, settings, key)
               true
             else
-              false
+              next false unless ca.sign_certs([certname])
+              if result = ca.get_certificate(certname)
+                next false unless save_file(result.body, certname, settings[:certdir], "Certificate")
```
justinstoller: If signing fails should we save the csr?

lib/puppetserver/ca/action/generate.rb
```diff
@@ -202,13 +202,22 @@ def generate_certs(certnames, alt_names, settings, digest)
             key, csr = generate_key_csr(certname, settings, digest, current_alt_names)
             next false unless csr
             next false unless ca.submit_certificate_request(certname, csr)
-            next false unless ca.sign_certs([certname])
+
+            # Check if the CA autosigned the cert
```
justinstoller: These bits inside the `.map` feel like they should be a separate method or three? I'm not sure what the best way to break them up would be though.

lib/puppetserver/ca/action/generate.rb
Magisus: I think that's what this does... are you saying we should save them earlier? (e.g. after CSR submission)

lib/puppetserver/ca/action/generate.rb
Magisus: If we do, does this need to also try to load it before generating it? And if we save the CSR we probably also need to save the keys that go with it.

lib/puppetserver/ca/action/generate.rb
justinstoller: yeah, sorry, do we want to save the keys after a successful cert request, regardless of the signing outcome?

lib/puppetserver/ca/action/generate.rb
justinstoller: I'm into saving the keys regardless. I'm indifferent to loading them. I could see reading bits from disk trying to pick up a failed generation at the right spot being a little magical and error prone.

The use case that I run into is that I do a `puppetserver ca generate --certname foo --subject-alt-name bar` and it submits a CSR for said certname but fails with a "update your server conf and restart your server to allow signing SANS". I do that, but then I have a CSR waiting on the server (that I can now sign) but I don't have the keys for it even if I do.

I'd be happy if there's an error the cli just said, "Help, here are the artifacts I created, you're going to have to resolve this on your own".

Magisus:
Updated.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -199,21 +199,35 @@ def generate_certs(certnames, alt_names, settings, digest)
...
             end
           end
           passed.all?
         end
 
+        def submit_csr(certname, ca, settings, digest, alt_names)
+          key, csr = generate_key_csr(certname, settings, digest, alt_names)
+          return false unless csr
+          return false unless ca.submit_certificate_request(certname, csr)
+          return false unless save_keys(certname, settings, key)
```
justinstoller: Ahhh, I tried setting autosign to true and allow-auth-extensions to false in my puppetserver.conf and then doing a generate with auth extensions and this still fails but it looks like it fails because an assumption of mine was wrong:
```
puppetserver-ca-cli justin (mergemaggie $) ‖ be exe/puppetserver-ca generate --certname foo
Error:
    When attempting to submit certificate request for 'foo', received:
      code: 400
      body: CSR 'foo' contains an authorization extension, which is disallowed. To allow authorization extensions, set allow-authorization-extensions to true in your
puppetserver.conf file, restart the puppetserver, and try signing this certificate again.
```
That's the csr submission failing rather than the sign failing.
Looking at the CA
```
puppetserver-ca-cli justin (mergemaggie $) ‖ tree ~/.puppetlabs/etc/puppet/ssl/
/Users/justin/.puppetlabs/etc/puppet/ssl/
├── ca
│   ├── ca_crl.pem
│   ├── ca_crt.pem
│   ├── ca_key.pem
│   ├── ca_pub.pem
│   ├── infra_crl.pem
│   ├── infra_inventory.txt
│   ├── infra_serials
│   ├── inventory.txt
│   ├── requests
│   │   └── foo.pem
│   ├── root_key.pem
│   ├── serial
│   └── signed
│       └── localhost.pem
├── certificate_requests
├── certs
│   ├── ca.pem
│   └── localhost.pem
├── crl.pem
├── private
├── private_keys
│   └── localhost.pem
└── public_keys
    └── localhost.pem
```

Unfortunately, the master has still saved the csr. I'm not sure if we should save the keys even if the submission fails, or if we should try to figure out if the request failed in a way that would leave the submission on the CA? Or if we should go back to requiring the cleaning of CSRs after any error in the process.

Magisus: I see two options here:
1) Does it make sense for puppetserver to save the CSR in this case? Even though it rejected it as a "bad request"? Maybe it shouldn't save it.
2) If we know what codes result in the CSR being saved, treat all of those as a "success" here. We wanted to submit a CSR, and we did.
I'm leaning towards (2), or else (1) while also updating puppetserver's error messaging to point out that the CSR will need to be resubmitted. I don't really want to go back to requiring cleaning of CSRs, because I think it would unclear to me as a user whether I'd also need to clean keys or do other things. If a CSR exists, it seems like a reasonable assumption that the keys would exist too...

justinstoller: It looks like we return either a 2xx, 400, or 500 (from [the handler](https://github.com/puppetlabs/puppetserver/blob/master/src/clj/puppetlabs/services/ca/certificate_authority_core.clj#L57)). Of the reasons we return 400 we'll have duplicate certs and invalid certnames will cause the cert to not be persisted (makes sense since we save it to a file based on the certname), other 400 codes represent problems with autosigning and will leave the csr on disk.

joshcooper: I've been thinking that agent-server comms should return 200 if it's a new CSR or the same CSR as what the agent already sent. Otherwise return 422 if the agent sends a different CSR with the same name, or 400 if the agent sends a bad request/illegal extension. The 422 means the server understands the request and it's valid, but the server can't process the request and the client shouldn't attempt to resend it.

Magisus: But that doesn't really help here, right? We still have some 400s that save the CSR, and others that don't. So I'm not sure we can be smart about this...?

Magisus: Ah okay. So would 2xx extend to cases with altnames/extensions, since ultimately the CSR submission is fine if the server chooses to ultimately change its settings and sign it?

joshcooper: That makes sense to me, only return success so the agent knows it can transition from the "I have a local CSR" state to the "CSR accepted, awaiting cert" state.

Magisus: I don't think we can change this behavior now though. These are more good things to keep in mind for a v2 of the CA API.

Magisus: So in the meantime I think I'm just going to change it to save the keys in all cases. If the CSR failed to submit, hopefully the server will present a useful error message saying so. If the user then tries to run `generate` again, we can tell them to use `puppet ssl clean` to clean up the old keys.

Magisus:
So the problem with this now is that you can go in and sign the CSR after receiving a 400, but then the cert will never get placed in the cert dir. It still appears in the signeddir, so not sure how much that matters. But it is different from how generate usually works. Still might be better than forcing a clean and regenerate.

PR #42, opened 2018-10-17
(SERVER-2351) Check CA service before generating certs offline

Magisus:
In order to avoid race conditions with serial numbers, we want to
avoid allowing multiple processes to issue certificates at the same
time. This commit adds a check to make sure that Puppet Server's CA
service is offline before issuing certs out-of-band using Ruby code,
when using `generate --ca-client`. It queries the simple status endpoint
and errors if it receives a "running" response from the server.

### DIFF
--- a/lib/puppetserver/ca/certificate_authority.rb
+++ b/lib/puppetserver/ca/certificate_authority.rb
@@ -41,7 +41,7 @@ module Puppetserver
                       body: SIGN_BODY,
                       type: :sign)
 
-        results.all? {|result| result == :success }
+        results.all? { |result| result == :success }
       end
 
       def revoke_certs(certnames)
@@ -50,7 +50,7 @@ module Puppetserver
                     body: REVOKE_BODY,
                     type: :revoke)
 
-        results.reduce {|prev, curr| worst_result(prev, curr) }
+        results.reduce { |prev, curr| worst_result(prev, curr) }
       end
 
       def submit_certificate_request(certname, csr)
@@ -60,7 +60,7 @@ module Puppetserver
                     headers: {'Content-Type' => 'text/plain'},
                     type: :submit)
 
-        results.all? {|result| result == :success }
+        results.all? { |result| result == :success }
       end
 
       # Make an HTTP PUT request to CA
--- a/spec/puppetserver/ca/action/generate_spec.rb
+++ b/spec/puppetserver/ca/action/generate_spec.rb
@@ -305,10 +305,36 @@ RSpec.describe Puppetserver::Ca::Action::Generate do
   end
 
   describe "ca-client flag" do
+    let(:result) { Utils::Http::Result.new('200', 'running') }
+    let(:connection) { double }
+
+    before do
+      allow_any_instance_of(Puppetserver::Ca::Utils::HttpClient).
+        to receive(:with_connection).and_yield(connection)
+      allow_any_instance_of(Puppetserver::Ca::Utils::HttpClient).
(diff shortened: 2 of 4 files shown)


lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -50,7 +50,7 @@ def revoke_certs(certnames)
                     body: REVOKE_BODY,
                     type: :revoke)
 
-        results.reduce {|prev, curr| worst_result(prev, curr) }
+        results.reduce { |prev, curr| worst_result(prev, curr) }
```
justinstoller: What's with this?

spec/puppetserver/ca/action/generate_spec.rb
```diff
@@ -305,10 +305,37 @@
...
+    it "refuses to generate a cert if the server is running" do
+      allow(connection).to receive(:get).and_return(result)
+      Dir.mktmpdir do |tmpdir|
+        with_temp_dirs tmpdir do |config|
+          allow(subject).to receive(:generate_authorized_certs) { true }
+          code = subject.run({'certnames' => ['foo'],
+                              'config' => config,
+                              'subject-alt-names' => '',
+                              'ca-client' => true})
+          puts stderr.string
```
justinstoller: some discovery?

lib/puppetserver/ca/certificate_authority.rb
Magisus: Asymmetrical spacing. Style guides I found ([ex](https://github.com/github/rubocop-github/blob/master/STYLEGUIDE.md)) all say spaces after curly braces, which I think is generally what we do elsewhere.

spec/puppetserver/ca/action/generate_spec.rb
Magisus: Oh yeah, whoops.

lib/puppetserver/ca/certificate_authority.rb
justinstoller: Huh, I don't think I've noticed that lately.

justinstoller:
Minus the errant `puts` I think this is good to go. Functionally tested it and it works like it should (though testing there wasn't a regression makes we question some of our other messaging):
```
16:40:53 puppetserver-ca-cli justin (mergemaggie $) ‖ be exe/puppetserver-ca generate --certname blaz --ca-client
CA service is running. Please stop it before attempting to generate certs offline.

16:43:07 puppetserver-ca-cli justin (mergemaggie $) ‖ be exe/puppetserver-ca generate --certname blaz
Successfully saved private key for blaz to /Users/justin/.puppetlabs/etc/puppet/ssl/private_keys/blaz.pem
Successfully saved public key for blaz to /Users/justin/.puppetlabs/etc/puppet/ssl/public_keys/blaz.pem
Successfully submitted certificate request for blaz
Error:
    Signed certificate blaz could not be found on the CA
Successfully signed certificate request for blaz
Successfully saved certificate for blaz to /Users/justin/.puppetlabs/etc/puppet/ssl/certs/blaz.pem
```

Magisus:
Oh that looks like maybe a bug around the autosigning code... It seems like we're trying and failing to download the cert (for the autosign check), and printing out an error when we fail. We shouldn't do that.

Magisus:
We might want to consider refactoring the certificate authority to return its errors instead of logging them directly. Not sure yet.

Magisus:
Got rid of the `puts`.

PR #44, opened 2018-11-06
(SERVER-2335) Refactor LocalCertificateAuthority

tvpartytonight:
Before this change, no state about the local CA was kept in memory; this
change introduces instance variables for cert, key, and crl to the class
and allows for them to be read from an object if the necessary files
existed.

### DIFF
--- a/lib/puppetserver/ca/action/import.rb
+++ b/lib/puppetserver/ca/action/import.rb
@@ -73,7 +73,8 @@ BANNER
 
         def import(loader, settings, signing_digest)
           ca = Puppetserver::Ca::LocalCertificateAuthority.new(signing_digest, settings)
-          master_key, master_cert = ca.create_master_cert(loader.key, loader.certs.first)
+          ca.load_ssl_components(loader)
+          master_key, master_cert = ca.create_master_cert
           return ca.errors if ca.errors.any?
 
           FileSystem.ensure_dirs([settings[:ssldir],
--- a/lib/puppetserver/ca/action/setup.rb
+++ b/lib/puppetserver/ca/action/setup.rb
@@ -77,8 +77,8 @@ BANNER
           ca = Puppetserver::Ca::LocalCertificateAuthority.new(signing_digest, settings)
 
           root_key, root_cert, root_crl = ca.create_root_cert
-          int_key, int_cert, int_crl = ca.create_intermediate_cert(root_key, root_cert)
-          master_key, master_cert = ca.create_master_cert(int_key, int_cert)
+          ca.create_intermediate_cert(root_key, root_cert)
+          master_key, master_cert = ca.create_master_cert
           return ca.errors if ca.errors.any?
 
           FileSystem.ensure_dirs([settings[:ssldir],
@@ -89,14 +89,14 @@ BANNER
                                   settings[:signeddir]])
 
           public_files = [
-            [settings[:cacert], [int_cert, root_cert]],
-            [settings[:cacrl], [int_crl, root_crl]],
-            [settings[:cadir] + '/infra_crl.pem', [int_crl, root_crl]],
+            [settings[:cacert], [ca.cert, root_cert]],
+            [settings[:cacrl], [ca.crl, root_crl]],
+            [settings[:cadir] + '/infra_crl.pem', [ca.crl, root_crl]],
             [settings[:hostcert], master_cert],
-            [settings[:localcacert], [int_cert, root_cert]],
-            [settings[:hostcrl], [int_crl, root_crl]],
+            [settings[:localcacert], [ca.cert, root_cert]],
+            [settings[:hostcrl], [ca.crl, root_crl]],
             [settings[:hostpubkey], master_key.public_key],
-            [settings[:capub], int_key.public_key],
+            [settings[:capub], ca.key.public_key],
             [settings[:cert_inventory], ca.inventory_entry(master_cert)],
             [settings[:cadir] + '/infra_inventory.txt', ''],
             [settings[:cadir] + '/infra_serials', ''],
@@ -107,7 +107,7 @@ BANNER
           private_files = [
             [settings[:hostprivkey], master_key],
             [settings[:rootkey], root_key],
-            [settings[:cakey], int_key],
+            [settings[:cakey], ca.key],
           ]
 
           files_to_check = public_files + private_files
--- a/lib/puppetserver/ca/local_certificate_authority.rb
+++ b/lib/puppetserver/ca/local_certificate_authority.rb
@@ -1,5 +1,6 @@
 require 'puppetserver/ca/host'
 require 'puppetserver/ca/utils/file_system'
+require 'puppetserver/ca/x509_loader'
 
 require 'openssl'
 
--- a/lib/puppetserver/ca/local_certificate_authority.rb
+++ b/lib/puppetserver/ca/local_certificate_authority.rb
@@ -36,11 +36,37 @@ module Puppetserver
         ["authorityKeyIdentifier", "keyid:always", false]
       ].freeze
 
+      attr_reader :cert, :key, :crl
+
       def initialize(digest, settings)
         @digest = digest
         @host = Host.new(digest)
         @settings = settings
         @errors = []
+
+        if ssl_assets_exist?
+          @cert, @key, @crl = load_ssl_assets
+        end
+      end
+
+      def ssl_assets_exist?
+        File.exist?(@settings[:cacert]) &&
+          File.exist?(@settings[:cakey]) &&
+          File.exist?(@settings[:cacrl])
+      end
+
+      def load_ssl_assets(settings=@settings)
+        loader = Puppetserver::Ca::X509Loader.new(@settings[:cacert], @settings[:cakey], @settings[:cacrl])
+        if loader.errors.empty?
+          cert = loader.certs.first
+          key = loader.key
+          crl = loader.crls.first
+          return cert, key, crl
+        else
+          @errors += loader.errors
+          @errors << "CA not initialized. Please set up your CA before attempting to generate certs offline."
+          nil
+        end
       end
 
       def errors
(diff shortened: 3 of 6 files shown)


spec/puppetserver/ca/local_certificate_authority_spec.rb
```diff
@@ -1,23 +1,42 @@
...
+    it 'loads ssl assets if they exist' do
+      expect(subject.cert).to be_kind_of(OpenSSL::X509::Certificate)
+      expect(subject.key).to be_kind_of(OpenSSL::PKey::RSA)
+      expect(subject.crl).to be_kind_of(OpenSSL::X509::CRL)
+    end
+
+    context 'when an ssl asset is missing' do
+      let(:settings) {
+        with_ca_in(Dir.mktmpdir) do |config|
+          STDOUT.puts config
```
Magisus: Was this left in from debugging?

spec/puppetserver/ca/local_certificate_authority_spec.rb
```diff
@@ -1,23 +1,42 @@
...
-      :root_ca_name => 'bulla',
-      :certname => 'ulla',
-      :keylength => 512,
-      :hostprivkey => '$privatekeydir/$certname.pem',
-      :hostpubkey => '$publickeydir/$certname.pem',
-      :csr_attributes => '$confdir/csr_attributes.yaml',
-      :serial => '$cadir/serial' } }
+    with_ca_in(Dir.mktmpdir) do |config|
+      return Puppetserver::Ca::Config::Puppet.new(config).load({})
+    end
```
Magisus: Oh nice, this seems way better. I wonder if there are other places where we could do something like this, for example in https://github.com/puppetlabs/puppetserver-ca-cli/blob/master/spec/puppetserver/ca/utils/http_client_spec.rb...

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -36,11 +36,30 @@ class LocalCertificateAuthority
...
+        File.exist?(@settings[:cacert]) &&
+          File.exist?(@settings[:cakey]) &&
+          File.exist?(@settings[:cacrl])
+      end
+
+      def load_ssl_assets(settings=@settings)
+        cert = OpenSSL::X509::Certificate.new(File.read(settings[:cacert]))
+        key = OpenSSL::PKey::RSA.new(File.read(settings[:cakey]))
+        crl = OpenSSL::X509::CRL.new(File.read(settings[:cacrl]))
+        return cert, key, crl
```
Magisus: It might be good to continue to use the X509Loader here, since it has a lot of nice validation and error messages. I still think it's valuable to tell users when their CA is not set up correctly, especially when all the files exist, because that probably indicates that something serious went wrong.

spec/puppetserver/ca/local_certificate_authority_spec.rb
tvpartytonight: I've updated the PR to have a commit that refactors that http_client_spec.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -36,11 +37,37 @@ class LocalCertificateAuthority
...
+          File.exist?(@settings[:cakey]) &&
+          File.exist?(@settings[:cacrl])
+      end
+
+      def load_ssl_assets(settings=@settings)
+        loader = Puppetserver::Ca::X509Loader.new(@settings[:cacert], @settings[:cakey], @settings[:cacrl])
+        if loader.errors.empty?
+          cert = OpenSSL::X509::Certificate.new(File.read(settings[:cacert]))
+          key = OpenSSL::PKey::RSA.new(File.read(settings[:cakey]))
+          crl = OpenSSL::X509::CRL.new(File.read(settings[:cacrl]))
```
Magisus: The loader should already save these in instance variables when it's done loading them, can we just use those so we don't have to read the files twice? (Could get expensive if the CRL is large.)

spec/puppetserver/ca/utils/http_client_spec.rb
```diff
@@ -10,66 +10,31 @@
...
   include Utils::SSL
+  let(:tmpdir) {Dir.mktmpdir}
+  let(:settings) {
+    with_ca_in(tmpdir) do |config|
+      Puppetserver::Ca::Config::Puppet.new(config).load({
+        :hostcert => "#{tmpdir}/hostcert.pem",
+        :hostprivkey => "#{tmpdir}/hostkey.pem"
+      })
+    end
+  }
```
Magisus: Nice!

justinstoller:
In the `setup` action we create a CA object and then call what are effectively class methods on the instance (`create_root_cert`, `create_intermediate_cert`).

Then in both `setup` and `import` actions we call `create_master_cert` and pass in what I would consider the CA in question's key and cert. `create_master_cert` then calls `sign_authorized_cert` passing its cert and key (like we do in `generate`).

It seems like we could update those actions/`LocalCertificateAuthority#sign_authorized_cert` as well. Not sure if that's being thorough or bikeshedding though?

tvpartytonight:
@justinstoller that might be something in your local setup, since it passes in CI and on my laptop?

Magisus:
I agree, I think we should take advantage of the new stored state more widely, using the instance variables internally at least wherever possible, and refactoring the generation methods as a result. Otherwise we kind of just have two models side by side in the same class, which is never great.

spec/puppetserver/ca/local_certificate_authority_spec.rb
```diff
@@ -1,23 +1,50 @@
...
-      :ca_name => 'bulla2',
-      :subject_alt_names => '',
-      :root_ca_name => 'bulla',
-      :certname => 'ulla',
-      :keylength => 512,
-      :hostprivkey => '$privatekeydir/$certname.pem',
-      :hostpubkey => '$publickeydir/$certname.pem',
-      :csr_attributes => '$confdir/csr_attributes.yaml',
-      :serial => '$cadir/serial' } }
+    with_ca_in(Dir.mktmpdir) do |config|
```
justinstoller: Dir.mktmpdir without a block won't clean up the tmpdir it creates, it will though if we use a block, though I don't think returning the config will work quite the same way.

spec/puppetserver/ca/local_certificate_authority_spec.rb
```diff
@@ -1,23 +1,50 @@
...
+      return Puppetserver::Ca::Config::Puppet.new(config).load({})
```
justinstoller: The `with_ca_in` above yields a config file and the confdir. If we want to fully root this action in the tmpdir we need to override the confdir when we call load.

justinstoller:
@tvpartytonight  I do have a `csr_attributes.yaml` in my user confdir. I was surprised that these tests started reading it though, which is new to this PR. It looks like before we gave the CA a settings hash that had an unresolved `"@confdir/csr_attritues.yaml"` and when we did a `File.exist?` on it it naturally returned false.

With your improvement to the tests we actually have a meaningful confdir and so it can load csr_attributes file if one exists.

I think it's a bug though that that confdir is my actual user's confdir and not the one rooted in the tmpdir. Made a comment regarding that on `Puppet.new(config).load({})` call.

spec/puppetserver/ca/local_certificate_authority_spec.rb
tvpartytonight: I've changed the tests to delete the tmp dir in an `after` block.

spec/puppetserver/ca/local_certificate_authority_spec.rb
tvpartytonight: I've added the confdir to the `load` method as needed.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -76,7 +101,7 @@ def format_time(time)
         time.strftime('%Y-%m-%dT%H:%M:%S%Z')
       end
 
-      def create_master_cert(ca_key, ca_cert)
+      def create_master_cert(ca_key=@key, ca_cert=@cert)
```
Magisus: Do we need these args here? We don't seem to be actually passing anything to this ever, and even in the test, I think calling `subject.key` is the same as just letting it use `@key`.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -89,52 +114,32 @@ def create_master_cert(ca_key, ca_cert)
...
-          end
-        else
-          @errors << "CA not initialized. Please set up your CA before attempting to generate certs offline."
-        end
-
-        return signing_cert, key
-      end
-
-      def sign_authorized_cert(int_key, int_cert, csr, alt_names = '')
+      def sign_authorized_cert(csr, alt_names = '', ca_cert=@cert, ca_key=@key)
```
Magisus: Same question here. Do we need the two args that just default to fields?

lib/puppetserver/ca/local_certificate_authority.rb
Magisus: I guess I really just don't like the look of the fields as default args. I kind of think with this refactor, we are establishing a model where this CA class is an entity that should only be able to do things with its verified, saved internals. So I'm in favor of removing them.

lib/puppetserver/ca/action/setup.rb
```diff
@@ -77,8 +77,8 @@ def generate_pki(settings, signing_digest)
           ca = Puppetserver::Ca::LocalCertificateAuthority.new(signing_digest, settings)
 
           root_key, root_cert, root_crl = ca.create_root_cert
```
justinstoller: nit: this could be a class method

lib/puppetserver/ca/action/setup.rb
```diff
@@ -77,8 +77,8 @@ def generate_pki(settings, signing_digest)
...
-          int_key, int_cert, int_crl = ca.create_intermediate_cert(root_key, root_cert)
-          master_key, master_cert = ca.create_master_cert(int_key, int_cert)
+          ca.create_intermediate_cert(root_key, root_cert)
```
justinstoller: God, I want to write something really Java-y here, like this could be a custom constructor:
```rb
ca = LocalCertificateAuthority.asIntermediateFrom(root_key, root_cert)
```

And I think the gist of that is applicable, but I'm failing to think of an idiomatic way call for it. Anyways, another nitpicky idea.

lib/puppetserver/ca/action/import.rb
```diff
@@ -73,7 +73,8 @@ def run(input)
 
         def import(loader, settings, signing_digest)
           ca = Puppetserver::Ca::LocalCertificateAuthority.new(signing_digest, settings)
-          master_key, master_cert = ca.create_master_cert(loader.key, loader.certs.first)
+          ca.load_ssl_components(loader)
```
justinstoller: would it be weird to pass the key and cert into `.new()`?

tvpartytonight: Initially I wanted the caller to be able to specify more about how the `LocalCertificateAuthority` was created via named arguments, but I decided against adding more complexity to the constructor's arguments and opted for a way to set the key and cert via `load_ssl_components`. I'd consider that method to be the public accessor/setter for the class instance vars introduced in this PR.

lib/puppetserver/ca/action/setup.rb
tvpartytonight: I totally agree; I kinda thought that much of the root_* aspects of this class should be removed entirely and put into some kind of Util...

PR #45, opened 2018-12-06
(SERVER-2392) Respect server_list & (SERVER-2317) Handle connection errors gracefully

justinstoller:


justinstoller:
Need to put in the ticket, also maybe some better testing (was wondering what folks thought about that). I'd also really like to put in some error handling around our connections since issues with this (in particular) are terrible to debug currently.

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -239,6 +245,26 @@ def parse_crl_usage(setting)
             :ignore
           end
         end
+
+        def update_for_server_list!(settings)
```
Magisus: So if I'm reading this right, we're giving precedence to `server` here. In Puppet, `server_list` takes precedence over `server`. See https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L1456-L1459 and https://github.com/puppetlabs/puppet/blob/master/lib/puppet/configurer.rb#L217-L238. I think we want to preserve that behavior here.

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -46,13 +52,33 @@ def load_key(key_path)
...
-                          cert: @cert, key: @key,
-                          &request)
+          begin
+            Net::HTTP.start(url.host, url.port,
+                            use_ssl: true, cert_store: @store,
+                            cert: @cert, key: @key,
+                            &request)
+          rescue StandardError => e
+            raise ConnectionFailed.new(
+              "Failed connecting to #{url.full_url}\n" +
```
Magisus: yesss

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -66,8 +66,9 @@ def load(cli_overrides = {})
           results ||= {}
           results[:main] ||= {}
           results[:master] ||= {}
+          results[:agent] ||= {}
 
-          overrides = results[:main].merge(results[:master])
+          overrides = results[:agent].merge(results[:main]).merge(results[:master])
```
Magisus: I'm not sure if I like this or not. This CA client in my mind isn't really associated with the puppet agent. Is there a specific reason you want to do this?

lib/puppetserver/ca/config/puppet.rb
justinstoller: Oh, good catch.

lib/puppetserver/ca/config/puppet.rb
justinstoller: Yeah, I think ideally it would read just the [main] section (like apply and a lot of puppet subcommands).

I think we'll break a bunch of folks changing the default from $certname to puppet unless we also read from the agent section. I have yet to find a test/dev set up that set server in [main] (though many set it in [agent]) which makes sense, and in all those cases it just worked before because the server in the agent section on the master was set to $certname.

justinstoller: More concisely, I don't think this is correct, but I don't think it is any less correct than also merging in the master section, and I think it will meaningfully ease folks' upgrade.

Magisus: Got it. Fair enough.

Fun fact, most of Puppet's CLI tools actually will prefer the little known `user` section over `[main]`: https://puppet.com/docs/puppet/5.4/config_file_main.html#config-sections. Found this when I assumed they'd use `[agent]` and couldn't figure out why it wasn't working. Since I think almost no one uses it though, I don't think we have account for it here.

justinstoller: I think there is, or was, a ca section too?

Magisus: Haven't heard of that one.

Magisus:
As far as additional testing, the only thing that comes to mind is adding something about the agent fallback (possibly even just to the first test in the file that currently has `master` as well as `main`), mostly for documentation of the behavior.

lib/puppetserver/ca/errors.rb
```diff
@@ -0,0 +1,11 @@
+module Puppetserver
+  module Ca
+    class Error < StandardError; end
+    class FileNotFound < Error; end
+    class InvalidX509Object < Error; end
+    class ConnectionFailed < Error; end
+
+    module Errors
```
justinstoller: what were you doing here, Justin?

justinstoller: Oh, I think I was going to move [CliParsing.handle_errors](https://github.com/puppetlabs/puppetserver-ca-cli/blob/master/lib/puppetserver/ca/utils/cli_parsing.rb#L53) here. I'll do that unless anyone objects.

lib/puppetserver/ca/cli.rb
```diff
@@ -86,7 +87,15 @@ def self.run(cli_args = ARGV, out = STDOUT, err = STDERR)
           if exit_code
             return exit_code
           else
-            return action.run(input)
+            begin
+              return action.run(input)
+            rescue Puppetserver::Ca::Error => e
+              logger.err "Fatal error when running action '#{action_argument}'"
+              logger.err "  Arguments: #{general_options.merge(input)}"
```
Magisus: I'm a little on the fence about printing out the args here. in a lot of cases, it just seems unnecessarily verbose, because the invocation is usually right on the previous line.
```
puppetserver-ca-cli git:(server_list) ./exe/puppetserver-ca list
Fatal error when running action 'list'
  Arguments: {}
  Error: Failed connecting to https://localhost:8140/puppet-ca/v1/certificate_statuses/any_key
  Root cause: Failed to open TCP connection to localhost:8140 (Connection refused - connect(2) for "localhost" port 8140)
```
```
puppetserver-ca-cli git:(server_list) ./exe/puppetserver-ca list --all
Fatal error when running action 'list'
  Arguments: {"all"=>true}
  Error: Failed connecting to https://localhost:8140/puppet-ca/v1/certificate_statuses/any_key
  Root cause: Failed to open TCP connection to localhost:8140 (Connection refused - connect(2) for "localhost" port 8140)
```
Is there a particular thing you wanted to surface this way?

justinstoller: I think it was just YAGNI

Magisus: Alright I'm in favor of removing it then.

justinstoller:
Updated

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -239,6 +247,20 @@ def parse_crl_usage(setting)
             :ignore
           end
         end
+
+        def update_for_server_list!(settings)
+          if settings.dig(:server_list, 0, 0) &&
+              settings[:ca_server] == '$server'
+
+            settings[:ca_server] = settings.dig(:server_list, 0, 0)
```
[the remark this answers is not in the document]
justinstoller: Aahhhh.... We should be doing both. I was being thick.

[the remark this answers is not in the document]
justinstoller: No, we don't use server anywhere, nor should we ever. I don't even know who I am.

justinstoller:
I think this is properly updated?

Magisus:
Going to do a bit of additional testing with this.

Magisus:
Installed this in a package and used the `server_list` setting to configure it.

PR #46, opened 2018-12-13
(SERVER-2401) Handle new exceptions when checking CA status

justinstoller:
From 03da206 which implements most of this functionality:
```
Previously, we were catching ECONNREFUSED to determine if the server was
up. When we moved to custom error classes we no longer raise an
ECONNREFUSED, instead raising ConnectionFailed in it (and many other
potential exceptions) place.

So we should now be rescuing ConnectionFailed, but we only want to
swallow exceptions that originated from ECONNREFUSED, re-raising
ConnectionFaileds that originate from other exceptions.

To do so we add the ability to wrap an exception in our custom Error
class and update appropriate places where we were raising.
```

### DIFF
--- a/lib/puppetserver/ca/errors.rb
+++ b/lib/puppetserver/ca/errors.rb
@@ -1,6 +1,13 @@
 module Puppetserver
   module Ca
-    class Error < StandardError; end
+    class Error < StandardError
+      attr_reader :wrapped
+
+      def wrap(ex)
+        @wrapped = ex
+      end
+    end
+
     class FileNotFound < Error; end
     class InvalidX509Object < Error; end
     class ConnectionFailed < Error; end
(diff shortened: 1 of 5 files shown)


lib/puppetserver/ca/errors.rb
```diff
@@ -1,6 +1,20 @@
 module Puppetserver
   module Ca
-    class Error < StandardError; end
+    class Error < StandardError
+      def self.create(ex, msg)
```
Magisus: Are you doing it this way because you wanted to keep the same signature on the `initialize` from `StandardError`?

justinstoller: Yeah, and I didn't love the idea of making it a second optional argument to initialize.

justinstoller:
I snuck in f7fbc63 which changes the help output for generate from:
```
        --ca-client                  Whether this cert will be used to request CA actions.\
                    Causes the cert to be generated offline.
```
to 
```
        --ca-client                  Whether this cert will be used to request CA actions.
                                     Causes the cert to be generated offline.
```

PR #47, opened 2019-01-11
(SERVER-2424) Add action for generating infra CRL files

Magisus:
Putting this up to get a check on the approach. Still want to write a couple more tests.


Magisus:
Also the naming. I don't love it, but I'm not sure if we might want to add some other things to this later. For example, it might be nice to let people add certnames to the inventory file using this tool.

Magisus:
Hmm... thoughts on the Travis failure? It says it installs old bundler but then won't use it? Why only on Ruby 2.3?

justinstoller:
I think the approach is good, though I have three comments:
1. The name of the action is "infracrl", this feels awkward since the rest of the "init" actions are verbs.
2. Is not having an inventory file prexisting an error case or should we create an empty one (or one with the master's certname) if one does not already exist?
3. Can the locations of the infra files be changed via Puppet Server's config?

lib/puppetserver/ca/action/infracrl.rb
```diff
@@ -0,0 +1,114 @@
+require 'optparse'
+
+require 'puppetserver/ca/utils/cli_parsing'
```
justinstoller: super nit: these could be alphabetized

lib/puppetserver/ca/action/infracrl.rb
```diff
@@ -0,0 +1,114 @@
...
+          end
+
+          puppet = Config::Puppet.new(config_path)
+          puppet.load
+          settings = puppet.settings
+          return 1 if Errors.handle_with_usage(@logger, puppet.errors)
+
+          inventory_file = File.join(settings[:cadir], 'infra_inventory.txt')
+          if !File.exist?(inventory_file)
+            @logger.err "Please create an inventory file at '#{inventory_file}' with
```
justinstoller: are these newlines in the final error message?

lib/puppetserver/ca/action/infracrl.rb
```diff
@@ -0,0 +1,114 @@
...
+              the certnames of your infrastructure nodes before proceeding with
+              infra CRL setup!"
+            return 1
+          end
+
+          # This can be left blank, puppetserver will populate it with the serials for
+          # the certnames in the inventory file when the server starts
+          FileSystem.write_file(File.join(settings[:cadir], 'infra_serials'), '', 0644)
```
justinstoller: Should we checking if this exists first?

lib/puppetserver/ca/action/infracrl.rb
```diff
@@ -0,0 +1,114 @@
...
+          ca = LocalCertificateAuthority.new(signer.digest, settings)
+          infra_crl = ca.create_crl_for(ca.cert, ca.key)
+          return ca.errors if ca.errors.any?
+
+          # Drop the full leaf CRL from the chain
+          crl_chain = ca.crl_chain.drop(1)
+          # Add the new clean CRL, that will be populated with infra nodes only
+          # as they are revoked
+          crl_chain.unshift(infra_crl)
+          FileSystem.write_file(File.join(settings[:cadir], 'infra_crl.pem'), crl_chain, 0644)
```
justinstoller: Should we be checking if this exists?

lib/puppetserver/ca/action/infracrl.rb
Magisus: Oh, and not creating one if it does? Yeah probably.

lib/puppetserver/ca/action/infracrl.rb
Magisus: I think so, I want to reformat it, it looks funny.

Magisus:
1. Yeah, totally agreed, taking suggestions :) I wasn't sure how we felt about multi-word actions. I also even considered a flag under setup, but decided against it.
2. I wasn't sure about pre-creating it. I kind of like that it's a nice prompt in FOSS that if you want this feature to actually do something, you have to do that bit yourself, in case they missed it in the docs.
3. Yes, they can be set by config :/ But our `setup` action doesn't respect those settings. So I think it's a separate effort to make that happen.

Magisus:
Updated to check for existing files.

Still taking name suggestions. Given that all of our other actions are verbs, maybe it would make more sense for this to be a flag on `setup`, like `--ca-client`?

justinstoller:
It looks like travis' ruby will use its bundler version even if another is installed, but it says it should ignore errors from that one... https://docs.travis-ci.com/user/languages/ruby/#bundler-20  I think travis/bundler are ...in flux. Adding `gem uninstall -v '>= 2' -i $(rvm gemdir)@global -ax bundler || true` to the pre_install script should might help.

justinstoller:
Do we have a ticket for the server config work?

justinstoller:
I'm indifferent to whether we rename the action to be a verb or make this a flag on setup, though I can't think of a good single verb for enabling infra crl.

justinstoller:
Aside from the naming issue, 👍 code-wise.

Magisus:
Do you mean https://tickets.puppetlabs.com/browse/SERVER-2297?

justinstoller:
I guess so!

Magisus:
So an argument against a flag on `setup` is that its interaction with regular `setup` is potentially unclear. Like if you specify the flag, but the CA is also not initialized yet, does it do that part too? And then if the CA is there but these files are not, just do these files? That seems potentially confusing, but also potentially the way to most often "do the right thing", if we're gonna put it with `setup`. Alternatively, it could error and tell the person to run regular `setup` instead (which will generate everything, including the infra CRL files), but that maybe seems like worse UX?

Something kind of vague but potentially useful could be to create a command called something like `update`, which is designed to bring the state of the CA in line with configuration. Eventually it could be used to migrate the CA between directories (or to a database), but for now would simply generate missing files for toggleable settings (I thiiink this is the only one we have, but meh).

It's all kind of gross and potentially weird...

Magisus:
Updated to make the action called `enable`. This could later be paired with a `disable` action and be used to toggle other settings (e.g. `allow-subject-alt-names`).

lib/puppetserver/ca/action/enable.rb
```diff
@@ -0,0 +1,145 @@
...
+        SUMMARY = "Setup infrastructure CRL based on a node inventory."
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca enable [--help]
+  puppetserver ca enable [--config PATH] [--infracrl]
+
+Description:
+  Performs actions necessary to enable certain CA modes.
+
+  The default puppet.conf is queried for configuration If using a custom
```
tvpartytonight: Missing period between "configuration If".

lib/puppetserver/ca/action/enable.rb
```diff
@@ -117,12 +128,15 @@ def parse(cli_args)
         def self.parser(parsed = {})
           OptionParser.new do |opts|
             opts.banner = BANNER
-            opts.on('--help', 'Display this setup specific help output') do |help|
+            opts.on('--help', 'Display this `enable` specific help output') do |help|
```
justinstoller: I don't think any of the other actions put the action name in backticks.

lib/puppetserver/ca/action/enable.rb
```diff
@@ -10,25 +10,27 @@
...
-  Assumes the existence of an `infra_inventory.txt` file in the CA
-  directory listing the certnames of the infrastructure nodes in the
-  Puppet installation. Generates the `infra_serials` file and the empty
-  CRL to be populated with revoked infrastructure nodes.
+  Performs actions necessary to enable certain CA modes.
 
-  To determine the target location, the default puppet.conf
-  is consulted for custom values. If using a custom puppet.conf
-  provide it with the --config flag
+  The default puppet.conf is queried for configuration If using a custom
```
justinstoller: This is a flag that everything has but this is the only place we document it like this. I'd remove this section and if you think we need to describe it more see if this works in the general help output.

Magisus: Decided to just drop it, I think the option help text is sufficient.

Magisus:
Removed those docs and also fixed the help text to be more consistent, so that if we ever add i18n here we have fewer strings to translate.

PR #48, opened 2019-03-06
(SERVER-2478) Do not create an `infra_serials` file

Magisus:
Puppet Server will always generate an `infra_serials` file based on the
contents of `infra_inventory` on startup when the Infra CRL is enabled.
Therefore, the gem doesn't need to try to create it and doesn't need to
block if it is already present. This will help with some weird upgrade
scenarios where an inventory and infra_serials file are present, but the
CRL itself is not.


Magisus:
Related to https://github.com/puppetlabs/puppetserver/pull/2003

Magisus:
If we merge this first, I'll release the gem and add the update to the other PR.

PR #49, opened 2019-06-11
(SERVER-2465) Find signing cert and crl based on private key

pcarlisle:
Instead of requiring the signing certificate to be first in the certificate
bundle, find the certificate that matches the private key for the CA, and find
the CRL that is issued by that certificate.

### DIFF
--- a/lib/puppetserver/ca/x509_loader.rb
+++ b/lib/puppetserver/ca/x509_loader.rb
@@ -5,7 +5,7 @@ module Puppetserver
     # Load, validate, and store x509 objects needed by the Puppet Server CA.
     class X509Loader
 
-      attr_reader :errors, :certs, :key, :crls
+      attr_reader :errors, :certs, :cert, :key, :crls, :crl
 
       def initialize(bundle_path, key_path, chain_path)
         @errors = []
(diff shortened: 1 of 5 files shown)


lib/puppetserver/ca/x509_loader.rb
```diff
@@ -5,31 +5,61 @@ module Ca
...
+      attr_reader :errors, :certs, :cert, :key, :crls, :crl
 
       def initialize(bundle_path, key_path, chain_path)
         @errors = []
 
         @certs = load_certs(bundle_path)
         @key = load_key(key_path)
         @crls = load_crls(chain_path)
+        @cert = find_signing_cert
+        @crl = find_leaf_crl
```
justinstoller: Would it make sense to pass the needed objects into these methods?

pcarlisle: i've seen competing styles here including intentional refactors to move to using instance variables... i'm fine either way and probably prefer passing them in

pcarlisle: I do think it's okay to use instance variables in an object like a loader if you have clear lifecycles

lib/puppetserver/ca/x509_loader.rb
```diff
@@ -121,7 +151,7 @@ def validate_full_chain(certs, crls)
           crls.each {|crl| store.add_crl(crl) }
         end
 
-        unless store.verify(certs.first)
+        unless store.verify(@cert)
```
justinstoller: the last line of the this methods doc is now wrong

Magisus:
Gonna try this out.

Magisus:
So if you reorder the certs, `import` works fine, and the server starts okay, but when you try to sign something with the CLI you get
```
Error:
    When attempting to sign certificate request 'maggie', received
      code: 500
      body: Internal Server Error: java.lang.IllegalArgumentException: The first certificate in the certificate chain does not match the expected public key
```
And a corresponding error in the server logs:
```
2019-06-12 10:17:43,434 ERROR [qtp1976849875-40] [p.r.core] Internal Server Error: java.lang.IllegalArgumentException: The first certificate in the certificate chain does not match the expected public key
	at com.puppetlabs.ssl_utils.SSLUtils.pemToCaCert(SSLUtils.java:533)
	at puppetlabs.ssl_utils.core$eval3339$pem__GT_ca_cert__3344$fn__3345.invoke(core.clj:635)
	at puppetlabs.ssl_utils.core$eval3339$pem__GT_ca_cert__3344.invoke(core.clj:628)
	at puppetlabs.puppetserver.certificate_authority$eval37597$autosign_certificate_request_BANG___37602$fn__37606.invoke(certificate_authority.clj:1113)
	at puppetlabs.puppetserver.certificate_authority$eval37597$autosign_certificate_request_BANG___37602.invoke(certificate_authority.clj:1104)
	at puppetlabs.puppetserver.certificate_authority$eval38100$sign_existing_csr_BANG___38105$fn__38109.invoke(certificate_authority.clj:1394)
	at puppetlabs.puppetserver.certificate_authority$eval38100$sign_existing_csr_BANG___38105.invoke(certificate_authority.clj:1389)
	at puppetlabs.puppetserver.certificate_authority$eval38168$set_certificate_status_BANG___38173$fn__38174.invoke(certificate_authority.clj:1433)
	at puppetlabs.puppetserver.certificate_authority$eval38168$set_certificate_status_BANG___38173.invoke(certificate_authority.clj:1427)
	at puppetlabs.services.ca.certificate_authority_core$certificate_status$fn__41771$fn__41800.invoke(certificate_authority_core.clj:289)
```

That might be a good place to start fixing the issues in the CA. I'm not sure if there are also issues agent-side where this assumption is made, openssl might just handle it correctly. /cc @joshcooper

lib/puppetserver/ca/x509_loader.rb
justinstoller: maybe not worth it?

PR #50, opened 2019-06-12
(SERVER-2559) Check CA for errors before enabling infra crl

Sharpie:
This commit updates `puppetserver ca enable --infracrl` to check the CA
state for errors before attempting to generate the infrastructure CRL.
Prior to this commit, errors encountered when loading the CA state would
cause instance variables such as the certificate and private key to be
unset. This would result in CRL creation failing due to unexpected `nil`
values without providing clear messaging as to the root cause of the
problem.

### DIFF
--- a/lib/puppetserver/ca/action/enable.rb
+++ b/lib/puppetserver/ca/action/enable.rb
@@ -97,8 +97,9 @@ MSG
           return signer.errors if signer.errors.any?
 
           ca = LocalCertificateAuthority.new(signer.digest, settings)
-          infra_crl = ca.create_crl_for(ca.cert, ca.key)
           return ca.errors if ca.errors.any?
+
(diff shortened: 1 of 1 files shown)


lib/puppetserver/ca/action/enable.rb
```diff
@@ -97,9 +97,10 @@ def create_infra_crl_chain(settings)
           return signer.errors if signer.errors.any?
 
           ca = LocalCertificateAuthority.new(signer.digest, settings)
-          infra_crl = ca.create_crl_for(ca.cert, ca.key)
           return ca.errors if ca.errors.any?
 
+          infra_crl = ca.create_crl_for(ca.cert, ca.key)
+
```
Sharpie: Could add another call to `return ca.errors if ca.errors.any?` here, but as far as I can tell `create_crl_for` does not add anything to the `errors` array if a failure occurs.

PR #51, opened 2019-06-18
(SERVER-2549) Add generate-csr action

thrnio:
This adds a new action, `generate-csr`, which can be used to generate a CSR and
private key for an intermediate CA. This CSR can then be signed by external PKI
infrastructure and then imported using the `import` action.

I just kind of picked a way to expose the functionality, but it should be pretty straightforward to either rename the command or move the functionality under an existing command.

### DIFF
--- a/lib/puppetserver/ca/action/generate_csr.rb
+++ b/lib/puppetserver/ca/action/generate_csr.rb
@@ -1,0 +1,123 @@
+require 'optparse'
+
+require 'puppetserver/ca/config/puppet'
+require 'puppetserver/ca/errors'
+require 'puppetserver/ca/local_certificate_authority'
+require 'puppetserver/ca/utils/cli_parsing'
+require 'puppetserver/ca/utils/file_system'
+require 'puppetserver/ca/utils/signing_digest'
+
+module Puppetserver
+  module Ca
+    module Action
+      class GenerateCsr
+        include Puppetserver::Ca::Utils
+
+        SUMMARY = 'Generate a CSR for an intermediate signing CA cert'
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca generate-csr [--help]
+  puppetserver ca generate-csr --output-dir PATH [--config PATH] [--ca-name NAME]
+
+Description:
+  Setup a root and intermediate signing CA for Puppet Server
+  and store generated CA keys, certs, crls, and associated
+  master related files on disk.
+
+Options:
+        BANNER
+
+        def initialize(logger)
+          @logger = logger
+        end
+
+        def run(input)
+          # Validate config_path provided
+          config_path = input['config']
+          if config_path
+            errors = FileSystem.validate_file_paths(config_path)
+            return 1 if Errors.handle_with_usage(@logger, errors)
+          end
+
+          # Load, resolve, and validate puppet config settings
+          settings_overrides = {}
+          settings_overrides[:ca_name] = input['ca-name'] unless input['ca-name'].empty?
+
+          puppet = Config::Puppet.new(config_path)
+          puppet.load(settings_overrides)
+          return 1 if Errors.handle_with_usage(@logger, puppet.errors)
+
+          # Load most secure signing digest we can for cers/crl/csr signing.
+          signer = SigningDigest.new
+          return 1 if Errors.handle_with_usage(@logger, signer.errors)
+
+          # Generate intermediate ca and put the results in the specified output directory
+          generate_csr(puppet.settings, signer.digest, input['output-dir'])
+          return 0
+        end
+
+        def generate_csr(settings, signing_digest, dir)
+          ca = Puppetserver::Ca::LocalCertificateAuthority.new(signing_digest, settings)
+
+          key, csr = ca.create_intermediate_csr
+
+          write_generated_files(key, csr, dir)
+        end
+
+        def write_generated_files(key, csr, dir)
+          FileSystem.write_file(File.join(dir, 'ca.key'), key.to_s, 0640)
+          FileSystem.write_file(File.join(dir, 'ca.csr'), csr.to_s, 0640)
+        end
+
+        def check_for_existing_ssl_files(dir)
+          files = [ File.join(dir, 'ca.key'), File.join(dir, 'ca.csr')]
+          errors = Puppetserver::Ca::Utils::FileSystem.check_for_existing_files(files)
+          if !errors.empty?
+            errors << 'Please delete these files if you want to generate a CSR'
+          end
+
+          errors
+        end
+
+        def parse(cli_args)
+          results = {}
+          parser = self.class.parser(results)
+          errors = CliParsing.parse_with_errors(parser, cli_args)
+
+          if results['output-dir'].nil? || results['output-dir'].empty?
+            errors << '    Must specify an output directory to store generated files in'
+          elsif !Dir.exist?(results['output-dir'])
+            errors << '    Specified output directory must exist'
+          else
+            errors.concat(check_for_existing_ssl_files(results['output-dir']))
+          end
+
+          errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
+          exit_code = errors_were_handled ? 1 : nil
+          return results, exit_code
+        end
+
+        def self.parser(parsed = {})
+          parsed['ca-name'] = ''
+          OptionParser.new do |opts|
+            opts.banner = BANNER
+            opts.on('--help', 'Display this command-specific help output') do |help|
+              parsed['help'] = true
+            end
+            opts.on('--output-dir PATH',
+                    'Path to directory to store generated key and csr in') do |dir|
+              parsed['output-dir'] = dir
+            end
+            opts.on('--config CONF', 'Path to puppet.conf') do |conf|
+              parsed['config'] = conf
+            end
+            opts.on('--ca-name NAME',
+                    'Common name to use for the CA signing cert') do |name|
+              parsed['ca-name'] = name
+            end
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -235,6 +235,13 @@ def create_intermediate_cert(root_key, root_cert)
         return nil
       end
 
+      def create_intermediate_csr
+        key = @host.create_private_key(@settings[:keylength])
+        csr = @host.create_csr(name: @settings[:ca_name], key: key)
```
Magisus: I think we want to add the `CA_EXSTENSIONS` to the CSR in this case, since we don't control the signing process. There is an example of that in here https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/certificate_request.rb in Puppet. Not sure exactly which bit we need, but likely something like https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/certificate_request.rb#L248 and https://github.com/puppetlabs/puppet/blob/master/lib/puppet/ssl/certificate_request.rb#L95.

spec/puppetserver/ca/action/generate_csr_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  describe 'flags' do
+    it 'takes a single output directory' do
+      Dir.mktmpdir do |tmpdir|
+        result, maybe_code = subject.parse(['--output-dir', tmpdir])
+        expect(maybe_code).to eq(nil)
+        expect(result['output-dir']).to eq(tmpdir)
+      end
+    end
+
+    describe 'flags' do
```
justinstoller: dupe describe block from above

lib/puppetserver/ca/action/generate_csr.rb
```diff
@@ -0,0 +1,123 @@
...
+        def write_generated_files(key, csr, dir)
+          FileSystem.write_file(File.join(dir, 'ca.key'), key.to_s, 0640)
+          FileSystem.write_file(File.join(dir, 'ca.csr'), csr.to_s, 0640)
+        end
+
+        def check_for_existing_ssl_files(dir)
+          files = [ File.join(dir, 'ca.key'), File.join(dir, 'ca.csr')]
+          errors = Puppetserver::Ca::Utils::FileSystem.check_for_existing_files(files)
+          if !errors.empty?
+            errors << 'Please delete these files if you want to generate a CSR'
```
justinstoller: what files?

lib/puppetserver/ca/action/generate_csr.rb
```diff
@@ -0,0 +1,123 @@
...
+          ca = Puppetserver::Ca::LocalCertificateAuthority.new(signing_digest, settings)
+
+          key, csr = ca.create_intermediate_csr
+
+          write_generated_files(key, csr, dir)
+        end
+
+        def write_generated_files(key, csr, dir)
+          FileSystem.write_file(File.join(dir, 'ca.key'), key.to_s, 0640)
+          FileSystem.write_file(File.join(dir, 'ca.csr'), csr.to_s, 0640)
```
justinstoller: should the output files be configurable?

justinstoller: having the dir configurable is probably enough, sorry

lib/puppetserver/ca/action/generate_csr.rb
thrnio: I based that on `action/generate.rb` which does the same thing. 

`Puppetserver::Ca::Utils::FileSystem.check_for_existing_files` adds `"Existing file at '#{path}'"` to `errors` for each file that already exists from the list, so those messages will also present, which is why this error message doesn't specify them. There's also a test that checks that stderr contains `ca.csr` and `ca.key` in the case of existing files.

lib/puppetserver/ca/action/generate_csr.rb
```diff
@@ -0,0 +1,123 @@
...
+            errors.concat(check_for_existing_ssl_files(results['output-dir']))
+          end
+
+          errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
+          exit_code = errors_were_handled ? 1 : nil
+          return results, exit_code
+        end
+
+        def self.parser(parsed = {})
+          parsed['ca-name'] = ''
```
justinstoller: Are we setting `ca-name` to be a string here so we can call `empty?` on it later? Would it work to just check for the key's existence later on and not give a default here? I don't think we should be unconditionally setting the value of the passed in hash to an empty string.

lib/puppetserver/ca/action/generate_csr.rb
justinstoller: I think CSRs are usually world readable?

lib/puppetserver/ca/action/generate_csr.rb
```diff
@@ -0,0 +1,123 @@
...
+        SUMMARY = 'Generate a CSR for an intermediate signing CA cert'
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca generate-csr [--help]
+  puppetserver ca generate-csr --output-dir PATH [--config PATH] [--ca-name NAME]
+
+Description:
+  Setup a root and intermediate signing CA for Puppet Server
+  and store generated CA keys, certs, crls, and associated
+  master related files on disk.
```
justinstoller: I don't think this is the correct description

lib/puppetserver/ca/action/generate_csr.rb
```diff
@@ -0,0 +1,123 @@
...
+  module Ca
+    module Action
+      class GenerateCsr
+        include Puppetserver::Ca::Utils
+
+        SUMMARY = 'Generate a CSR for an intermediate signing CA cert'
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca generate-csr [--help]
+  puppetserver ca generate-csr --output-dir PATH [--config PATH] [--ca-name NAME]
```
justinstoller: I think I like the `prepare` name better. Maybe with a `--csr` flag for this behavior? Something like prepare would be too vague to do this w/o specifying an additional flag, I think. But I don't know what would really go with other preparation... creating the directories?? I can only think of "check to make sure my file system's correct" or "check that these artifacts are correct" which are checks we already do in the setup and generate actions.

Or, on the other hand, something more specific like `ca-request`, so `generate-csr` couldn't be construed to mean a regular node csr.

Thoughts @Magisus 	?

spec/utils/ssl/ca.rb
```diff
@@ -0,0 +1,199 @@
+require 'openssl'
+
+module Utils
+  module SSL
+    class CA
```
justinstoller: can we combine and share some of this with spec/utils/ssl.rb?

lib/puppetserver/ca/action/generate_csr.rb
thrnio: I like `prepare`, but the fact that we would need something like a `--csr` has a bad smell to me as I think about it. Needing a flag to choose a subaction when we're not sure what else we would use the command for seems confusing. But `ca-request` *also* makes me hesitate because it's not actually a verb, while the rest of the actions are. Something like `generate-ca-request` is almost certainly too long, though, so idk.

spec/utils/ssl/ca.rb
thrnio: done

lib/puppetserver/ca/action/generate_csr.rb
thrnio: fixed

lib/puppetserver/ca/action/generate_csr.rb
thrnio: seems reasonable

lib/puppetserver/ca/local_certificate_authority.rb
thrnio: should be working and tested

spec/puppetserver/ca/action/generate_csr_spec.rb
thrnio: oops, fixed

lib/puppetserver/ca/action/generate_csr.rb
thrnio: good catch, fixed

lib/puppetserver/ca/action/generate_csr.rb
Magisus: I think if we're going to have a generic command, we already have one that makes the most sense: `generate`. It already has a special case flag for generating a CA client cert, `--ca-client`, and I don't personally have a problem with giving it a different one, `--ca-csr`, which generates a new key and CSR with CA extensions. The state of the existing CA doesn't matter at this point: it's valid to generate one of these CSRs whether or not you have an existing CA. You're then free to do what you like it with: use it as a new intermediate for your Puppet CA, or use it to bootstrap the Puppet CA with an external root.

I know we've been splitting these actions into "stuff to set up the CA" vs "stuff to do with the CA you already have". But in my mind this one could be either, depending on what you wanted to do with it. In fact, our help text is already misleading about `generate`, because it already has a mode in which the server needs to be stopped, but is still listed in the "require running" [section](https://github.com/puppetlabs/puppetserver-ca-cli/blob/master/lib/puppetserver/ca/cli.rb#L27-L51). So I'm not sure we should tie ourselves to that dichotomy.

It might be useful to get an opinion from CS on this.

thrnio: The other thing that comes up with using `generate` is the question of what the behavior should be. Right now with the separate command, it writes out to a required specified directory, but that doesn't seem appropriate if we use `generate`­­­—conditionally required parameters seem like a bad idea. I'm not sure what would be appropriate behavior in that case.

PR #52, opened 2019-06-26
(maint) Add executable to gemspec

pcarlisle:
This makes it possible to run the executable with bundler from a git
checkout.


justinstoller:
Will this also cause the `puppetserver-ca` file to be installed when users install the gem?

Magisus:
And do we consider that a good or a bad thing?

pcarlisle:
It seems like it will.

PR #53, opened 2019-07-10
(SERVER-2589) Add a --certname option to `list` action

reidmv:
The new --certname option on the list action behaves similiarly to how
the option operates on other actions, in this case by limiting the
output to the certnames specified, and causing the command to return a
non-zero exit code in the event a certname specified is not able to be
listed (because it is not present).

This brings `puppetserver ca list` closer to parity with older commands
such as `puppet cert list`.

Initial implementation is not performance optimized. It doesn't look
like the certificate_statuses API actually supports returning a filtered
list, so this commit is forced to filter the results of requesting all
certificate statuses.

### DIFF
--- a/lib/puppetserver/ca/action/list.rb
+++ b/lib/puppetserver/ca/action/list.rb
@@ -89,7 +89,7 @@ Options:
         end
 
         def output_certs_by_state(requested, signed = [], revoked = [], missing = [])
-          if revoked.empty? && signed.empty? && requested.empty?
+          if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
             @logger.inform "No certificates to list"
             return
           end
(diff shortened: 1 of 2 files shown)


lib/puppetserver/ca/action/list.rb
```diff
@@ -60,16 +71,24 @@ def run(input)
...
 
-          return 0
+          return missing.any? \
+            ? 1
+            : 0
         end
 
-        def output_certs_by_state(requested, signed = [], revoked = [])
+        def output_certs_by_state(requested, signed = [], revoked = [], missing = [])
           if revoked.empty? && signed.empty? && requested.empty?
```
Magisus: If you only input one certname and it's missing, this will be true and we will return, not printing the warning about missing certs.

Magisus:
Found one bug, otherwise this looks good to me.

Magisus:
If we can get this updated and merged before sometime tomorrow, it can go out in the upcoming Z releases.

reidmv:
@Magisus submitted fix commit for that bug

PR #55, opened 2019-08-07
(maint) Allow Bundler version to float

Sharpie:
This commit changes the Bundler version constraint in the gemspec file
from `~> 1.16` to `>= 1.16` and allows Bundler 2.x to be used.


pcarlisle:
this is also in #52

PR #56, opened 2019-08-14
(SERVER-2552) Initialize leaf CRL during import

Sharpie:
This commit updates the `puppetserver ca import` command to initialize
a CRL for the intermediate CA if the loader does not find one in the
`crl-chain` file. Not having a pre-generated CRL is the typical case as
creating a CRL requires access to the private key of the intermediate
CA. An external CA generally should not have access to this key and is
therefore not able to provide a pre-generated CRL along with the
intermediate CA cert.

### DIFF
--- a/lib/puppetserver/ca/local_certificate_authority.rb
+++ b/lib/puppetserver/ca/local_certificate_authority.rb
@@ -70,6 +70,35 @@ module Puppetserver
         @crl = loader.crl
       end
 
+      # Initialize SSL state
+      #
+      # This method is similar to {#load_ssl_components}, but has extra
+      # logic for initializing components that may not be present when
+      # the CA is set up for the first time. For example, SSL components
+      # provided by an external CA will often not include a pre-generated
+      # leaf CRL.
+      #
+      # @note Check {#errors} after calling this method for issues that
+      #   may have occurred during initialization.
+      #
+      # @param loader [Puppetserver::Ca::X509Loader]
+      # @return [void]
+      def initialize_ssl_components(loader)
+        @cert_bundle = loader.certs
+        @key = loader.key
+        @cert = loader.cert
+
+        if loader.crl.nil?
+          loader.crl = create_crl_for(@cert, @key)
+
+          loader.validate_full_chain(@cert_bundle, loader.crls)
+          @errors += loader.errors
+        end
+
+        @crl_chain = loader.crls
+        @crl = loader.crl
+      end
+
       def errors
         @errors += @host.errors
       end
--- a/lib/puppetserver/ca/x509_loader.rb
+++ b/lib/puppetserver/ca/x509_loader.rb
@@ -33,15 +33,15 @@ module Puppetserver
         signing_cert
       end
 
+      # Find a CRL in the chain issued by the signing cert
+      #
+      # @return [OpenSSL::X509::CRL] If a CRL is found.
+      # @return [nil] If no CRL is found.
       def find_leaf_crl
         return if @crls.empty? || @cert.nil?
 
         leaf_crl = @crls.find do |crl|
           crl.issuer == @cert.subject
-        end
-
-        if leaf_crl.nil?
-          @errors << 'Could not find CRL issued by CA certificate'
         end
 
         leaf_crl
@@ -59,7 +59,7 @@ module Puppetserver
           validate_cert_and_key(pkey, @cert)
         end
 
-        unless bundle.empty? || @cert.nil?
+        unless bundle.empty? || @cert.nil? || @crl.nil?
           validate_full_chain(bundle, chain)
         end
       end
@@ -122,6 +122,15 @@ module Puppetserver
         end
 
         return crls
+      end
+
+      # Replace the CRL for the signing cert of this loader
+      #
+      # @param new_crl [OpenSSL::X509::CRL]
+      # @return [void]
+      def crl=(new_crl)
+        @crl = new_crl
+        @crls = [new_crl] + @crls.reject {|c| c.issuer == new_crl.issuer }
       end
 
       def validate_cert_and_key(key, cert)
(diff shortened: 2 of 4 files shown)


lib/puppetserver/ca/x509_loader.rb
```diff
@@ -124,6 +124,15 @@ def load_crls(chain_path)
         return crls
       end
 
+      # Replace the CRL for the signing cert of this loader
+      #
+      # @param new_crl [OpenSSL::X509::CRL]
+      # @return [void]
+      def crl=(new_crl)
+        @crl = new_crl
+        @crls = [new_crl] + @crls.reject {|c| c.issuer == new_crl.issuer }
```
joshcooper: Not sure if this applies but IIRC `OpenSSL::X509::Name` doesn't implement comparable until ruby 2.4, so this comparison may not work they way you want. I handled this in puppet by calling `X509#to_utf8`, however, I had to monkey patch that for older ruby versions 🤦‍♂

Sharpie: Hmm. I cribbed this from up in the `find_leaf_crl` method where we compare `crl.issuer` against `cert.subject`, which is also a `OpenSSL::X509::Name`:

https://github.com/puppetlabs/puppetserver-ca-cli/blob/1.3.2/lib/puppetserver/ca/x509_loader.rb#L39-L41

Given the spec tests run under Ruby 2.3, it seems like some comparison logic must be in place.

Sharpie: Looks like `OpenSSL::X509::Name` does include the `Comparable` module in Ruby 2.3 and also implements equality methods:

https://github.com/ruby/ruby/blob/v2_3_0/ext/openssl/ossl_x509name.c#L454

joshcooper: Gah, sorry, I'm misremembering. It's `OpenSSL::X509::Certificate` that doesn't implement `#==`:

```
irb(main):004:0> left = OpenSSL::X509::Certificate.new(File.read('./spec/fixtures/ssl/ca.pem'))
=> #<OpenSSL::X509::Certificate: subject=#<OpenSSL::X509::Name:0x00007ff5599376b8>, issuer=#<OpenSSL::X509::Name:0x00007ff559937690>, serial=#<OpenSSL::BN:0x00007ff559937640>, not_before=1970-01-01 00:00:00 UTC, not_after=2029-04-19 22:31:22 UTC>
irb(main):005:0> right  = OpenSSL::X509::Certificate.new(File.read('./spec/fixtures/ssl/ca.pem'))
=> #<OpenSSL::X509::Certificate: subject=#<OpenSSL::X509::Name:0x00007ff559926660>, issuer=#<OpenSSL::X509::Name:0x00007ff559926728>, serial=#<OpenSSL::BN:0x00007ff559926818>, not_before=1970-01-01 00:00:00 UTC, not_after=2029-04-19 22:31:22 UTC>
irb(main):006:0> left == right
=> false
```

Sharpie: Ah, nice to know about certificates. Looks like Ruby calls through to `X509_NAME_cmp` when comparing `OpenSSL::X509::Name` which does a comparison of the bytes in the canonical encoding of the Name:

https://github.com/openssl/openssl/blob/OpenSSL_1_1_1a/crypto/x509/x509_cmp.c#L179

justinstoller:
I think a side effect of this will be if you corrupt your local CRL then it will create a new one for you when you run `puppetserver ca generate`

Sharpie:
I would expect a corrupted CRL to cause the loader to register a "Could not parse entry" error during `load_crls`:

https://github.com/puppetlabs/puppetserver-ca-cli/blob/1.3.2/lib/puppetserver/ca/x509_loader.rb#L107-L113

justinstoller:
If the header/footer no longer matches, then it won't trigger the line you linked to which I'll admit is far fetched. I guess what I meant is that this behavior is desired for the import action, but the method updated to generate the crl is called anytime the local ca is instantiated, which is not desirable, and could mask some other issue.

Sharpie:
If no header / footer pairs are present to match, then the next line down in `load_crls` should trigger with "Could not detect any crls".

Any other mismatch seems extremely unlikely to parse as valid ASN.1.

Sharpie:
And, for the edge case where two corrupted CRLs happen to line up in a way that parses, a full chain validation is still done and should fail due to signatures no longer being valid.

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -66,6 +66,14 @@ def load_ssl_components(loader)
         @cert_bundle = loader.certs
         @key = loader.key
         @cert = loader.cert
+
+        if loader.crl.nil?
+          loader.crl = create_crl_for(@cert, @key)
```
justinstoller: `load_ssl_components` is called in `initialize`. I think this creating of the crl is behavior that is desirable in the import action, but not necessarily in every case that we'd instantiate the local CA. I think this logic should be moved to, or at least driven from, the import action in a way that avoids it being called on every `LocalCertificateAuthority.new`.

Sharpie: I initially tried to do it that way, but the problem there is that `LocalCertificateAuthority` owns the logic for creating CRLs which depends on the `@digest` that the CA loads from Puppet configuration. The CA also can't create a CRL until the cert and private key have been loaded.

One way to cut the knot may be to move this logic to a different method on `LocalCertificateAuthority`, say `initialize_ssl_components`, that is only called by the `import` action. Would that be a workable solution?

justinstoller: sounds good!

Sharpie: The CRL creation has been moved to a separate `initialize_ssl_components` action.

PR #57, opened 2019-11-27
(PE-27489) Add ability to specify ttl

tvpartytonight:
This change adds in the ability to specify a ttl for the generate and
sign actions.

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -75,6 +75,9 @@ BANNER
                     'Causes the cert to be generated offline.') do |ca_client|
               parsed['ca-client'] = true
             end
+            opts.on('--ttl TTL', 'The time-to-live for each cert generated and signed') do |ttl|
+              parsed['ttl'] = ttl
+            end
           end
         end
 
@@ -140,7 +143,7 @@ BANNER
             return 1 if check_server_online(puppet.settings)
             all_passed = generate_authorized_certs(certnames, alt_names, puppet.settings, signer.digest)
           else
-            all_passed = generate_certs(certnames, alt_names, puppet.settings, signer.digest)
+            all_passed = generate_certs(certnames, alt_names, puppet.settings, signer.digest, input['ttl'])
           end
           return all_passed ? 0 : 1
         end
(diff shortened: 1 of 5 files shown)


lib/puppetserver/ca/action/generate.rb
```diff
@@ -232,7 +235,7 @@ def generate_certs(certnames, alt_names, settings, digest)
               @logger.inform "Certificate for #{certname} was autosigned."
               true
```
Magisus: Can we add something in here (a warning maybe?) that when the cert is autosigned, the TTL will be ignored?

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -35,11 +45,40 @@ def make_ca_url(resource_type = nil, certname = nil)
...
+        match = /^(\d+)(s|m|h|d|y)?$/.match(ttl)
+        if match
+          if match[2]
+            match[1].to_i * UNITMAP[match[2]].to_i
+          else
+            ttl
+          end
+        else
+          @logger.err "Error:"
+          @logger.err " '{ttl}' is an invalid ttl value"
```
Magisus: I think this needs a '#' on it

Magisus:
I also manually tested this, and it seems to work great!

lib/puppetserver/ca/action/generate.rb
```diff
@@ -75,6 +75,9 @@ def self.parser(parsed = {})
                     'Causes the cert to be generated offline.') do |ca_client|
               parsed['ca-client'] = true
             end
+            opts.on('--ttl TTL', 'the time-to-live for each cert generated and signed') do |ttl|
```
justinstoller: All of the other help texts start with a capital letter.

lib/puppetserver/ca/action/sign.rb
```diff
@@ -32,6 +32,9 @@ class Sign
         def self.parser(parsed = {})
           OptionParser.new do |opts|
             opts.banner = BANNER
+            opts.on('--ttl TTL', 'the time-to-live for each cert signed') do |ttl|
```
justinstoller: ditto

lib/puppetserver/ca/action/generate.rb
```diff
@@ -230,9 +233,12 @@ def generate_certs(certnames, alt_names, settings, digest)
             # Check if the CA autosigned the cert
             if download_cert(ca, certname, settings)
```
justinstoller: could we move this into a separate method? the containing method is already big and getting bigger, this conditional seems like a good chunk to split out??

PR #59, opened 2020-04-01
(SERVER-2591) Display auth extensions for certificates

pcarlisle:


pcarlisle:
Something I ran into here is that if there are no alt names for a cert the whole field is skipped on display, which is more of an issue now that there's a potential third field after. If the output is meant to be parseable at all then each field needs to be displayed for all certs or none of them.

Magisus:
I don't think the default output is _really_ meant to be parseable. We have a ticket for adding JSON formatted output: https://tickets.puppetlabs.com/browse/SERVER-2252.

Magisus:
Current output:
```
Signed Certificates:
    localhost                             (SHA256)  73:0E:9A:FD:14:1E:F0:9C:FA:F0:65:42:9B:B0:14:91:F5:3D:43:47:64:83:10:80:97:73:73:FC:37:A0:EB:C1	alt names: ["DNS:puppet", "DNS:localhost"]	authorization extensions: [pp_cli_auth: true]
    maggie                                (SHA256)  E4:1A:49:5C:70:D5:77:51:E0:02:7B:D8:C4:E3:CC:49:17:19:C5:27:0D:A4:D8:5C:1E:F4:B3:C1:59:B4:5F:36	authorization extensions: []
Revoked Certificates:
    testing       (SHA256)  5C:10:41:AA:02:8A:6B:B8:15:2A:4A:05:0A:1A:F4:41:7C:8E:90:42:13:AF:7A:2D:7D:B8:DA:FB:88:8A:85:13	authorization extensions: []
```
So looking at that, alt names aren't displayed when there aren't any, but auth exts are. I think we should pick one and be consistent. Since this out put is supposed to be human-readable, I think I lean towards not displaying it when it's empty. And then maybe we can also do the JSON ticket I mentioned some time soon.

mwaggett:
+1 on removing the pry stuff from the Gemfile, since they're just developer tools.

pcarlisle:
Okay it's updated per feedback. I do think it might be worth having the discussion about output a little more broadly because though I've changed it here I don't think that I agree with it. As a heavy command line user I just want my software to output text that I can consistently parse with the basic unix cli tools, and json is not a solution to that.

pcarlisle:
Also fine putting that off.

Magisus:
Fair enough. What format makes the most sense to you? (We can either ticket it or fix it here, no strong opinion)

Magisus:
Also @justinstoller you filed the original format flag tickets. Thoughts on this?

justinstoller:
Happy to have that conversation, @pcarlisle .

Originally, I had it in my head that the default CLI output should be a "pretty" formatted thing, somewhat based on the original output of puppet cert. This was informed by UX research that we did with puppet cert a long time ago (see the [ticket](https://tickets.puppetlabs.com/browse/PUP-6268) and the UX test [results](https://confluence.puppetlabs.com/display/RAPArchive/Cert+Information+Display+Research+-+June%2C+2016) and eventual [PR](https://github.com/puppetlabs/puppet/pull/5023)). I think that sentiment gels with the kind of CLI output decisions that are happening elsewhere in the company (I'm thinking of Eric0's recent relay demo, specifically).

I tried to find current guidance that we may have for CLI tools and I didn't find any. I found the old [puppet-nogui/patterns/cli_interaction.md](https://github.com/puppetlabs/puppet-nogui/blob/master/patterns/cli_interaction.md), but it contains both "The default approach to the CLI should be non-interactive" & "Build output for humans first" which seem to somewhat contradictory. And the whole thing is really old.

I think that those guidelines I mentioned earlier are nice, but probably not company policy anymore and we can could be pretty free to do what we feel is best, as long as it's defensible. Assuming we're not going to add the `--format` flag anytime soon (which could support json and/or line oriented output), I don't have a strong opinion on the direction the default formatting should go now.

PR #61, opened 2020-10-23
(SERVER-2894) Add migrate command

tvpartytonight:


lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -159,6 +159,37 @@ def make_store(bundle, crl_usage, crls = nil)
...
+        # Returns true if it receives back a response of "running", and false if
+        # no connection can be made, or a different response is received.
+        def self.check_server_online(settings, logger)
+          status_url = URL.new('https', settings[:ca_server], settings[:ca_port], 'status', 'v1', 'simple', 'ca')
+          begin
+            # Generating certs offline is necessary if the master cert has been destroyed
+            # or compromised. Since querying the status endpoint does not require a client cert, and
+            # we commonly won't have one, don't require one for creating the connection.
+            # Additionally, we want to ensure the server is stopped before migrating the CA dir to
+            # avoid issues with writing to the ca dir and moving it.
```
Magisus, line 173: `ca` -> `CA`

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+require 'puppetserver/ca/utils/http_client'
+
+module Puppetserver
+  module Ca
+    module Action
+      class Migrate
+        include Puppetserver::Ca::Utils
+        PUPPETSERVER_CA_DIR = '/etc/puppetlabs/puppetserver/ca'
+
+        SUMMARY = "Migrate the CA directory out from puppet/ssl/ca to puppetserver/ca"
```
Magisus, line 12: "Migrate the CA directory from its current location to `/etc/puppetlabs/puppetserver/ca`"

Since I think this can migrate from wherever it's currently configured, right?

spec/puppetserver/ca/action/migrate_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  subject { Puppetserver::Ca::Action::Migrate.new(logger) }
+
+  describe 'flags' do
+    it 'has no required arguments' do
+      result, maybe_code = subject.parse([])
+      expect(maybe_code).to be(nil)
+      expect(result['config']).to eq(nil)
+    end
+    it 'takes a custom puppet.conf location' do
```
Magisus, line 19: Missing newline

spec/puppetserver/ca/action/migrate_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  describe 'validation' do
+    it 'prints the help output & returns 1 if invalid flags are given' do
+      _, code = subject.parse(['--hello', '--certname', "amy.net"])
+      expect(code).to eq(1)
+      expect(stderr.string).to match(/Error.*--hello/m)
+    end
+  end
+
+  describe '#run' do
+    let(:filesystem) { Puppetserver::Ca::Utils::FileSystem}
```
Magisus, line 35: Missing space before last bracket

Magisus (commented):
This looks good! Just a couple minor stylistic notes. Thanks for adding those nice explanations of the error cases.

lib/puppetserver/ca/action/migrate.rb
tvpartytonight: Yeah, that's correct.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+  puppet.conf file in your installation, or supply one using the `--config` flag.
+Options:
+BANNER
+
+        def initialize(logger)
+          @logger = logger
+        end
+
+        def run(input)
+          config_path = input['config']
```
mwaggett, line 30: How is the default config path handled, i.e. when the `--config` flag isn't passed? We might want a test for that?

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+          puppet = Config::Puppet.new(config_path)
+          puppet.load
+          return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+
+          errors = FileSystem.check_for_existing_files(PUPPETSERVER_CA_DIR)
+          if !errors.empty?
+            instructions = <<-ERR
+Migration will not overwrite the directory at #{PUPPETSERVER_CA_DIR}. Have you already
+run this migration tool? It is likely that you have already successfully run the migration
+or do not need to run it.
```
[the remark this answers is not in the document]
tvpartytonight: They could already be on Puppet 7; I am reticent to tell anybody to delete their CA without actually having enough context to know that it is safe to do so.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+SUCCESS_MESSAGE
+          return 0
+        end
+
+        def migrate(old_cadir, new_cadir=PUPPETSERVER_CA_DIR)
+          FileUtils.mv(old_cadir, new_cadir)
+          FileUtils.symlink(new_cadir, old_cadir)
+        end
+
+        def parse(args)
```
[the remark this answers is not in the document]
tvpartytonight: The main cli class calls this for each action.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+          results = {}
+          parser = self.class.parser(results)
+          errors = CliParsing.parse_with_errors(parser, args)
+          errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
+          exit_code = errors_were_handled ? 1 : nil
```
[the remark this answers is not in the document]
tvpartytonight: It needs to be nil, not 0 to match checking in the primary cli class.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca migrate [--help]
+  puppetserver ca migrate [--config PATH]
+
+Description:
+  Migrate an existing CA directory in the puppet directory to puppetserver. This is for
+  upgrading from Puppet platform 6.x to Puppet 7. Use the currently configured
+  puppet.conf file in your installation, or supply one using the `--config` flag.
+Options:
```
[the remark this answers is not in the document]
tvpartytonight: This is how all the actions format their `Banner` data; the `parser` method is called and to fill in specific options for each action class.

lib/puppetserver/ca/action/migrate.rb
tvpartytonight: There is a test for when the `--config` flag is not passed, in the spec for "no required arguments".

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -159,6 +159,37 @@ def make_store(bundle, crl_usage, crls = nil)
...
+            self.new(settings, with_client_cert: false).with_connection(status_url) do |conn|
+              result = conn.get
+              if result.body == "running"
+                logger.err "CA service is running. Please stop it before attempting to run this command."
+                true
+              else
+                false
+              end
+            end
+            true
```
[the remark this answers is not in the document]
tvpartytonight: good catch! This shouldn't be there, as it will indeed override that false above.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+        PUPPETSERVER_CA_DIR = '/etc/puppetlabs/puppetserver/ca'
+
+        SUMMARY = "Migrate the CA directory /etc/puppetlabs/puppetserver/ca"
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca migrate [--help]
+  puppetserver ca migrate [--config PATH]
+
+Description:
+  Migrate an existing CA directory in the puppet directory to puppetserver. This is for
```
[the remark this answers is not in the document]
tvpartytonight: Feel like it's already in the summary, but I can add it to the description too if you feel strongly.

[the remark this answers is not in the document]
Magisus: The former should be left vague, as we just look at the setting and then move it. The second one, I agree should be more explicit, especially since it isn't configurable.

lib/puppetserver/ca/action/migrate.rb
Magisus: Also fwiw, the `Config::Puppet` class handles the case where `config_path` is nil: https://github.com/puppetlabs/puppetserver-ca-cli/blob/master/lib/puppetserver/ca/config/puppet.rb#L35-L41

tvpartytonight:
I think I've addressed all the feedback, and it's finally run on travis-ci.com. Let me know if there's anything I missed.

lib/puppetserver/ca/action/migrate.rb
mwaggett: I saw the test for when `--config` isn't passed, but it doesn't test that a default is used, just that it's nil. But I guess that is handled elsewhere.

lib/puppetserver/ca/action/migrate.rb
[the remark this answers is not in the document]
mwaggett: ahh gotcha, thanks

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -0,0 +1,95 @@
...
+        PUPPETSERVER_CA_DIR = '/etc/puppetlabs/puppetserver/ca'
+
+        SUMMARY = "Migrate the existing CA directory to /etc/puppetlabs/puppetserver/ca"
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca migrate [--help]
+  puppetserver ca migrate [--config PATH]
+
+Description:
+  Migrate an existing CA directory at /etc/puppetlabs/puppetserver/ca. This is for
```
mwaggett, line 19: This implies to me that the existing CA directory is at that path. I think this is clearer:
```suggestion
  Migrate an existing CA directory to /etc/puppetlabs/puppetserver/ca. This is for
```

lib/puppetserver/ca/action/migrate.rb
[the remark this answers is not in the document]
mwaggett: Should we call out the Puppet 7 case? Is it if they're using a fresh Puppet 7 install? Or will upgrading to Puppet 7 do the migration for them? I assume the former..

mwaggett (approved):
Wanna squash?

tvpartytonight:
Squashed.

PR #65, opened 2020-11-02
(SERVER-2895) Symlink to old cadir when setting up with the new default

mwaggett:
This commit adds a method to symlink the new default cadir to the old default
cadir location when running the `setup` and `import` commands. This ensures
that, even on new installs, we can access the cadir from the old location
because there are still tools in the ecosystem that expect the old location.

### DIFF
--- a/lib/puppetserver/ca/utils/config.rb
+++ b/lib/puppetserver/ca/utils/config.rb
@@ -1,3 +1,5 @@
+require 'puppetserver/ca/utils/file_system'
+
 module Puppetserver
   module Ca
     module Utils
@@ -17,6 +19,40 @@ module Puppetserver
               name
             end
           end.sort.uniq.join(", ")
+        end
+
+        def self.puppet_confdir
+          if running_as_root?
+            '/etc/puppetlabs/puppet'
+          else
+            "#{ENV['HOME']}/.puppetlabs/etc/puppet"
+          end
+        end
+
+        def self.puppetserver_confdir(puppet_confdir)
+          File.join(File.dirname(puppet_confdir), 'puppetserver')
+        end
+
+        def self.default_ssldir(confdir = puppet_confdir)
+          File.join(confdir, 'ssl')
+        end
+
+        def self.old_default_cadir(confdir = puppet_confdir)
+          File.join(confdir, 'ssl', 'ca')
+        end
+
+        def self.new_default_cadir(confdir = puppet_confdir)
+          File.join(puppetserver_confdir(confdir), 'ca')
+        end
+
+        def self.symlink_to_old_cadir(current_cadir, puppet_confdir)
+          old_cadir = old_default_cadir(puppet_confdir)
+          new_cadir = new_default_cadir(puppet_confdir)
+          return if current_cadir != new_cadir
+          # This is only run on setup/import, so there should be no files in the
+          # old cadir, so it should be safe to forcibly remove it (which we need
+          # to do in order to create a symlink).
+          Puppetserver::Ca::Utils::FileSystem.forcibly_symlink(new_cadir, old_cadir)
         end
 
       end
--- a/lib/puppetserver/ca/utils/file_system.rb
+++ b/lib/puppetserver/ca/utils/file_system.rb
@@ -50,6 +50,11 @@ module Puppetserver
           errors
         end
 
+        def self.forcibly_symlink(source, link_target)
+          FileUtils.remove_dir(link_target, true)
+          FileUtils.symlink(source, link_target)
+        end
+
         def initialize
           @user, @group = find_user_and_group
         end
--- a/spec/utils/ssl.rb
+++ b/spec/utils/ssl.rb
@@ -1,5 +1,6 @@
 require 'openssl'
 require 'fileutils'
+require 'puppetserver/ca/utils/config'
 
 module Utils
   module SSL
@@ -109,6 +110,56 @@ module Utils
         [master]
           cadir = #{ca_dir}
           ssldir = #{ssl_dir}
+          keylength = 512
+        CONF
+      end
+
+      not_before = Time.now - 1
+
+      root_key = OpenSSL::PKey::RSA.new(512)
+      root_cert = create_cert(root_key, 'foo')
+
+      leaf_key = OpenSSL::PKey::RSA.new(512)
+      File.open(key_file, 'w') do |f|
+        f.puts leaf_key.to_pem
+      end
+
+      leaf_cert = create_cert(leaf_key, 'bar', root_key, root_cert)
+
+      File.open(bundle_file, 'w') do |f|
+        f.puts leaf_cert.to_pem
+        f.puts root_cert.to_pem
+      end
+
+      root_crl = create_crl(root_cert, root_key)
+      leaf_crl = create_crl(leaf_cert, leaf_key)
+
+      File.open(chain_file, 'w') do |f|
+        f.puts leaf_crl.to_pem
+        f.puts root_crl.to_pem
+      end
+
+
+      block.call(bundle_file, key_file, chain_file, config_file)
+    end
+
+    def with_files_in_default_dirs(tmpdir, &block)
+      fixtures_dir = File.join(tmpdir, 'fixtures')
+      confdir = File.join(tmpdir, 'puppet')
+
+      FileUtils.mkdir_p fixtures_dir
+      FileUtils.mkdir_p Puppetserver::Ca::Utils::Config.default_ssldir(confdir)
+      FileUtils.mkdir_p Puppetserver::Ca::Utils::Config.new_default_cadir(confdir)
+
+      bundle_file = File.join(fixtures_dir, 'bundle.pem')
+      key_file = File.join(fixtures_dir, 'key.pem')
+      chain_file = File.join(fixtures_dir, 'chain.pem')
+      config_file = File.join(fixtures_dir, 'puppet.conf')
+
+      File.open(config_file, 'w') do |f|
+        f.puts <<-CONF
+        [master]
+          confdir = #{confdir}
           keylength = 512
         CONF
       end
--- a/README.md
+++ b/README.md
@@ -73,8 +73,26 @@ interactive prompt that will allow you to experiment.
 
 To install this gem onto your local machine, run `bundle exec rake install`.
 
-To release a new version, update the version number in `version.rb`, and then
-speak with Release Engineering.
+### Testing
+To test your changes on a VM:
+1. Build the gem with your changes: `gem build puppetserver-ca.gemspec`
+1. Copy the gem to your VM: `scp puppetserver-ca-<version>.gem <your-vm>:.`
+1. Install puppetserver (FOSS) by installing the relevant release package and then installing the puppetserver package. For example:
+    ```
+    $ wget http://nightlies.puppet.com/yum/puppet6-nightly-release-el-7.noarch.rpm
+    $ rpm -i puppet6-nightly-release-el-7.noarch.rpm
+    $ yum update
+    $ yum install -y puppetserver
+    ```
+1. Restart your shell so that puppet's bin dir is on your $PATH: `exec bash`
+1. Install the gem into puppet's gem directory using puppet's gem command:
+    ```
+    $ /opt/puppetlabs/puppet/bin/gem install --install-dir "/opt/puppetlabs/puppet/lib/ruby/vendor_gems" puppetserver-ca-<version>.gem
+    ```
+1. To confirm that installation was successful, run `puppetserver ca --help`
+
+### Releasing
+To release a new version, run the [release pipeline](https://jenkins-master-prod-1.delivery.puppetlabs.net/job/platform_puppetserver-ca_init-multijob_main/), which will bump the version, tag, build, and release the gem.
 
 
 ## Contributing & Support
--- a/lib/puppetserver/ca/action/import.rb
+++ b/lib/puppetserver/ca/action/import.rb
@@ -4,6 +4,7 @@ require 'puppetserver/ca/config/puppet'
 require 'puppetserver/ca/errors'
 require 'puppetserver/ca/local_certificate_authority'
 require 'puppetserver/ca/utils/cli_parsing'
+require 'puppetserver/ca/utils/config'
 require 'puppetserver/ca/utils/file_system'
 require 'puppetserver/ca/utils/signing_digest'
 require 'puppetserver/ca/x509_loader'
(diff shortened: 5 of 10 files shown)


lib/puppetserver/ca/utils/file_system.rb
```diff
@@ -50,6 +50,17 @@ def self.check_for_existing_files(one_or_more_paths)
           errors
         end
 
+        def self.symlink_to_old_cadir(current_cadir)
```
mwaggett, line 53: lmk if there's a better shared place to put this, nothing seemed like an obviously good place

lib/puppetserver/ca/utils/file_system.rb
```diff
@@ -50,6 +50,17 @@ def self.check_for_existing_files(one_or_more_paths)
...
+          old_cadir = '/etc/puppetlabs/puppet/ssl/ca'
+          new_cadir = '/etc/puppetlabs/puppetserver/ca'
```
mwaggett, line 55: Should I use the same `running_as_root?` branching that Justin had in his PR? That might make testing possible.. But not sure if we really only wanna do this linking if we're using that^ specific default.

mwaggett:
I did a manual test of `puppetserver ca setup` on a FOSS install and it created the cadir in the right place and symlinked to the old place.

lib/puppetserver/ca/utils/file_system.rb
justinstoller: imo, having running from source be as similar to running from packages is something we should shoot for.

lib/puppetserver/ca/utils/file_system.rb
```diff
@@ -50,6 +50,17 @@ def self.check_for_existing_files(one_or_more_paths)
...
+          return if current_cadir != new_cadir
+          # Files should have already been moved if we're using the new cadir
```
justinstoller, line 57: Files won't and shouldn't exist during this.

lib/puppetserver/ca/utils/file_system.rb
mwaggett: I'm not sure what you mean. Are you saying I should or should not do the same branching as you did?

lib/puppetserver/ca/utils/file_system.rb
mwaggett: ah yeah, I can make this comment more accurate

lib/puppetserver/ca/utils/file_system.rb
justinstoller: I really want to make this just a forceful symlink function and if it really needs to hardcode the ca defaults then move it to the `local_certificate_authority` but as is matches pretty well with how this module is structured. It _might_ be time to refactor it (the local ca class didn't exist when we made the file_system module) but I'm okay waiting a bit before doing a refactor.

lib/puppetserver/ca/utils/file_system.rb
justinstoller: Yeah, I'd prefer we do the branching.

lib/puppetserver/ca/utils/file_system.rb
mwaggett: I could make this more generic, but I think we'd still need another new method wrapped around it because we need to check whether we're using the new default or not before trying to link. Do you think that's worthwhile?

lib/puppetserver/ca/utils/file_system.rb
Magisus: I think that maybe means we don't need the `rm` right? Since this function is called during initial setup, and will be using the new location?

mwaggett: yeah, perhaps.. but seeing how much fun we've had with symlinks in the past, it seems safer to delete it just in case.

Magisus: Heh fair enough. Seems alright I guess.

lib/puppetserver/ca/utils/file_system.rb
```diff
@@ -50,6 +50,17 @@ def self.check_for_existing_files(one_or_more_paths)
...
+          # so it should be safe to forcibly remove it (which we need to do
+          # in order to create a symlink).
+          FileUtils.remove_dir(old_cadir, true)
+          FileUtils.symlink(new_cadir, old_cadir)
```
tvpartytonight, line 61: These two calls can be combined into a single one using the [force](https://ruby-doc.org/stdlib-2.5.0/libdoc/fileutils/rdoc/FileUtils.html#method-i-ln_s) option.

```suggestion
FileUtils.symlink(new_cadir, old_cadir, force: true)
```

mwaggett: I don't think that will do what we want. If the target is a directory, the symlink will be nested under it. So we'd end up with `$old_cadir/$new_cadir -> $new_cadir`. I just tried this out in irb and it seems consistent with [the docs](https://ruby-doc.org/stdlib-2.7.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-ln_s), but let me know if I've missed something.

tvpartytonight: Oh blerg, I misread that doc, good thing you tested it!

spec/utils/ssl.rb
```diff
@@ -142,6 +142,59 @@ def with_files_in(tmpdir, &block)
...
     end
 
+    def with_files_in_default_dirs(tmpdir, &block)
+      fixtures_dir = File.join(tmpdir, 'fixtures')
+      default_ssldir = File.join(ENV['HOME'], '.puppetlabs', 'etc', 'puppet', 'ssl')
+      default_cadir = File.join(ENV['HOME'], '.puppetlabs', 'etc', 'puppetserver', 'ca')
+
+      FileUtils.mkdir_p fixtures_dir
+      # Clean up anything already here
+      FileUtils.remove_dir(default_ssldir, true)
```
Magisus, line 152: Soo I'm actually not sure how I feel about this. This will clean up anything I have in my normal dev dir as a side effect of running these tests, which I feel could be unexpected and possibly destructive. But if others think we should treat that dev dir as unreliable (i.e. anything could mess with it at any time), I'll defer to that.

Magisus: We could potentially back up what's there and put it back at the end?

lib/puppetserver/ca/local_certificate_authority.rb
```diff
@@ -37,6 +37,38 @@ class LocalCertificateAuthority
...
+        end
+      end
+
+      def self.puppetserver_confdir
+        if Puppetserver::Ca::Utils::FileSystem.running_as_root?
+          '/etc/puppetlabs/puppetserver'
+        else
+          "#{ENV['HOME']}/.puppetlabs/etc/puppetserver"
+        end
+      end
```
Magisus, line 54: Can we utilize these where Justin had similar logic?

Magisus: Like somehow consolidate with https://github.com/puppetlabs/puppetserver-ca-cli/blob/main/lib/puppetserver/ca/config/puppet.rb#L53-L74

mwaggett: yeah, I was thinking about that. Wasn't sure how much we wanted to keep the config stuff as its own thing? but I can update.

Magisus: Could possibly put these in with the filesystem stuff and call them both places? Not totally sure what the right separation of concerns is, but the logic is finicky and hardcoded and I'd rather that only be in one place.

justinstoller: fwiw, I think these methods are available on the "puppet" config instance here: https://github.com/puppetlabs/puppetserver-ca-cli/pull/65/files#diff-8a63d3dddb93d86b02aba1619d4d8e877d75584395f677c0dbbce246e40d0b81R58

Passing either the whole config instance (since the settings are also a member of it) or gather these in the caller would probably work?

mwaggett: ugh diff links on PRs like that never work for me

mwaggett: but I think I know where you're talking about

Magisus: They don't work for me either.

mwaggett: hmm those puppet config methods are _not_ available in the tests.. Does it make sense to create an instance just to access those methods?

mwaggett: okay I just updated Justin's code to use these methods.

spec/utils/ssl.rb
mwaggett: @justinstoller @tvpartytonight Do y'all have opinions on this?

lib/puppetserver/ca/local_certificate_authority.rb
Magisus: How would you feel about defining these in https://github.com/puppetlabs/puppetserver-ca-cli/blob/main/lib/puppetserver/ca/utils/config.rb as static functions? That feels like a better home for them than in this class. I don't _really_ like requiring this file in the config class (a little bit worried about winding up in a circular dependency by accident).

mwaggett: sure

mwaggett: done

lib/puppetserver/ca/utils/config.rb
```diff
@@ -19,6 +19,42 @@ def self.munge_alt_names(names)
...
+        def self.new_default_cadir
+          File.join(self.puppetserver_confdir, 'ca')
+        end
+
+        def self.symlink_to_old_cadir(current_cadir)
+          return if current_cadir != new_default_cadir
+          # This is only run on setup/import, so there should be no files in the
+          # old cadir, so it should be safe to forcibly remove it (which we need
+          # to do in order to create a symlink).
+          Puppetserver::Ca::Utils::FileSystem.forcibly_symlink(new_default_cadir, old_default_cadir)
```
Magisus, line 55: Should add the require for this file, to avoid any order-dependent issues.

lib/puppetserver/ca/local_certificate_authority.rb
Magisus: Thanks, I think that feels more right.

spec/utils/ssl.rb
tvpartytonight: I feel more ok about modifying the `puppetserver/ca` directory, and less ok about deleting the `puppet/ssl` directory. I think that it is quite reasonable for this repo to modify the contents of `puppetserver/ca`, but `puppet/ssl` seems more of an unexpected side effect.

Not sure if this applies here, but could the structure of this be changed to allow for the tests to modify different locations, while the actual production code makes that impossible? For instance, the actual [migrate](https://github.com/puppetlabs/puppetserver-ca-cli/blob/main/lib/puppetserver/ca/action/migrate.rb#L57) call in `run` only ever calls it with 1 argument, but the actual method [definition](https://github.com/puppetlabs/puppetserver-ca-cli/blob/main/lib/puppetserver/ca/action/migrate.rb#L66-L69) has a default that can be overridden, the sole purpose for [easier testing](https://github.com/puppetlabs/puppetserver-ca-cli/blob/main/spec/puppetserver/ca/action/migrate_spec.rb#L85-L92).

lib/puppetserver/ca/utils/config.rb
mwaggett: ahh yes, thanks. so many moving parts haha

Magisus (approved):

Magisus:
Will merge after squash.

spec/utils/ssl.rb
mwaggett: hmm, I'm not sure if it applies either, since the way I'm testing this doesn't call `symlink_to_old_cadir` directly. But I'll think about this. And we can do Maggie's backup suggestion if I can't come up with something better.

mwaggett:
Did another manual test for good measure and it still looks good. 
Just gonna think on the testing stuff over night, probs.

justinstoller:
I think we should be setting the "default_{new,old}_cadir" to be relative to $confdir. That would let us test this easier and be consistent with the work we just landed.

mwaggett:
@justinstoller relative to the actual confdir, as opposed to the defaults? I agree that would be easier to test, but then my question is: if someone is using a non-default confdir, do we still want to create the symlink? My understanding was that the point of this symlink is so tools that hard-code the (default) cadir will continue to work. If we symlink to a cadir under a non-default confdir, won't those tools still be broken?

Magisus:
Wouldn't those tools already be broken if someone was using a non-default confdir then? Seems like that's not a case that actually happens in production (i.e. the confdir is expected to be default in that environment), so we can safely parameterize it without risking real use cases, but also allow easier testing.

justinstoller:
So my understanding of the use cases for this tool are that 99% of it is for bootstrapping a new install with 0 custom configuration. There exists the ability to do custom configuration and some advanced users may do so. I expect those that have changed any of the defaults to "know what they're doing" and not need a symlink. However, if someone has set a $confdir to a custom value, but are otherwise expecting $cadir to be rooted in the same place as usual our new move could be disruptive.

Basically, I don't think users who are setting custom values (anywhere) when they use this command _need_ this symlink. However, I also don't think it's _incorrect_ behavior to create the symlink in the case of a custom $confdir case. I do think it is _incorrect_ to link a custom $cadir to the old cadir.

Magisus (approved):

lib/puppetserver/ca/utils/config.rb
```diff
@@ -22,39 +22,40 @@ def self.munge_alt_names(names)
...
-            "#{ENV['HOME']}/.puppetlabs/etc/puppetserver"
-          end
+        def self.puppetserver_confdir(puppet_confdir)
+          File.join(File.dirname(puppet_confdir), 'puppetserver')
         end
 
-        def self.default_ssldir
-          File.join(self.puppet_confdir, 'ssl')
+        def self.default_ssldir(confdir = nil)
+          confdir ||= puppet_confdir
```
justinstoller, line 37: why not just make the default in the arg list be puppet_confdir?

mwaggett: yeeeah, I thought about that.. it seemed kinda weird to me to call methods in the argument list, but I can change it.

Magisus (approved):

PR #68, opened 2020-12-07
(SERVER-2931) Do not warn about old CA dir location when migrating

Magisus:
This commit adds an internal flag to disable the warning about using the
old CA dir location, so that it can be disabled when running the
`migrate` command.

### DIFF
--- a/lib/puppetserver/ca/config/puppet.rb
+++ b/lib/puppetserver/ca/config/puppet.rb
@@ -25,7 +25,7 @@ module Puppetserver
 
         def self.parse(config_path, logger)
           instance = new(config_path)
-          instance.load({}, logger)
+          instance.load(logger: logger)
 
           return instance
         end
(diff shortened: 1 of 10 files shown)


lib/puppetserver/ca/config/puppet.rb
```diff
@@ -212,7 +213,7 @@ def parse_text(text)
        private
 
 
-        def find_cadir(configured_cadir, confdir, ssldir, logger)
+        def find_cadir(configured_cadir, confdir, ssldir, logger, ca_dir_warn)
```
mwaggett, line 216: This doesn't do anything yet..

Magisus:
We don't have this warning on the version in the 6.x branch, so nothing to do there.

lib/puppetserver/ca/config/puppet.rb
Magisus: Yeah I was still testing this, sorry.

Magisus:
It works now. Tested in a package build.

mwaggett (approved):

mwaggett:
> We don't have this warning on the version in the 6.x branch, so nothing to do there.

I'm a little confused what you mean by this..

Magisus:
There's no warning on config load about the CA dir being wrong in the 1.x stream. We have the migration command, but none of the rest of it. So it's safe to target this at `main`.

mwaggett:
ahhh gotcha, didn't even notice that this was targeting `main` 😅

PR #69, opened 2020-12-09
(SERVER-2917) Ensure old cadir symlink has same perms as actual cadir

mwaggett:
see individual commit messages for details

### DIFF
--- a/lib/puppetserver/ca/action/migrate.rb
+++ b/lib/puppetserver/ca/action/migrate.rb
@@ -66,6 +66,14 @@ SUCCESS_MESSAGE
         def migrate(old_cadir, new_cadir=PUPPETSERVER_CA_DIR)
           FileUtils.mv(old_cadir, new_cadir)
           FileUtils.symlink(new_cadir, old_cadir)
+          # Ensure the symlink has the same ownership as the actual cadir.
+          # This requires using `FileUtils.chown` rather than `File.chown`, as
+          # the latter will update the ownership of the target rather than the
+          # link itself.
+          # Symlink permissions are ignored in favor of the target's permissions,
+          # so we don't have to change those.
+          cadir = File.stat(new_cadir)
+          FileUtils.chown(cadir.uid, cadir.gid, old_cadir)
         end
 
         def parse(args)
--- a/lib/puppetserver/ca/config/puppet.rb
+++ b/lib/puppetserver/ca/config/puppet.rb
@@ -66,10 +66,13 @@ module Puppetserver
 
           results ||= {}
           results[:main] ||= {}
+          # The [master] config section is deprecated
+          # We now favor [server], but support both for backwards compatibility
           results[:master] ||= {}
+          results[:server] ||= {}
           results[:agent] ||= {}
 
-          overrides = results[:agent].merge(results[:main]).merge(results[:master])
+          overrides = results[:agent].merge(results[:main]).merge(results[:master]).merge(results[:server])
           overrides.merge!(cli_overrides)
 
           @settings = resolve_settings(overrides).freeze
--- a/lib/puppetserver/ca/utils/http_client.rb
+++ b/lib/puppetserver/ca/utils/http_client.rb
@@ -174,7 +174,7 @@ module Puppetserver
             self.new(settings, with_client_cert: false).with_connection(status_url) do |conn|
               result = conn.get
               if result.body == "running"
-                logger.err "CA service is running. Please stop it before attempting to run this command."
+                logger.err "Puppetserver service is running. Please stop it before attempting to run this command."
                 true
               else
                 false
--- a/spec/puppetserver/ca/action/generate_spec.rb
+++ b/spec/puppetserver/ca/action/generate_spec.rb
@@ -405,7 +405,7 @@ RSpec.describe Puppetserver::Ca::Action::Generate do
                               'subject-alt-names' => '',
                               'ca-client' => true})
           expect(code).to eq(1)
-          expect(stderr.string).to include("CA service is running")
+          expect(stderr.string).to include("server service is running")
         end
       end
     end


spec/puppetserver/ca/action/migrate_spec.rb
```diff
@@ -87,6 +87,11 @@
       expect(File.directory?(new_cadir)).to eq(true)
       expect(File.symlink?(old_cadir)).to eq(true)
       expect(File.readlink(old_cadir)).to eq(new_cadir)
+      old = File.stat(old_cadir)
+      new = File.stat(new_cadir)
+      expect(old.uid).to eq(new.uid)
+      expect(old.gid).to eq(new.gid)
+      expect(old.mode).to eq(new.mode)
```
mwaggett, line 94: I think we should not actually add these tests, but wanted to see what other
people thought. Two reasons: 1. these tests pass without code changes, since
we're running in a test environment and the same process is setting up all these
dirs and running the tests. 2. the test basically just copies the same method
calls used in the source, so it doesn't seem especially valuable.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -66,6 +66,10 @@ def run(input)
         def migrate(old_cadir, new_cadir=PUPPETSERVER_CA_DIR)
           FileUtils.mv(old_cadir, new_cadir)
           FileUtils.symlink(new_cadir, old_cadir)
+          # Ensure the symlink has the same ownership and permissions as the actual cadir
+          cadir = File.stat(new_cadir)
+          File.chown(cadir.uid, cadir.gid, old_cadir)
+          File.chmod(cadir.mode, old_cadir)
```
mwaggett, line 72: `chmod` doesn't actually work on symlinks. From the man page:

> chmod never changes the permissions of symbolic links; the chmod system call cannot change their permissions. This is not a problem since the permissions of symbolic links are never used.  However, for each symbolic link listed on the command line, chmod changes the permissions of the pointed-to file.  In contrast, chmod ignores symbolic links encountered during recursive directory traversals.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -66,6 +66,10 @@ def run(input)
         def migrate(old_cadir, new_cadir=PUPPETSERVER_CA_DIR)
           FileUtils.mv(old_cadir, new_cadir)
           FileUtils.symlink(new_cadir, old_cadir)
+          # Ensure the symlink has the same ownership and permissions as the actual cadir
+          cadir = File.stat(new_cadir)
+          File.chown(cadir.uid, cadir.gid, old_cadir)
```
mwaggett, line 71: `chown` needs the `-h` flag to operate on symlinks. From the man page:
> Symbolic links named by arguments are silently left unchanged unless -h is used.

I'm not sure how to do this in Ruby, other than by shelling out.

spec/puppetserver/ca/action/migrate_spec.rb
Magisus: Agreed, seems kinda pointless for the reasons you mention.

lib/puppetserver/ca/action/migrate.rb
Magisus: Hm, yeah, doesn't really look like that's supported... Thoughts on shelling out here? Doesn't seem like the worst thing.

lib/puppetserver/ca/action/migrate.rb
Magisus: So we're thinking we can just leave this out then? Makes sense to me.

lib/puppetserver/ca/action/migrate.rb
mwaggett: I think shelling out is probably fine, just wanted to check in with others to see if anyone happened to be aware of another way.

lib/puppetserver/ca/action/migrate.rb
mwaggett: yeah, I'm gonna remove this.

mwaggett:
I tested this on a centos 7 VM by installing FOSS puppetserver, my version of this gem, running `puppetserver ca setup`, and `puppetserver ca migrate`:
```
[root@lyric-testament ~]# ls -l /etc/puppetlabs/puppet/ssl/
total 4
lrwxrwxrwx 1 puppet puppet   31 Dec  9 18:43 ca -> /etc/puppetlabs/puppetserver/ca
```

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -66,6 +66,13 @@ def run(input)
         def migrate(old_cadir, new_cadir=PUPPETSERVER_CA_DIR)
           FileUtils.mv(old_cadir, new_cadir)
           FileUtils.symlink(new_cadir, old_cadir)
+          # Ensure the symlink has the same ownership as the actual cadir.
+          # The `-h`, `--no-dereference` flag is required to affect symlinks,
+          # which is not available in Ruby, so we have to shell out.
+          # Symlink permissions are ignored in favor of the target's permissions,
+          # so we don't have to change those.
+          cadir = File.stat(new_cadir)
+          `chown -h #{cadir.uid}:#{cadir.gid} #{old_cadir}`
```
donoghuc, line 75: Sorry for a drive by review, but this tab was still open from statndup.... I think if you are going to shell out you should almost certainly use open3 for it. https://docs.ruby-lang.org/en/2.0.0/Open3.html

mwaggett: @donoghuc how come? I don't think we need any output..

though now I am remembering @justinstoller suggested putting this in a `try` block, so I can update that..

donoghuc: hm, i figured you would want the status at least. Looks like system or open3 are preferred https://medium.com/zendesk-engineering/running-a-child-process-in-ruby-properly-febd0a2b6ec8

mwaggett: @justinstoller Did you have something in mind for what we should do if the `chown` fails? I'm thinking just warn and do nothing?? ¯\_(ツ)_/¯

Magisus: It was my understanding too last time I looked into this (admittedly a long time ago) that backticks and `sh` are generally to be avoided.

mwaggett: okay, I can change it. I didn't see anything with a cursory google search, but I didn't look very hard.

lib/puppetserver/ca/action/migrate.rb
donoghuc: I'm still not sure why `FileUtils#chown` does not work? I see you tried with `File#chown`. 
```
root@cas-ThinkPad-T460p:/tmp/test# ls -pla
total 76
drwxrwxr-x   2 cas  cas   4096 Dec  9 11:09 ./
drwxrwxrwt 555 root root 69632 Dec  9 11:11 ../
lrwxrwxrwx   1 cas  cas      3 Dec  9 11:09 bar -> foo
-rw-rw-r--   1 cas  cas      0 Dec  9 11:09 foo
root@cas-ThinkPad-T460p:/tmp/test# irb
irb(main):001:0> require 'fileutils'
=> true
irb(main):002:0> FileUtils.chown('root','root','/tmp/test/bar')
=> ["/tmp/test/bar"]
irb(main):003:0> 
root@cas-ThinkPad-T460p:/tmp/test# ls -pla
total 76
drwxrwxr-x   2 cas  cas   4096 Dec  9 11:09 ./
drwxrwxrwt 555 root root 69632 Dec  9 11:12 ../
lrwxrwxrwx   1 root root     3 Dec  9 11:09 bar -> foo
-rw-rw-r--   1 cas  cas      0 Dec  9 11:09 foo
```

lib/puppetserver/ca/action/migrate.rb
donoghuc: https://github.com/puppetlabs/puppetserver-ca-cli/pull/69#discussion_r539576034

lib/puppetserver/ca/action/migrate.rb
mwaggett: huh interesting

justinstoller: Damn, you Cas. I just came here to say that.

mwaggett: super weird to me that `chown` exists under both `File` and `FileUtils` and that it would be have differently for each, but.. ¯\_(ツ)_/¯ I'll take it.

Magisus (approved):

mwaggett:
okay this works

PR #70, opened 2020-12-09
(SERVER-2917) Ensure old cadir symlink has same perms as actual cadir

mwaggett:
see individual commit messages for details

### DIFF
--- a/lib/puppetserver/ca/action/migrate.rb
+++ b/lib/puppetserver/ca/action/migrate.rb
@@ -66,15 +66,7 @@ SUCCESS_MESSAGE
 
         def migrate(old_cadir, new_cadir=PUPPETSERVER_CA_DIR)
           FileUtils.mv(old_cadir, new_cadir)
-          FileUtils.symlink(new_cadir, old_cadir)
-          # Ensure the symlink has the same ownership as the actual cadir.
-          # This requires using `FileUtils.chown` rather than `File.chown`, as
-          # the latter will update the ownership of the target rather than the
-          # link itself.
-          # Symlink permissions are ignored in favor of the target's permissions,
-          # so we don't have to change those.
-          cadir = File.stat(new_cadir)
-          FileUtils.chown(cadir.uid, cadir.gid, old_cadir)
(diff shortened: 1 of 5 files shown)


mwaggett:
tested the `setup` action on a vm

Magisus (approved):

lib/puppetserver/ca/utils/config.rb
```diff
@@ -53,6 +53,14 @@ def self.symlink_to_old_cadir(current_cadir, puppet_confdir)
...
           # to do in order to create a symlink).
           Puppetserver::Ca::Utils::FileSystem.forcibly_symlink(new_cadir, old_cadir)
+          # Ensure the symlink has the same ownership as the actual cadir.
+          # This requires using `FileUtils.chown` rather than `File.chown`, as
+          # the latter will update the ownership of the target rather than the
+          # link itself.
+          # Symlink permissions are ignored in favor of the target's permissions,
+          # so we don't have to change those.
+          cadir = File.stat(new_cadir)
+          FileUtils.chown(cadir.uid, cadir.gid, old_cadir)
```
Magisus, line 63: Do you think it might make sense to either include these in `forcibly_symlink` or create a separate function for this perm handling that can be used here and in `setup`?

mwaggett: You mean here and in `migrate`? This function is used in both `setup` and `import`.

I thought about reworking `migrate` to make use of this shared function, but decided it probably wasn't worth it. Keeping it the way it is makes mergeups easier too. 

I could include this in `forcibly_symlink`, but wasn't totally sure if this was something we'd always want when symlinking.

justinstoller: I'd be into adding this logic to `forcibly_symlink` and calling it in `migrate`.

Magisus (approved):

lib/puppetserver/ca/utils/config.rb
mwaggett: okay done

PR #71, opened 2021-03-04
(maint) Clarify that --config is optional to migrate

justinstoller:


justinstoller:
Do not merge. Pending feedback from community.

lib/puppetserver/ca/action/migrate.rb
```diff
@@ -18,8 +18,9 @@ class Migrate
 
 Description:
   Migrate an existing CA directory to #{PUPPETSERVER_CA_DIR}. This is for
-  upgrading from Puppet Platform 6.x to Puppet 7. Use the currently configured
-  puppet.conf file in your installation, or supply one using the `--config` flag.
+  upgrading from Puppet Platform 6.x to Puppet 7. Uses the default puppet.conf
+  in your installation, override by supplying the `--config` flag.
```
mwaggett, line 22: maybe, if we wanna be super duper explicit, we could say:
```suggestion
  in your installation, or use a different config by supplying the `--config` flag.
```
?

lib/puppetserver/ca/cli.rb
```diff
@@ -28,8 +28,8 @@ class Cli
       INIT_ACTIONS = {
         'import'   => Action::Import,
         'setup'    => Action::Setup,
-        'enable' => Action::Enable,
-        'migrate' => Action::Migrate,
+        'enable'   => Action::Enable,
+        'migrate'  => Action::Migrate,
```
mwaggett, line 32: I thought we decided we weren't going to do this kind of alignment? My IDE does it automatically, but I usually put it back, so I just wanna be clear on where we've landed with that.

lib/puppetserver/ca/cli.rb
```diff
@@ -121,6 +121,9 @@ def self.parse_general_inputs(inputs)
           opts.on('--version', 'Display the version') do |v|
             parsed['version'] = true
           end
+          opts.on('--config', 'Path to a custom puppet.conf') do |v|
```
mwaggett, line 124: Was the `--config` option just totally ignored before?? Is there something else we need to do to ensure the custom config is used?

justinstoller: ......... weird... I did not mean to include this in my change.

lib/puppetserver/ca/cli.rb
justinstoller: I don't think it's idiomatic to do it in clojure, I do think it's idiomatic in Ruby. I don't know that I'm the best at following that though.

lib/puppetserver/ca/cli.rb
mwaggett: so.. are you saying this shouldn't be here? How does passing `--config` work now?

justinstoller: Each action defines it.
At one point I thought that was dumb and started to do a maint to refactor it all into the cli class, but didn't finish it. Somehow when I checked out this branch some of that work came along?

mwaggett: ahh, gotcha. so it sounds like this should be removed.

lib/puppetserver/ca/cli.rb
mwaggett: ahh, okay. I don't really care, just wasn't sure what we had agreed on.

mwaggett:
Are you planning to update this @justinstoller ?

justinstoller:
Planning's a strong word, I at one point intended to.

I've removed the code that sneaked in besides the doc update and added your suggestion to the wording.

mwaggett (approved):

mwaggett:
not sure what's up with all those test failures..

Magisus:
That's still the Facter bug. It's been fixed, but this pulls in only released versions. I guess they haven't done a release yet.

Magisus (approved):

PR #72, opened 2021-03-08
(SERVER-2338) Ensure `ca-client` cert has certname as altname

Magisus:
We previously updated the certificate signing code in puppetserver
proper to always add the certname as a subject alt name when signing a
certificate. This commit updates the `generate` action of the CLI gem to
also ensure that the certname of the host the cert is being generated
for is added as an alt name. Because `generate` is used to create certs
for other hosts, we can't use the value of `certname` from settings in
the CA code itself, but rather need to use the value of `certname` that
is passed into the command.

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -296,7 +296,9 @@ BANNER
         end
 
         def process_alt_names(alt_names, certname)
-          return '' if alt_names.empty?
+          # It is recommended (and sometimes enforced) to always include
+          # the certname as a SAN, see RFC 2818 https://tools.ietf.org/html/rfc2818#section-3.1.
+          return "DNS:#{certname}" if alt_names.empty?
 
           current_alt_names = alt_names.dup
           # When validating the cert, OpenSSL will ignore the CN field if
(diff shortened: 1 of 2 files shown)


lib/puppetserver/ca/action/generate.rb
```diff
@@ -296,7 +296,9 @@ def check_for_existing_ssl_files(certname, settings)
         end
 
         def process_alt_names(alt_names, certname)
-          return '' if alt_names.empty?
+          # It is recommended (and sometimes enforced) to always include
+          # the certname as a SAN, see RFC 2818 https://tools.ietf.org/html/rfc2818#section-3.1.
+          return "DNS:#{certname}" if alt_names.empty?
```
mwaggett, line 301: What if there are other alt names? Wouldn't we still want to add the certname?

Magisus: We do. See below.

Magisus: Every other case I could find, we were already doing it.

mwaggett: ahh, gotcha, didn't see the rest of this method 😅 
Do we have tests for those cases? I didn't see one near the test you added, so might be good to add.

Magisus: We test most of these already from several angles. I added one for specifically the line of code below this.

mwaggett (approved):

PR #74, opened 2021-03-24
(maint) Fix bad memoization in default certname

Magisus:
The `default_certname` method would do the right thing the first time it
was called, and then afterwards fail to get the hostname. This commit
fixes the bad short circuiting logic.

### DIFF
--- a/lib/puppetserver/ca/config/puppet.rb
+++ b/lib/puppetserver/ca/config/puppet.rb
@@ -79,15 +79,14 @@ module Puppetserver
         end
 
         def default_certname
-          @certname ||=
-            hostname = Facter.value(:hostname)
-            domain = Facter.value(:domain)
-            if domain and domain != ''
-              fqdn = [hostname, domain].join('.')
-            else
-              fqdn = hostname
-            end
-            fqdn.chomp('.')
+          hostname = Facter.value(:hostname)
+          domain = Facter.value(:domain)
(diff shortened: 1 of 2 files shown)


mwaggett (approved):

Magisus:
The travis failures are caused by this [commit](https://github.com/puppetlabs/facter/pull/2315) and are not fixed when I roll forward to latest (what this PR is currently pinned to). 

@puppetlabs/night-s-watch do you see what might be going on here?

The tests are failing because `fqdn` is nil [here](https://github.com/puppetlabs/puppetserver-ca-cli/blob/8349c2af841bd394434d9428e06926e7072c5c47/lib/puppetserver/ca/config/puppet.rb#L85).

Magisus:
These tests all pass for me locally, I don't feel like we need to block merge on getting an updated Facter (given we need a release in order to pull it in here).

mwaggett:
@Magisus will our other pipelines pass if we merge this?

Magisus:
The tests on this repo are failing before this change. Our pipelines get Facter in a totally different way than these unit tests, and are passing now. My guess is in that context they have ffi available and are therefore fine.

Magisus:
We have a floating dependency on facter so these tests will fix themselves the next time they release.

mwaggett (approved):

PR #75, opened 2021-04-10
(SERVER-2993) Detect migrated CA dir when calcuating settings

Magisus:
This commit updates the settings resolution logic to detect when the CA
dir has been migrated, to return the new location in that case. If the
new dir does not have CA content, the old location is returned
(including if no dir exists).

### DIFF
--- a/spec/puppetserver/ca/config/puppet_spec.rb
+++ b/spec/puppetserver/ca/config/puppet_spec.rb
@@ -1,4 +1,6 @@
 require 'spec_helper'
+require 'tmpdir'
+
 require 'puppetserver/ca/config/puppet'
 
 RSpec.describe 'Puppetserver::Ca::Config::Puppet' do
@@ -69,7 +71,7 @@ RSpec.describe 'Puppetserver::Ca::Config::Puppet' do
       conf.load
 
       expect(conf.errors).to be_empty
-      expect(conf.settings[:cacert]).to eq('/foo/bar/ca/ca_crt.pem')
+      expect(conf.settings[:localcacert]).to eq('/foo/bar/certs/ca.pem')
       expect(conf.settings[:cacrl]).to eq('/fizz/buzz/crl.pem')
       expect(conf.settings[:hostpubkey]).to eq('/agent/pubkeys/fooberry.pem')
     end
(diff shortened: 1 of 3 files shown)


Magisus:
fwiw in my manual testing, I also verified that this works fine for most operations even without this change (the symlink is sufficient). So if this doesn't make the release, that's fine IMO.

spec/puppetserver/ca/config/puppet_spec.rb
```diff
@@ -69,7 +71,7 @@
       conf.load
 
       expect(conf.errors).to be_empty
-      expect(conf.settings[:cacert]).to eq('/foo/bar/ca/ca_crt.pem')
+      expect(conf.settings[:localcacert]).to eq('/foo/bar/certs/ca.pem')
```
mwaggett, line 74: What's the reason for this change?

spec/puppetserver/ca/config/puppet_spec.rb
```diff
@@ -241,4 +243,73 @@
...
+
+        expect(File).to receive(:exist?)
+          .with("#{File.dirname(confdir)}/puppetserver/ca/ca_crt.pem")
+          .and_return(true)
+
+        conf = Puppetserver::Ca::Config::Puppet.new(puppet_conf)
+        settings = conf.load
+
+        expect(settings[:ssldir]).to eq(File.join(confdir, 'ssl'))
+        expect(settings[:cadir]).to eq(File.join(File.dirname(confdir), 'puppetserver/ca'))
```
mwaggett, line 290: might be worth putting `File.dirname(confdir)` into a variable, since it's used three times ¯\_(ツ)_/¯ I don't feel super strongly though.

mwaggett (approved):

mwaggett:
I'm guessing those Travis failures are still the Facter thing, but I really don't love having to just ignore test failures. We don't have to address it here, but it seems valuable to have a way to easily switch to dev versions of facter for testing purposes.

spec/puppetserver/ca/config/puppet_spec.rb
Magisus: Justin also made this change in his original PR for this on the other branch: the problem is that the location of `cacrt` becomes unpredictable with this change, without a bunch of mocking, but this is testing verifying that we pick up the customer `ssldir` location. So this change switches to testing a setting that still interpolates that.

Magisus:
We might be able to override it in the gemfile to point at Git? Fwiw these tests only fail on Travis, they pass for me locally.

spec/puppetserver/ca/config/puppet_spec.rb
mwaggett: ahh gotcha, thanks

mwaggett (approved):

PR #76, opened 2021-04-20
(SERVER-2932) Update terminology in ca cli

pcarlisle:
Update internal terminology, replacing master with server and blacklist with
denylist. The only user facing change here is updating the cli to correctly
read `serverport` from `puppet.conf` as a newer name for `masterport`.

### DIFF
--- a/spec/puppetserver/ca/config/puppet_spec.rb
+++ b/spec/puppetserver/ca/config/puppet_spec.rb
@@ -338,4 +338,41 @@ RSpec.describe 'Puppetserver::Ca::Config::Puppet' do
       expect(conf.default_certname).to eq("foo")
     end
   end
+
+  it 'loads masterport correctly' do
+    masterport = 1234
+    Dir.mktmpdir do |tmpdir|
+      puppet_conf = File.join(tmpdir, 'puppet.conf')
+      File.open puppet_conf, 'w' do |f|
+        f.puts(<<-INI)
+          [server]
+            masterport = #{masterport}
+        INI
+      end
+
+      conf = Puppetserver::Ca::Config::Puppet.new(puppet_conf)
+      settings = conf.load(logger: logger)
+
+      expect(settings[:ca_port]).to eq("#{masterport}")
+      expect(stderr.string).to be_empty
+    end
+  end
+  it 'loads serverport correctly' do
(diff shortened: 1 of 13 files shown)


spec/puppetserver/ca/config/puppet_spec.rb
```diff
@@ -338,4 +338,41 @@
...
+      end
+
+      conf = Puppetserver::Ca::Config::Puppet.new(puppet_conf)
+      settings = conf.load(logger: logger)
+
+      expect(settings[:ca_port]).to eq("#{masterport}")
+      expect(stderr.string).to be_empty
+    end
+  end
+  it 'loads serverport correctly' do
```
mwaggett, line 360: Would it be worth testing that `serverport` is preferred if both `masterport` and `serverport` are set? Or should we explicitly disallow that?

spec/puppetserver/ca/config/puppet_spec.rb
```diff
@@ -338,4 +338,41 @@
...
   end
+
+  it 'loads masterport correctly' do
+    masterport = 1234
+    Dir.mktmpdir do |tmpdir|
+      puppet_conf = File.join(tmpdir, 'puppet.conf')
+      File.open puppet_conf, 'w' do |f|
+        f.puts(<<-INI)
+          [server]
+            masterport = #{masterport}
```
mwaggett, line 349: (nit) Does it make sense to put this setting in the `master` section instead of `server`? I'm imagining that someone would probably either make all the server->master changes or none of them? But it probably doesn't really matter here, especially if we're already testing the section renames elsewhere.

pcarlisle: I don't think it matters, particularly because we just rename [master] to [server] internally when needed before any other processing happens.

spec/puppetserver/ca/config/puppet_spec.rb
pcarlisle: While we do prefer serverport, I didn't test this because it wasn't something that might fail.

pcarlisle:
i couldn't find the list either but i did update to blocklist

mwaggett (approved):

PR #78, opened 2021-06-23
(SERVER-2251) Add '--verbose' flag to 'puppetserve ca' cli tool

Tu2607:
There are more changes like adding the logger to certain methods in `http_client.rb` but I elected not to include that in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,11 @@ To create a new keypair and certificate for a certname:
 puppetserver ca generate --certname foo.example.com
 ```
 
+For enabling verbose mode:
+```
+puppetserver ca --verbose <action>
+```
+
 For more details, see the help output:
 ```
 puppetserver ca --help
@@ -68,7 +73,7 @@ for more details.
 ## Development
 
 After checking out the repo, run `bin/setup` to install dependencies. Then,
-run `rake spec` to run the tests. You can also run `bin/console` for an
+run `bundle exec rake spec` to run the tests. You can also run `bin/console` for an
 interactive prompt that will allow you to experiment.
 
 To install this gem onto your local machine, run `bundle exec rake install`.
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ To create a new keypair and certificate for a certname:
 puppetserver ca generate --certname foo.example.com
 ```
 
-For enabling verbose mode:
+To enable verbose mode:
 ```
 puppetserver ca --verbose <action>
 ```
--- a/lib/puppetserver/ca/cli.rb
+++ b/lib/puppetserver/ca/cli.rb
@@ -64,8 +64,9 @@ BANNER
 
 
       def self.run(cli_args = ARGV, out = STDOUT, err = STDERR)
-        logger = Puppetserver::Ca::Logger.new(:info, out, err)
         parser, general_options, unparsed = parse_general_inputs(cli_args)
+        logger = general_options.key?('verbose') ? Puppetserver::Ca::Logger.new(:debug, out, err)
+                                                    : Puppetserver::Ca::Logger.new(:info, out, err)
 
         if general_options['version']
           logger.inform Puppetserver::Ca::VERSION
@@ -120,6 +121,9 @@ BANNER
           end
           opts.on('--version', 'Display the version') do |v|
             parsed['version'] = true
+          end
+          opts.on('--verbose', 'Display low-level information') do |verbose|
+            parsed['verbose'] = true
           end
 
           opts.separator ACTION_OPTIONS
--- a/lib/puppetserver/ca/cli.rb
+++ b/lib/puppetserver/ca/cli.rb
@@ -65,8 +65,9 @@ BANNER
 
       def self.run(cli_args = ARGV, out = STDOUT, err = STDERR)
         parser, general_options, unparsed = parse_general_inputs(cli_args)
-        logger = general_options.key?('verbose') ? Puppetserver::Ca::Logger.new(:debug, out, err)
-                                                    : Puppetserver::Ca::Logger.new(:info, out, err)
+        level = general_options.delete('verbose') ? :debug : :info
+
+        logger = Puppetserver::Ca::Logger.new(level, out, err)
 
         if general_options['version']
           logger.inform Puppetserver::Ca::VERSION
--- a/lib/puppetserver/ca/logger.rb
+++ b/lib/puppetserver/ca/logger.rb
@@ -13,6 +13,10 @@ module Puppetserver
         @err = err
       end
 
+      def level
+        @level
+      end
+
       def debug(text)
         if @level >= LEVELS[:debug]
           @out.puts(text)
--- a/lib/puppetserver/ca/utils/http_client.rb
+++ b/lib/puppetserver/ca/utils/http_client.rb
@@ -96,23 +96,24 @@ module Puppetserver
             url = url_overide || @url
             headers = DEFAULT_HEADERS.merge(headers)
 
+            @logger.debug("Making a GET request at #{url}")
+
             request = Net::HTTP::Get.new(url.to_uri, headers)
             result = @conn.request(request)
-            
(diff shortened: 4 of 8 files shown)


README.md
```diff
@@ -55,6 +55,11 @@ To create a new keypair and certificate for a certname:
 puppetserver ca generate --certname foo.example.com
 ```
 
+For enabling verbose mode:
```
mwaggett, line 58: Just to be consistent with the wording we use for the other commands, I would suggest:
```suggestion
To enable verbose mode:
```

lib/puppetserver/ca/logger.rb
```diff
@@ -13,6 +13,10 @@ def initialize(level = :info, out = STDOUT, err = STDERR)
         @err = err
       end
 
+      def level
+        @level
+      end
+
```
mwaggett, line 19: Instead of defining this method, you could use an `attr_reader`, or `attr_accessor` if you think it might be useful to set this value in addition to getting it. This has more info about those: https://www.rubyguides.com/2018/11/attr_accessor/ or feel free to chat with any of us about it!

spec/shared_examples/cli_parsing.rb
```diff
@@ -17,4 +17,12 @@
     expect(stderr.string).to be_empty
     expect(first_code).to be 0
   end
+
+  it 'raise the verbose flag' do
+    args = ['--verbose', action].compact
+    _,parsed,_ = Puppetserver::Ca::Cli.parse_general_inputs(args)
+    expect(parsed['verbose']).to be true
+    logger = Puppetserver::Ca::Logger.new(:debug, stdout, stderr)
+    expect(logger.level).to be Puppetserver::Ca::Logger::LEVELS[:debug]
```
mwaggett, line 26: hmm, since you're creating a new `logger` here, I'm not sure that this is actually testing that the CLI inputs are correctly passed to the Logger class. Does the test still pass if you remove lines 22-24?

mwaggett (commented):
Great first PR! Just a few small things.. 😜 

you did it!

lib/puppetserver/ca/cli.rb
```diff
@@ -64,8 +64,9 @@ class Cli
 
 
       def self.run(cli_args = ARGV, out = STDOUT, err = STDERR)
-        logger = Puppetserver::Ca::Logger.new(:info, out, err)
         parser, general_options, unparsed = parse_general_inputs(cli_args)
+        logger = general_options.key?('verbose') ? Puppetserver::Ca::Logger.new(:debug, out, err)
```
justinstoller, line 68: Some style ideas:
1. Nothing else needs to read the `verbose` key out of these options so you might want to get the value while removing it. You can do that with `delete()`, eg `general_options.delete('verbose')`
2. I think it's cleaner to put the value of the log level in the conditional and then pass that into the Logger's constructor, eg
```rb
level = general_options.key?('verbose') ? :debug : :info
logger = Puppetserver::Ca::Logger.new(level, out, err)
```

lib/puppetserver/ca/logger.rb
justinstoller: agree with Molly, to contrast, using a getter like this is valuable if you need to do some kind of translation, like if you wanted to return the symbolic name of the level.

spec/shared_examples/cli_parsing.rb
Tu2607: It does. I just wanted to make sure that the level number is correct for the sake of it.

lib/puppetserver/ca/logger.rb
Tu2607: I did discuss that with Justin and I ended up going with a getter function. I could definitely make the change if it helps with readability.

spec/shared_examples/cli_parsing.rb
mwaggett: no yeah, I think that's fine, but I think you need something else to confirm that passing the `--verbose` flag ends up creating a Logger at the debug level. Does that make sense? Happy to chat more about it tomorrow.

lib/puppetserver/ca/cli.rb
Tu2607: Right on. I'll make these changes.

spec/shared_examples/cli_parsing.rb
Tu2607: Yep that crossed my mind but I did not know how to express that because the logger that will be created will be a local variable to the function `run`. And I wasn't sure how I can check that local variable.

lib/puppetserver/ca/logger.rb
justinstoller: `attr_reader` is more of a Ruby specific idiomatic thing

spec/shared_examples/cli_parsing.rb
mwaggett: yeah, I wasn't sure off the top of my head exactly how to test that haha, but happy to pair on it tomorrow if you want!

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -96,6 +98,8 @@ def get(url_overide = nil, headers = {})
 
             request = Net::HTTP::Get.new(url.to_uri, headers)
             result = @conn.request(request)
+            
+            @logger.debug("Making a get request at url #{url}")
```
Magisus, line 102: Can we move these to before we make the connection? And put the types in all caps? (e.g. `GET`) That's usually how HTTP request types are formatted.

Tu2607: Yeah. When you mean put the types in all caps, you mean like "Making a GET request..." right?

Magisus: Yes, and same with the others.

Tu2607: Your wish is my command.  It will be capitalized 😄

spec/puppetserver/ca/cli_spec.rb
```diff
@@ -57,4 +57,17 @@
       'sign',
       /.*Usage.* puppetserver ca sign.*Display this command-specific help output.*/m
   end
+
+  #This test is not specific for clean, but just an action to validate our test
+  #would work with other options as well.
```
mwaggett, line 62: (nit) we usually put a space between the `#` and the words
```suggestion
  # This test is not specific for clean, but just an action to validate our test
  # would work with other options as well.
```

spec/puppetserver/ca/cli_spec.rb
```diff
@@ -57,4 +57,17 @@
...
+  it 'raise the verbose flag' do
+    args = ['--verbose', 'clean'].compact
+    action_class = Puppetserver::Ca::Cli::VALID_ACTIONS[args[1]]
+    expect(action_class).to receive(:new).and_wrap_original  do |original, logger|
```
mwaggett, line 66: (nit) extra space
```suggestion
    expect(action_class).to receive(:new).and_wrap_original do |original, logger|
```

lib/puppetserver/ca/utils/http_client.rb
mwaggett: +1 to both of Maggie's suggestions

mwaggett (commented):
Have you done any manual testing?

lib/puppetserver/ca/logger.rb
mwaggett: were you gonna make the change to use `attr_reader`?

spec/puppetserver/ca/cli_spec.rb
Tu2607: Justin and I worked through that test yesterday. I manually tested by replacing clean with other action like revoke, list, generate, etc... I ended up picking clean just for a representation.  Thanks for spotting the extra space btw 😄

mwaggett (approved):
yaaaaaay!

PR #79, opened 2021-06-26
(SERVER-2251) Add '--verbose' flag to 'puppetserver ca' cli tool

Tu2607:
Reformat the output of the URL for the logger in verbose mode.

### DIFF
--- a/lib/puppetserver/ca/utils/http_client.rb
+++ b/lib/puppetserver/ca/utils/http_client.rb
@@ -96,7 +96,7 @@ module Puppetserver
             url = url_overide || @url
             headers = DEFAULT_HEADERS.merge(headers)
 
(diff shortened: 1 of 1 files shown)


Magisus:
Can you update the commit message to be more specific to this change? Something like "Print URL string instead of raw Ruby object"?

Tu2607:
> Can you update the commit message to be more specific to this change? Something like "Print URL string instead of raw Ruby object"? You can either keep the ticket number or tag it as "(maint)".

I tried 😢

mwaggett:
@Tu2607 you should be able to do a `git commit --amend` and then force push to your branch `git push -f origin HEAD`

Magisus (approved):

mwaggett (approved):

PR #80, opened 2021-06-29
(SERVER-3006) Provide parsable (json) output for puppetserver ca commands

Tu2607:
Here's a sample output in JSON format execute by `list --certname tu --format json | python -m json.tool` (Piped through python for pretty print)
```
[
    {
        "authorization_extensions": {},
        "dns_alt_names": [
            "DNS:tu"
        ],
        "fingerprint": "18:52:07:6E:6F:E0:42:40:20:75:B8:80:0F:5D:26:94:0A:22:DC:2A:DD:67:81:90:BA:B3:13:8D:AD:B3:B5:95",
        "fingerprints": {
            "SHA1": "74:6F:63:4B:94:0B:EE:EE:1F:D9:60:DD:36:9D:E8:76:6C:A0:38:F7",
            "SHA256": "18:52:07:6E:6F:E0:42:40:20:75:B8:80:0F:5D:26:94:0A:22:DC:2A:DD:67:81:90:BA:B3:13:8D:AD:B3:B5:95",
            "SHA512": "F7:0D:4E:C0:8C:A8:8A:8C:2B:5B:49:27:63:EB:39:C5:26:B7:DB:28:E8:7D:63:DD:E1:BF:E8:B6:06:A4:28:44:48:FD:81:89:EB:37:7E:54:12:9C:4D:1C:EC:63:63:A4:70:11:E8:08:F1:90:F5:B2:6C:99:CE:6A:03:C5:7A:4A",
            "default": "18:52:07:6E:6F:E0:42:40:20:75:B8:80:0F:5D:26:94:0A:22:DC:2A:DD:67:81:90:BA:B3:13:8D:AD:B3:B5:95"
        },
        "name": "tu",
        "not_after": "2026-06-28T19:18:10UTC",
        "not_before": "2021-06-28T19:18:10UTC",
        "serial_number": 3,
        "state": "revoked",
        "subject_alt_names": [
            "DNS:tu"
        ]
    }
]
```

### DIFF
--- a/lib/puppetserver/ca/action/list.rb
+++ b/lib/puppetserver/ca/action/list.rb
@@ -62,11 +62,11 @@ Options:
           certnames = input['certname'] || []
           all = input['all']
           output_format = input['format'] || "text"
-          
+
           unless VALID_FORMAT.include?(output_format)
             Errors.handle_with_usage(@logger, ["Unknown format flag '#{output_format}'. Valid formats are '#{VALID_FORMAT.join("', '")}'."])
             return 1
-          end 
+          end
 
           if all && certnames.any?
             Errors.handle_with_usage(@logger, ['Cannot combine use of --all and --certname.'])
@@ -81,23 +81,25 @@ Options:
           puppet = Config::Puppet.parse(config)
           return 1 if Errors.handle_with_usage(@logger, puppet.errors)
 
-          filter_names = certnames.any? \
-            ? lambda { |x| certnames.include?(x['name']) }
-            : lambda { |x| true }
+          if certnames.any?
+            filter_names = lambda { |x| certnames.include?(x['name']) }
+          else
+            filter_names = lambda { |x| true }
+          end
 
           all_certs = get_all_certs(puppet.settings).select { |cert| filter_names.call(cert) }
           requested, signed, revoked = separate_certs(all_certs)
           missing = certnames - all_certs.map { |cert| cert['name'] }
 
-          (all || certnames.any?)\
-          ? output_certs_by_state(all, output_format, requested, signed, revoked, missing)
-          : output_certs_by_state(all, output_format, requested)
+          if (all || certnames.any?)
+            output_certs_by_state(all, output_format, requested, signed, revoked, missing)
+          else
+            output_certs_by_state(all, output_format, requested)
+          end
 
-          return missing.any? \
-            ? 1
-            : 0
+          return missing.any? ? 1 : 0
         end
-        
+
         def output_certs_by_state(all, output_format, requested, signed = [], revoked = [], missing = [])
           if output_format == 'json'
             output_certs_json_format(all, requested, signed, revoked, missing)
@@ -115,25 +117,20 @@ Options:
                              "revoked" => revoked }.to_json
             @logger.inform(grouped_cert)
           else
-            unless requested.empty?
-              grouped_cert["requested"] = requested
-            end
+            grouped_cert["requested"] = requested unless requested.empty?
+            grouped_cert["signed"] = signed unless signed.empty?
+            grouped_cert["revoked"] = revoked unless revoked.empty?
+            grouped_cert["missing"] = missing unless missing.empty?
 
-            unless signed.empty?
-              grouped_cert["signed"] = signed
+            # If neither the '--all' flag or the '--certname' flag was passed in
+            # and the requested cert array is empty, we output a JSON object
+            # with an empty 'requested' key. Otherwise, we display
+            # any of the classes that are currently in grouped_cert
+            if grouped_cert.empty?
+              @logger.inform({ "requested" => requested }.to_json)
+            else
+              @logger.inform(grouped_cert.to_json)
             end
-
-            unless revoked.empty?
-              grouped_cert["revoked"] = revoked
-            end
-
-            unless missing.empty?
-              grouped_cert["missing"] = missing
-            end
-
-            grouped_cert.empty? \
-            ? @logger.inform({ "requested" => requested }.to_json) 
-            : @logger.inform(grouped_cert.to_json)
           end
         end
 
@@ -162,7 +159,7 @@ Options:
             @logger.inform "Missing Certificates:"
             missing.each do |name|
               @logger.inform "    #{name}"
-            end 
+            end
           end
         end
 
@@ -214,7 +211,12 @@ Options:
 
         def get_all_certs(settings)
           result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses
-          result ? JSON.parse(result.body) : []
+
+          if result
+            return JSON.parse(result.body)
+          else
+            return []
+          end
         end
 
         def parse(args)
@@ -225,8 +227,11 @@ Options:
 
           errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
 
-          exit_code = errors_were_handled ? 1 : nil
-
+          if errors_were_handled
+            exit_code = 1
+          else
+            exit_code = nil
+          end
           return results, exit_code
         end
       end
--- a/lib/puppetserver/ca/action/list.rb
+++ b/lib/puppetserver/ca/action/list.rb
@@ -30,6 +30,7 @@ Options:
       BANNER
 
         BODY = JSON.dump({desired_state: 'signed'})
+        VALID_FORMAT = ['text', 'json']
 
         def initialize(logger)
           @logger = logger
@@ -47,6 +48,9 @@ Options:
             opts.on('--all', 'List all certificates') do |a|
               parsed['all'] = true
             end
+            opts.on('--format FORMAT', "Valid formats are: 'text' (default), 'json'") do |f|
+              parsed['format'] = f
+            end
             opts.on('--certname NAME[,NAME]', Array, 'List the specified cert(s)') do |cert|
               parsed['certname'] = cert
             end
@@ -57,6 +61,12 @@ Options:
           config = input['config']
           certnames = input['certname'] || []
           all = input['all']
+          output_format = input['format'] || "text"
+          
+          unless VALID_FORMAT.include?(output_format)
+            Errors.handle_with_usage(@logger, ["Unknown format flag '#{output_format}'. Valid formats are '#{VALID_FORMAT.join("', '")}'"])
+            return 1
+          end 
 
           if all && certnames.any?
             Errors.handle_with_usage(@logger, ['Cannot combine use of --all and --certname'])
@@ -79,16 +89,55 @@ Options:
           requested, signed, revoked = separate_certs(all_certs)
           missing = certnames - all_certs.map { |cert| cert['name'] }
 
-          (all || certnames.any?) \
-            ? output_certs_by_state(requested, signed, revoked, missing)
-            : output_certs_by_state(requested)
+          (all || certnames.any?)\
+          ? output_certs_by_state(all, output_format, requested, signed, revoked, missing)
+          : output_certs_by_state(all, output_format, requested)
 
           return missing.any? \
             ? 1
             : 0
         end
+        
+        def output_certs_by_state(all, output_format, requested, signed = [], revoked = [], missing = [])
+          if output_format == 'json'
+            output_certs_json_format(all, requested, signed, revoked, missing)
+          else
+            output_certs_text_format(requested, signed, revoked, missing)
+          end
+        end
 
-        def output_certs_by_state(requested, signed = [], revoked = [], missing = [])
+        def output_certs_json_format(all, requested, signed, revoked, missing)
+          if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
+            @logger.inform([].to_json)
+            return
+          end
+
+          if all
+            grouped_cert = { "requested" => requested,
+                             "signed" => signed,
+                             "revoked" => revoked,
+                             "missing" => missing }.to_json
+            @logger.inform(grouped_cert)
+          else
+            unless requested.empty?
+              @logger.inform(requested.to_json)
+            end
+
+            unless signed.empty?
+              @logger.inform(signed.to_json)
+            end
+
+            unless revoked.empty?
+              @logger.inform(revoked.to_json)
+            end
+
+            unless missing.empty?
+              @logger.inform(missing.to_json)
+            end
+          end
+        end
+
+        def output_certs_text_format(requested, signed, revoked, missing)
           if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
             @logger.inform "No certificates to list"
             return
@@ -113,7 +162,7 @@ Options:
             @logger.inform "Missing Certificates:"
             missing.each do |name|
               @logger.inform "    #{name}"
-            end
+            end 
           end
         end
 
--- a/lib/puppetserver/ca/action/list.rb
+++ b/lib/puppetserver/ca/action/list.rb
@@ -64,12 +64,12 @@ Options:
           output_format = input['format'] || "text"
           
           unless VALID_FORMAT.include?(output_format)
-            Errors.handle_with_usage(@logger, ["Unknown format flag '#{output_format}'. Valid formats are '#{VALID_FORMAT.join("', '")}'"])
+            Errors.handle_with_usage(@logger, ["Unknown format flag '#{output_format}'. Valid formats are '#{VALID_FORMAT.join("', '")}'."])
             return 1
           end 
 
           if all && certnames.any?
-            Errors.handle_with_usage(@logger, ['Cannot combine use of --all and --certname'])
+            Errors.handle_with_usage(@logger, ['Cannot combine use of --all and --certname.'])
             return 1
           end
 
@@ -107,33 +107,33 @@ Options:
         end
 
         def output_certs_json_format(all, requested, signed, revoked, missing)
-          if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
-            @logger.inform([].to_json)
-            return
-          end
+          grouped_cert = {}
 
           if all
             grouped_cert = { "requested" => requested,
                              "signed" => signed,
-                             "revoked" => revoked,
-                             "missing" => missing }.to_json
+                             "revoked" => revoked }.to_json
             @logger.inform(grouped_cert)
           else
             unless requested.empty?
-              @logger.inform(requested.to_json)
+              grouped_cert["requested"] = requested
             end
 
             unless signed.empty?
-              @logger.inform(signed.to_json)
+              grouped_cert["signed"] = signed
             end
 
             unless revoked.empty?
-              @logger.inform(revoked.to_json)
+              grouped_cert["revoked"] = revoked
             end
 
             unless missing.empty?
-              @logger.inform(missing.to_json)
+              grouped_cert["missing"] = missing
             end
+
+            grouped_cert.empty? \
+            ? @logger.inform({ "requested" => requested }.to_json) 
+            : @logger.inform(grouped_cert.to_json)
           end
         end
 
--- a/spec/puppetserver/ca/action/list_spec.rb
+++ b/spec/puppetserver/ca/action/list_spec.rb
@@ -1,6 +1,7 @@
 require 'puppetserver/ca/action/list'
 require 'puppetserver/ca/cli'
 require 'puppetserver/ca/logger'
+require 'json'
 
 RSpec.describe 'Puppetserver::Ca::Action::List' do
   let(:err)    { StringIO.new }
@@ -77,7 +78,10 @@ RSpec.describe 'Puppetserver::Ca::Action::List' do
       allow(action).to receive(:get_all_certs).and_return(result)
       exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'json'})
       expect(exit_code).to eq(0)
-      expect(out.string).to match(/\{.*"name\":\"baz\".*\"fingerprints\".*\}.*.\{.*"name\":\"foo\".*\}/) 
+      parsed_output = JSON.parse(out.string)
+      expect(parsed_output).to be_a_kind_of(Hash)
+      expect(parsed_output).to have_key("signed")
+      expect(parsed_output).to have_key("requested")
     end
 
     it 'does not throw error when text format option is passed in' do
@@ -91,7 +95,20 @@ RSpec.describe 'Puppetserver::Ca::Action::List' do
       allow(action).to receive(:get_all_certs).and_return(result)
       exit_code = action.run({'all' => true, 'format' => 'json'})
       expect(exit_code).to eq(0)
-      expect(out.string).to match(/\{.*"requested\".*\}.*.\{.*"signed\".*\}.*.\{.*"revoked\".*\}/)
+      parsed_output = JSON.parse(out.string)
+      expect(parsed_output).to be_a_kind_of(Hash)
+      expect(parsed_output).to have_key("signed")
+      expect(parsed_output).to have_key("requested")
+      expect(parsed_output).to have_key("revoked")
+    end
+
+    it 'output the non-existent cert as a JSON object with the missing key' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['pup'], 'format' => 'json'})
+      expect(exit_code).to eq(1)
+      parsed_output = JSON.parse(out.string)
+      expect(parsed_output).to be_a_kind_of(Hash)
+      expect(parsed_output).to have_key("missing")
     end
   end
 end
--- a/spec/puppetserver/ca/action/list_spec.rb
+++ b/spec/puppetserver/ca/action/list_spec.rb
@@ -77,14 +77,21 @@ RSpec.describe 'Puppetserver::Ca::Action::List' do
       allow(action).to receive(:get_all_certs).and_return(result)
       exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'json'})
       expect(exit_code).to eq(0)
-      expect(out.string).to match(/\{.*"name\":\"foo\".*\"fingerprints\".*\}/) 
+      expect(out.string).to match(/\{.*"name\":\"baz\".*\"fingerprints\".*\}.*.\{.*"name\":\"foo\".*\}/) 
+    end
+
+    it 'does not throw error when text format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'text'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/Signed Certificates:.*foo.*\(SHA256\).*three.*alt names:.*"DNS:foo", "DNS:bar".*/m)
     end
 
     it 'returns the correct output, including empty keys with the --all option' do
       allow(action).to receive(:get_all_certs).and_return(result)
       exit_code = action.run({'all' => true, 'format' => 'json'})
       expect(exit_code).to eq(0)
-      expect(out.string).to match(/\{.*"requested\".*\}.*.\{.*"signed\".*\}.*.\{.*"revoked\".*\}.*.\{.*"missing\".*\}/)
+      expect(out.string).to match(/\{.*"requested\".*\}.*.\{.*"signed\".*\}.*.\{.*"revoked\".*\}/)
     end
   end
 end
--- a/spec/puppetserver/ca/action/list_spec.rb
+++ b/spec/puppetserver/ca/action/list_spec.rb
@@ -66,5 +66,25 @@ RSpec.describe 'Puppetserver::Ca::Action::List' do
       expect(out.string).to match(/Signed Certificates:.*foo.*\(SHA256\).*three.*alt names:.*"DNS:foo", "DNS:bar".*/m)
       expect(out.string).to match(/Missing Certificates:.*fake.*/m)
     end
+
+    it 'errors when unknown format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'test'})
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not throw error when json format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/\{.*"name\":\"foo\".*\"fingerprints\".*\}/) 
+    end
+
+    it 'returns the correct output, including empty keys with the --all option' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'all' => true, 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/\{.*"requested\".*\}.*.\{.*"signed\".*\}.*.\{.*"revoked\".*\}.*.\{.*"missing\".*\}/)
+    end
   end
 end


lib/puppetserver/ca/action/list.rb
```diff
@@ -78,8 +87,17 @@ def run(input)
           all_certs = get_all_certs(puppet.settings).select { |cert| filter_names.call(cert) }
           requested, signed, revoked = separate_certs(all_certs)
           missing = certnames - all_certs.map { |cert| cert['name'] }
+          
+          if all && output_format == 'json'
+            grouped_cert = { "requested" => requested,
+                             "signed" => signed, 
+                             "revoked" => revoked,
+                             "missing" => missing }.to_json
+            puts grouped_cert
```
Magisus, line 96: This should use `@logger.inform` rather than `puts`. We try to route everything through our logger for better encapsulation.

Tu2607: Gotcha. Thanks for letting me know!

Magisus (commented):
I think if someone doesn't pass `all`, they can't get `json` output right now, but I think we want to support that too, i.e. it should just output the requested certs in json format.

Same for if someone requests a specific certname; we want that to be able to output json also.

It might make sense to try to move the conditional logic (json vs not json) into some of the lower functions? That way you don't have to duplicate logic about what other options are passed (like `all` or `certname`).

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +66,11 @@
       expect(out.string).to match(/Signed Certificates:.*foo.*\(SHA256\).*three.*alt names:.*"DNS:foo", "DNS:bar".*/m)
       expect(out.string).to match(/Missing Certificates:.*fake.*/m)
     end
+
+    it 'errors when unknown format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'text'})
+      expect(exit_code).to eq(1)
+    end
   end
```
mwaggett, line 75: would probably be good to have a positive test as well

mwaggett (commented):
working cat

lib/puppetserver/ca/action/list.rb
```diff
@@ -78,8 +87,17 @@ def run(input)
...
+            return 0 
+          end
 
-          (all || certnames.any?) \
+          ((all || certnames.any?) && output_format == nil) \
```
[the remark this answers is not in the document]
Tu2607: I put the `nil` check there because the previous check on line 65 only take in account that the user enter some sort of format.  But make sense since I was returning in the block right above it.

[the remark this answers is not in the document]
mwaggett: yep!

lib/puppetserver/ca/action/list.rb
```diff
@@ -78,42 +87,67 @@ def run(input)
...
           missing = certnames - all_certs.map { |cert| cert['name'] }
+          
+          if all && output_format == 'json'
+            grouped_cert = { "requested" => requested,
+                             "signed" => signed, 
+                             "revoked" => revoked,
+                             "missing" => missing }.to_json
+            @logger.inform(grouped_cert)
+            return 0 
+          end
```
Magisus, line 98: Do folks have thoughts on whether it makes sense to always have all the keys in the JSON output, even if some of the lists are empty? That maybe feels like better JSON etiquette, but doesn't match what we're doing for text output.

If we don't want keys with empty lists, then I think it might make sense to incorporate this into `output_certs_by_format`, and only add keys to the has that actually have certs associated.

Tu2607: Yeah I can make that change.  Am I waiting on feedback from folks on how they want to do this?

lib/puppetserver/ca/action/list.rb
```diff
@@ -47,6 +47,9 @@ def self.parser(parsed = {})
             opts.on('--all', 'List all certificates') do |a|
               parsed['all'] = true
             end
+            opts.on('--format FORMAT', Array, 'Output in parsable json format') do |f|
```
justinstoller, line 50: This type shouldn't be an Array, IMO. That would be for a description like `--format FORMAT[,FORMAT]` (see `--certname` below) or usage like `--format json,text` which... I don't know how or why you'd want it output as multiple formats at once. And it'd be misleading to allow that since we then just take the first format passed.

justinstoller: I also think this help should say, "Valid formats are: text (default), json" or similar.

And then I think we should handle treating this key as an "enum" of either "text" or "json" (maybe symbols??) What do others think of that??

lib/puppetserver/ca/action/list.rb
```diff
@@ -78,42 +87,67 @@ def run(input)
...
           requested, signed, revoked = separate_certs(all_certs)
           missing = certnames - all_certs.map { |cert| cert['name'] }
+          
+          if all && output_format == 'json'
+            grouped_cert = { "requested" => requested,
+                             "signed" => signed, 
+                             "revoked" => revoked,
+                             "missing" => missing }.to_json
+            @logger.inform(grouped_cert)
+            return 0
```
justinstoller, line 97: We should try to avoid returning like this from the middle of a large method body.

lib/puppetserver/ca/action/list.rb
Tu2607: I see what you meant.  I was thinking of something similar to how I would be accessing ARGV so I did that.

lib/puppetserver/ca/action/list.rb
justinstoller: One thing we could try to do to avoid that is create a data structure that has all the info in it and then pass it to the correct printer method (perhaps even wrapping that choice in a method).

lib/puppetserver/ca/action/list.rb
justinstoller: I don't have a strong opinion on the format (empty array values vs no keys where there'd be empty array values), however I do think that we should refactor the printing to take place in one spot so that shared behaviors are near each other and we aren't returning from the middle of a large method.

lib/puppetserver/ca/action/list.rb
Tu2607: I ended up moving the logic down into the function `output_cert_by_states` so that I don't have to return there. Currently debugging why I can't print out the json with python pretty print.

lib/puppetserver/ca/action/list.rb
Magisus: Agreed. Whatever's easiest then, and we should make sure to test that it does the right thing (either skip keys or put empty arrays in them).

Tu2607: I made changes so now that empty lists don't show up with the `--all` flag. Also, delegated the large chunks of logic down to the `output_certs_by_state`.

lib/puppetserver/ca/action/list.rb
```diff
@@ -47,6 +47,9 @@ def self.parser(parsed = {})
             opts.on('--all', 'List all certificates') do |a|
               parsed['all'] = true
             end
+            opts.on('--format FORMAT', "Valide formats are: 'text'(default), 'json") do |f|
```
justinstoller, line 50:
```suggestion
            opts.on('--format FORMAT', "Valid formats are: 'text' (default), 'json") do |f|
```

lib/puppetserver/ca/action/list.rb
```diff
@@ -57,6 +60,12 @@ def run(input)
           config = input['config']
           certnames = input['certname'] || []
           all = input['all']
+          output_format = input['format']
```
justinstoller, line 63: So, I'm a little torn here, where right now we have an either in formats with a default vs how I feel like formats are often treated (where you may have many). I would probably go with making `output_format` have a default of 'text' like
```rb
output_format = input['format'] || 'text'
```
Then you don't have to check for `nil`, and by doing `case` or `if` / `elsif` blocks it makes it easy to add new formats.

If we didn't go that route and leaned into the either or nature of this, I would probably change the variable `output_format` to be something more like a json flag, eg
```rb
output_as_json = input['format'] == 'json'
```

lib/puppetserver/ca/action/list.rb
```diff
@@ -57,6 +60,12 @@ def run(input)
...
+          
+          if output_format != nil && output_format != 'json'
```
justinstoller, line 65: `output_format` here would be valid as either 'json' or 'text', so the validation probably should be more like ['text', 'json'].include?(output_format)` rather than the `!= 'json'`.

lib/puppetserver/ca/action/list.rb
Tu2607: I like the first suggestion.  Eliminate the need to have check for `nil`.

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,41 +88,62 @@ def run(input)
...
+              all ? grouped_cert["missing"] = missing : @logger.inform(missing.to_json)
+            else
+              @logger.inform "Missing Certificates:"
+              missing.each do |name|
+                @logger.inform "    #{name}"
+              end
+            end 
+          end
+
+          if all
```
justinstoller, line 145: Will this always print a potentially empty json object?

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,41 +88,62 @@ def run(input)
...
           return missing.any? \
             ? 1
             : 0
         end
 
-        def output_certs_by_state(requested, signed = [], revoked = [], missing = [])
+        def output_certs_by_state(all, output_format, requested, signed = [], revoked = [], missing = [])
+          grouped_cert = {}
           if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
             @logger.inform "No certificates to list"
```
justinstoller, line 103: Shouldn't there be a JSON equivalent of this message?

Tu2607: I wasn't sure of that. I could add it in.  But wouldn't that just be an empty json object?

lib/puppetserver/ca/action/list.rb
Tu2607: Ahhh. I forgot to add in the check for the `output_format`. Thanks for catching that.

lib/puppetserver/ca/action/list.rb
Magisus: I think if we're going to have a `format` flag that takes an arg, we should set ourselves up so it'd be easy to add other args later (e.g. yaml).

lib/puppetserver/ca/action/list.rb
mwaggett, line 50: since I can't make a suggestion on Justin's suggestion..
```suggestion
            opts.on('--format FORMAT', "Valid formats are: 'text' (default), 'json'") do |f|
```

lib/puppetserver/ca/action/list.rb
mwaggett: might even be useful to have a `valid_output_format?` method that does this check

lib/puppetserver/ca/action/list.rb
```diff
@@ -57,6 +60,12 @@ def run(input)
           config = input['config']
           certnames = input['certname'] || []
           all = input['all']
+          output_format = input['format']
+          
+          if output_format != nil && output_format != 'json' 
+            Errors.handle_with_usage(@logger, ["Unknown format flag #{output_format}"])
```
mwaggett, line 66: hmm, should this message also include what the valid formats are?

lib/puppetserver/ca/action/list.rb
mwaggett: I think an empty JSON object makes sense.

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,41 +88,62 @@ def run(input)
...
           if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
             @logger.inform "No certificates to list"
             return
           end
 
           unless requested.empty?
-            @logger.inform "Requested Certificates:"
-            output_certs(requested)
+            if output_format
+              all ? grouped_cert["requested"] = requested : @logger.inform(requested.to_json)
```
mwaggett, line 109: This seems weird to me, or at least difficult to read.. imo it's confusing to have a single-line conditional where the two cases are doing totally different things, e.g. assigning a value vs. logging a message.

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,41 +88,62 @@ def run(input)
...
+            else
+              @logger.inform "Requested Certificates:"
+              output_certs(requested)
+              end
```
mwaggett, line 113:
```suggestion
            end
```

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +66,26 @@
       expect(out.string).to match(/Signed Certificates:.*foo.*\(SHA256\).*three.*alt names:.*"DNS:foo", "DNS:bar".*/m)
       expect(out.string).to match(/Missing Certificates:.*fake.*/m)
     end
+
+    it 'errors when unknown format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'text'})
```
mwaggett, line 72: We want to allow 'text', right?

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,41 +88,62 @@ def run(input)
...
+            ? output_certs_by_state(all, output_format, requested, signed, revoked, missing)
+            : output_certs_by_state(all, output_format, requested)
 
           return missing.any? \
             ? 1
             : 0
         end
 
-        def output_certs_by_state(requested, signed = [], revoked = [], missing = [])
+        def output_certs_by_state(all, output_format, requested, signed = [], revoked = [], missing = [])
```
mwaggett, line 100: sooo I kinda think we should include all keys in the JSON output, even if some of the values are empty lists..
I feel like that will require less error-handling from consumers, which seems nice?

spec/puppetserver/ca/action/list_spec.rb
Tu2607: Good eyes, I was goofing around with the test and did that only for json.

lib/puppetserver/ca/action/list.rb
justinstoller: I think if you tell the command to output JSON everything should be in JSON. I'm not sure if an empty string is valid JSON but I'm fine with an empty object.

lib/puppetserver/ca/action/list.rb
justinstoller: lol

lib/puppetserver/ca/action/list.rb
justinstoller: I just noticed the description of the PR has the output as a JSON array. Where here we talk about returning an empty object. I think we should return the same JSON type for every invocation, preferably an object.

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,15 +89,52 @@ def run(input)
           requested, signed, revoked = separate_certs(all_certs)
           missing = certnames - all_certs.map { |cert| cert['name'] }
 
-          (all || certnames.any?) \
+          if output_format == 'json'
+            (all || certnames.any?)\
+            ? output_certs_json_format(all, requested, signed, revoked, missing)
+            : output_certs_json_format(all, requested)
+          else
+            (all || certnames.any?)\
```
justinstoller, line 97: Nice.

Now instead of duplicating the requested vs all logic in a formatting conditional I think we should wrap the call to `output_certs_json_format` and `output_certs_by_state` with one method (maybe `output_certs_by_state` and then rename the current method to `output_certs_text_format`) that calls the correct formatter based a flag you pass. Then the logic of which certs to display is totally separate from the logic about how to format them

Tu2607: Your wish is my command 🙇🏼‍♂️

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,16 +89,55 @@ def run(input)
...
+
+          if all
+            grouped_cert = { "requested" => requested,
+                             "signed" => signed,
+                             "revoked" => revoked,
+                             "missing" => missing }.to_json
+            @logger.inform(grouped_cert)
+          else
+            unless requested.empty?
+              @logger.inform(requested.to_json)
```
justinstoller, line 123: So right now the output for various invocations looks like this:
explicit `--certname` flag where the certname does not exist:
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format text --certname rupert
Missing Certificates:
    rupert
```
vs
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format json --certname rupert 
["rupert"]
```
`--certname` cert exists and is signed:
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format text --certname localhost
Signed Certificates:
    localhost       (SHA256)  8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C   alt names: ["DNS:puppet", "DNS:localhost"]      authorization extensions: [pp_cli_auth: true]
```
vs
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format json --certname localhost | jq .
[
  {
    "name": "localhost",
    "dns_alt_names": [
      "DNS:puppet",
      "DNS:localhost"
    ],
    "state": "signed",
    "authorization_extensions": {
      "pp_cli_auth": "true"
    },
    "subject_alt_names": [
      "DNS:puppet",
      "DNS:localhost"
    ],
    "not_before": "2021-06-30T18:05:03UTC",
    "fingerprints": {
      "SHA1": "C1:05:3F:1E:30:B0:C2:A3:FA:95:FB:0F:31:54:38:03:51:2C:FC:8B",
      "SHA256": "8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C",
      "SHA512": "06:4F:94:92:B5:3C:8A:AF:58:96:0D:23:8F:E9:85:3B:9E:26:57:0F:1F:12:D1:40:EE:6E:FF:5B:01:D4:FA:57:7E:D4:50:A6:7D:99:F2:79:E0:5D:0F:EE:EF:B0:59:1F:33:16:89:0C:C6:7D:69:EE:05:80:3E:1D:F8:48:42:EE",
      "default": "8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C"
    },
    "not_after": "2026-06-30T18:05:03UTC",
    "serial_number": 2,
    "fingerprint": "8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C"
  }
]
```
all
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format text --all
Signed Certificates:
    localhost       (SHA256)  8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C   alt names: ["DNS:puppet", "DNS:localhost"]      authorization extensions: [pp_cli_auth: true]
```
vs
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format json --all | jq .
{
  "requested": [],
  "signed": [
    {
      "name": "localhost",
      "dns_alt_names": [
        "DNS:puppet",
        "DNS:localhost"
      ],
      "state": "signed",
      "authorization_extensions": {
        "pp_cli_auth": "true"
      },
      "subject_alt_names": [
        "DNS:puppet",
        "DNS:localhost"
      ],
      "not_before": "2021-06-30T18:05:03UTC",
      "fingerprints": {
        "SHA1": "C1:05:3F:1E:30:B0:C2:A3:FA:95:FB:0F:31:54:38:03:51:2C:FC:8B",
        "SHA256": "8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C",
        "SHA512": "06:4F:94:92:B5:3C:8A:AF:58:96:0D:23:8F:E9:85:3B:9E:26:57:0F:1F:12:D1:40:EE:6E:FF:5B:01:D4:FA:57:7E:D4:50:A6:7D:99:F2:79:E0:5D:0F:EE:EF:B0:59:1F:33:16:89:0C:C6:7D:69:EE:05:80:3E:1D:F8:48:42:EE",
        "default": "8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C"
      },
      "not_after": "2026-06-30T18:05:03UTC",
      "serial_number": 2,
      "fingerprint": "8A:DE:19:07:93:48:1B:31:B2:BB:FA:37:74:88:D3:DF:53:5A:B2:77:FF:70:97:C8:23:F1:EB:80:9B:53:4F:7C"
    }
  ],
  "revoked": [],
  "missing": []
}
```
bare invocation (defaults to pending):
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format json | jq .
[]
```
vs
```
No certificates to list
```
with a pending CSR:
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format text
Requested Certificates:
    fooberry       (SHA256)  63:70:1E:97:32:BA:CE:AB:70:6B:6B:24:91:92:7B:27:0F:DE:69:F1:64:34:E9:2A:35:E7:24:39:96:2E:56:84
```
vs
```
 ~/Work/server/puppetserver-ca-cli | bundle exec exe/puppetserver-ca list --format json | jq .
[
  {
    "name": "fooberry",
    "state": "requested",
    "dns_alt_names": [],
    "subject_alt_names": [],
    "authorization_extensions": {},
    "fingerprint": "63:70:1E:97:32:BA:CE:AB:70:6B:6B:24:91:92:7B:27:0F:DE:69:F1:64:34:E9:2A:35:E7:24:39:96:2E:56:84",
    "fingerprints": {
      "SHA1": "21:72:43:DE:4A:EB:83:74:02:38:4C:6F:5F:06:E1:D9:DB:AA:04:5D",
      "SHA256": "63:70:1E:97:32:BA:CE:AB:70:6B:6B:24:91:92:7B:27:0F:DE:69:F1:64:34:E9:2A:35:E7:24:39:96:2E:56:84",
      "SHA512": "90:65:D4:11:FD:86:AC:FC:54:EA:5D:67:F7:4B:EA:33:0B:61:1E:04:03:C3:50:B9:4A:68:78:66:11:27:59:04:DC:20:D6:2A:EA:82:D0:6F:F8:AF:FD:D6:4F:2A:23:22:26:C8:2D:A7:F9:91:56:3F:9F:BC:DA:09:C9:3E:25:A5",
      "default": "63:70:1E:97:32:BA:CE:AB:70:6B:6B:24:91:92:7B:27:0F:DE:69:F1:64:34:E9:2A:35:E7:24:39:96:2E:56:84"
    }
  }
]
```

justinstoller: So, if you pass `--all` you get a JSON object, but if you don't you get a JSON array. And if you pass `--certname` the contents of the array are either an object representing a cert or a string representing a certname that is missing a cert on the CA. If you don't pass either flag you get a potentially empty list of pending csrs. In both the missing cert and pending csr cases the text output gives you some nice context to say what kind of entity you're dealing with.

Tu2607: Is there any trimming or changes I should make?

justinstoller: So I think we should have one type of entity returned by all the invocations. And I think it should be a JSON object because they are easier to extend with more information later, also the keys will provide the kind of information that the heading were providing before. eg `{ "missing": [ "rupert" ] }` when `--certname rupert` doesn't exist on the CA.

justinstoller: That leaves a little bit of weirdness where we already discussed for `--all` showing keys with empty arrays if there isn't any objects to populate them, eg `{ "revoked": [], "signed": [], "pending": [ {...} ], "missing": [] }`  However, it seems nice that excluding keys is valuable when we're explicitly excluding that class of object from the output, as we do with missing `--certname` certs or the default listing of pending certs.

mwaggett: okay I've gotten a little confused.. I'm not really understanding how this method takes `--certname` into account, so I don't see why we wouldn't want to log `grouped_cert` for all cases..

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +66,25 @@
...
       expect(out.string).to match(/Missing Certificates:.*fake.*/m)
     end
+
+    it 'errors when unknown format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'test'})
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not throw error when json format option is passed in' do
```
mwaggett, line 76: Should we also test that `text` can be passed, even though it's also the default?

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +66,25 @@
...
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/\{.*"name\":\"foo\".*\"fingerprints\".*\}/)
```
mwaggett, line 80: Should we test that this includes info for the 'baz' cert as well?

lib/puppetserver/ca/action/list.rb
```diff
@@ -57,6 +61,12 @@ def run(input)
           config = input['config']
           certnames = input['certname'] || []
           all = input['all']
+          output_format = input['format'] || "text"
+          
+          unless VALID_FORMAT.include?(output_format)
+            Errors.handle_with_usage(@logger, ["Unknown format flag '#{output_format}'. Valid formats are '#{VALID_FORMAT.join("', '")}'"])
```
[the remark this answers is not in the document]
Tu2607: I did put a period in there but then I look at the other error string and it didn't have a period so I dropped the period...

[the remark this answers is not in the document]
mwaggett: heh. yeah, we're not especially consistent, so it doesn't really matter. but I personally find it jarring when there's a message with a period in the middle but not one on the end. ¯\_(ツ)_/¯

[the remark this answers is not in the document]
justinstoller: I'm probably the perpetrator of many of those. I have to stop myself from writing things like the below constantly:
```
unknown format flag 'foo'.  (Valid formats are: text (default), yaml and json
```

lib/puppetserver/ca/action/list.rb
justinstoller: All the certs passed in have already been filtered by any `--certname` requirements on line 88 of this PR. But mainly I wanted to clarify about when I would include keys vs exclude keys based on my above suggestion and our previous discussion about having all keys when `--all` was given.

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,16 +89,55 @@ def run(input)
...
+          if revoked.empty? && signed.empty? && requested.empty? && missing.empty?
+            @logger.inform([].to_json)
+            return
+          end
+
+          if all
+            grouped_cert = { "requested" => requested,
+                             "signed" => signed,
+                             "revoked" => revoked,
+                             "missing" => missing }.to_json
```
justinstoller, line 119: Talked with Tu in chat about this already, but for review purposes, "missing" isn't a valid key in `--all` behavior.

lib/puppetserver/ca/action/list.rb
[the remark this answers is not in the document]
Tu2607: I'll add the period to the other error string then

lib/puppetserver/ca/action/list.rb
mwaggett: I guess I'm saying I would probably include all keys all the time.

justinstoller: I'm a little concerned that doing a:
```
puppetserver ca list
{
  "pending": [ <foo csr info>, <bar csr info> ],
  "revoked": [],
  "signed": []
}
```
could be confusing when there _are_ signed and revoked certs, we just don't show them by default.

Magisus: Agreed, this seems bad.

mwaggett: hmm okay, now that I look at it, I think that's a good point. I was mostly thinking about a computer parsing this, which would mean it would have to account for nil keys when trying to get at this info, similar to my argument for including all the keys for `all`. whereas the 'text' format is for human consumption.

Based on all that, I'm wondering if it is actually better to keep the original format, where using `--certname` gives you an array of certs and you could just determine the state from that, if you even need it. I definitely understand wanting to return a consistent object type, but I'm wondering if we should be treating the `--all` case vs. the `--certname` case as more different than we have been. I'm not really sure of the use cases for using one or the other.

justinstoller: Yeah, I thought that we could make `--certname` return an object and be easier to parse than the sorted-by-state way to show other results by doing
```
puppetserver ca list --certname foo,unknowncert

{
  "foo": <foo cert or csr info>,
  "unknowncert": { "state": "missing" }
}
```
but that's _real_ different in format from the other. I where on the spectrum ease of parsing for specific workflows is worth the user's cognitive overhead of different formats?

justinstoller: You make a good point about, users aren't going to see this so the semantic difference between "a key is here because you request that state be output even though it may be empty" and "a key is always going to be here" probably doesn't mean much? I'm not adamant that we need to make that distinction, but I'm inclined to.

Tu2607:
Here's another formatting issue that I run into and don't know how you guys feels about it.  When I ran this command `list --certname tu,localhost`, I got two separate JSON object since tu was revoked and locahost is signed. Should they be grouped together or should they remain two separate JSON objects?

justinstoller:
> Here's another formatting issue that I run into and don't know how you guys feels about it. When I ran this command list --certname tu,localhost, I got two separate JSON object since tu was revoked and locahost is signed. Should they be grouped together or should they remain two separate JSON objects?

grouped

lib/puppetserver/ca/action/list.rb
```diff
@@ -79,16 +89,55 @@ def run(input)
...
+            unless revoked.empty?
+              grouped_cert["revoked"] = revoked
+            end
+
+            unless missing.empty?
+              grouped_cert["missing"] = missing
+            end
+
+            grouped_cert.empty? \
+            ? @logger.inform({ "requested" => requested }.to_json)
```
justinstoller, line 135: It'd be nice to at least have a comment here about how we should only get here if neither `--all` or `--certname` was passed and so we will just be giving a list of the CSRs.

lib/puppetserver/ca/action/list.rb
```diff
@@ -113,7 +159,7 @@ def output_certs_by_state(requested, signed = [], revoked = [], missing = [])
             @logger.inform "Missing Certificates:"
             missing.each do |name|
               @logger.inform "    #{name}"
-            end
+            end
```
justinstoller, line 162: space at end of line

lib/puppetserver/ca/action/list.rb
```diff
@@ -165,19 +211,28 @@ def separate_certs(all_certs)
...
+            return JSON.parse(result.body)
+          else 
+            return []
+          end
         end
 
         def parse(args)
           results = {}
           parser = self.class.parser(results)
+          exit_code = 0
```
justinstoller, line 225: why are you setting this here?

Tu2607: Declare a variable before setting it? Or should I have just return the exit code inside of the if/else block. My old habit of writing C code is showing...

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +66,32 @@
...
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'test'})
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not throw error when json format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/\{.*"name\":\"baz\".*\"fingerprints\".*\}.*.\{.*"name\":\"foo\".*\}/)
```
justinstoller, line 80: IMO, it'd be good to do something like:
```rb
parsed_outpute = nil
expect {
  parsed_output = JSON.parse(out.string)
}.not_to raise_error
expect(parsed_output).to be_a_kind_of(Hash)
```

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +66,32 @@
...
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'text'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/Signed Certificates:.*foo.*\(SHA256\).*three.*alt names:.*"DNS:foo", "DNS:bar".*/m)
+    end
+
+    it 'returns the correct output, including empty keys with the --all option' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'all' => true, 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect(out.string).to match(/\{.*"requested\".*\}.*.\{.*"signed\".*\}.*.\{.*"revoked\".*\}/)
```
justinstoller, line 94: Similarly to above regarding parsing the JSON output. I would prefer to see us parse the out.string with JSON.parse and then make assertions about the data structure we get back.

lib/puppetserver/ca/action/list.rb
justinstoller: Yeah  ( : 
You don't have to declare a variable or zero it out before using it.
This is just setting it to one thing and then mutating it later on, you can remove this line and then leave the rest as-is.

Tu2607: Ruby's scopes are interesting. Didn't know I could use a variable declared inside an if statement to be usable outside of it. Pretty neat.

lib/puppetserver/ca/action/list.rb
```diff
@@ -57,9 +61,16 @@ def run(input)
           config = input['config']
           certnames = input['certname'] || []
           all = input['all']
+          output_format = input['format'] || "text"
+          filter_names = []
```
mwaggett, line 65: I think we can get rid of this too, for the same reason as not needing to set `exit_code` up here. Leaving it is a little confusing, since it gets defined as a lambda below, but this implies that it should be an array.

lib/puppetserver/ca/action/list.rb
```diff
@@ -71,24 +82,59 @@ def run(input)
...
+            grouped_cert["revoked"] = revoked unless revoked.empty?
+            grouped_cert["missing"] = missing unless missing.empty?
+
+            # If neither the '--all' flag or the '--certname' flag was passed in
+            # then we will display the requested certificates. Otherwise, display
+            # the classes that are in grouped_cert
+            if grouped_cert.empty?
+              @logger.inform({ "requested" => requested }.to_json)
+            else
+               @logger.inform(grouped_cert.to_json)
```
mwaggett, line 132: (🦅 👁️ nit) 
```suggestion
              @logger.inform(grouped_cert.to_json)
```

lib/puppetserver/ca/action/list.rb
```diff
@@ -71,24 +82,59 @@ def run(input)
...
+            grouped_cert["requested"] = requested unless requested.empty?
+            grouped_cert["signed"] = signed unless signed.empty?
+            grouped_cert["revoked"] = revoked unless revoked.empty?
+            grouped_cert["missing"] = missing unless missing.empty?
+
+            # If neither the '--all' flag or the '--certname' flag was passed in
+            # then we will display the requested certificates. Otherwise, display
+            # the classes that are in grouped_cert
+            if grouped_cert.empty?
+              @logger.inform({ "requested" => requested }.to_json)
```
mwaggett, line 130: I'm a little confused about this.. If we've reached this case, `requested` must be empty, right? I'm not understanding why we need this conditional at all.. The caller of this function seems to cover only passing in `requested` certs when `--all` nor `--certname` was passed, so I think you'd still get the desired result without this `if` block.

This also doesn't seem to match what we do for the `text` mode - when everything is empty, we say that, so I would expect the JSON result to just be an empty hash?

lib/puppetserver/ca/action/list.rb
```diff
@@ -176,8 +227,11 @@ def parse(args)
 
           errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
 
-          exit_code = errors_were_handled ? 1 : nil
-
+          if errors_were_handled
+            exit_code = 1
+          else
+            exit_code = nil
```
mwaggett, line 233: I know this was here before, but didn't we talk about how maybe this should be `0` instead of `nil`? since `nil` isn't an exit code??

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +67,40 @@
...
     end
+
+    it 'errors when unknown format option is passed in' do
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'test'})
+      expect(exit_code).to eq(1)
+    end
+
+    it 'does not throw error when json format option is passed in' do
+      parsed_output = nil
```
mwaggett, line 78: I don't _think_ this is needed.

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +67,40 @@
...
+
+    it 'returns the correct output, including empty keys with the --all option' do
+      parsed_output = nil
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'all' => true, 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect {
+        parsed_output = JSON.parse(out.string)
+      }.not_to raise_error
+      expect(parsed_output).to be_a_kind_of(Hash)
```
mwaggett, line 103: I think it would be good to test that all the keys are present in this hash.

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -66,5 +67,40 @@
...
+
+    it 'does not throw error when json format option is passed in' do
+      parsed_output = nil
+      allow(action).to receive(:get_all_certs).and_return(result)
+      exit_code = action.run({'certname' => ['foo', 'baz'], 'format' => 'json'})
+      expect(exit_code).to eq(0)
+      expect {
+        parsed_output = JSON.parse(out.string)
+      }.not_to raise_error
+      expect(parsed_output).to be_a_kind_of(Hash)
```
mwaggett, line 85: Would probably be good to test that the keys we expect are present.

mwaggett (commented):
almost there

lib/puppetserver/ca/action/list.rb
Tu2607: Ah Justin wanted the case if we passed no args to `list`, we would output whatever that is currently in `requested`.

lib/puppetserver/ca/action/list.rb
Tu2607: I tried changing it to 0 and tested it.  The way that we have thing set up over `cli.rb` would require the `exit_code` here to not be a number.

spec/puppetserver/ca/action/list_spec.rb
Tu2607: Right on. I'll look into it.

lib/puppetserver/ca/action/list.rb
Tu2607: I need to use auto-formatter more...

lib/puppetserver/ca/action/list.rb
mwaggett: but isn't that taken care of by line 98, by just not passing anything for the other states?
https://github.com/puppetlabs/puppetserver-ca-cli/pull/80/files#diff-805a8aad50ad3c6c816ce6c5607c63d12a5ab8c37a5994877013fd87f88c3a74R98 ?

lib/puppetserver/ca/action/list.rb
mwaggett: ahh, gotcha. we can leave it then.

lib/puppetserver/ca/action/list.rb
Tu2607: https://github.com/puppetlabs/puppetserver-ca-cli/blob/58c082b7fe8ac77bd8cf267795577e25b17bfc85/lib/puppetserver/ca/cli.rb#L94

lib/puppetserver/ca/action/list.rb
Tu2607: https://github.com/puppetlabs/puppetserver-ca-cli/blob/58c082b7fe8ac77bd8cf267795577e25b17bfc85/lib/puppetserver/ca/cli.rb#L94

spec/puppetserver/ca/action/list_spec.rb
Tu2607: The `parsed_output` ?

mwaggett: setting it ahead of time, yeah. right? since you set it within the `expect`?

lib/puppetserver/ca/action/list.rb
Tu2607: My thought process was that at any given moment if `grouped_cert` was empty, that meant that we didn't pass in any arguments, otherwise there has to be at least something in the classes, hence why I did that. I see what you meant and I wasn't sure of how I could accommodate to Justin's suggestion so I landed on that :)

spec/puppetserver/ca/action/list_spec.rb
Tu2607: Apparently that not declaring it there didn't work...I commented that out and the test broke lol.

mwaggett: interesting! I think this has to do with the fact that we're using a block in the `expect`. 

I'm actually remembering reading that it's not a best practice to have tests like `expect { blah }.not_to raise_error` - you can just set `parsed_output = JSON.parse(out.string)` and if an error is raised, the test will fail anyway.

lib/puppetserver/ca/action/list.rb
Magisus: I think this takes care of the case when someone passed no flag, _and_ there are no requested certs. Makes it so we will output `{ requested: [] }` instead of just `{}`, which I think makes sense to me. Might want to update the comment to describe the specific circumstance when we expect to hit this branch.

Tu2607: ^ What I tried to explain in a very poorly way.

spec/puppetserver/ca/action/list_spec.rb
Tu2607: Neat! Ruby's scoping is pretty interesting and it's something I'm gonna have to catch up on.

Magisus: I think because it's not guaranteed if/when the block will run, you need to declare it first, otherwise it's local to scope of the block. Blocks are [closures](https://www.sitepoint.com/closures-ruby/): they can reference things declared outside of them at the time they were created, and do things like update their values, like we do here. But things created inside of them aren't visible outside, so just declaring `parsed_output` inside the block doesn't let us use it outside of that block.

lib/puppetserver/ca/action/list.rb
mwaggett: let me see if I can lay this out to un-confuse us both.. 😅 

- for the `--all` case, we pass params for all states and always output all keys, even if some states are empty ✅ 
- for the `--certnames` case, we pass params for all states, but only output keys for non-empty states ✅ 
  - I don't think it's possible for `grouped_cert` to be empty in this case, since anything we don't recognize will land in `missing` ✅ 
- for the no flags case, we only pass a param for `requested`, so the rest will be empty ✅ 
  - the only way for `grouped_cert` to be empty is if there are no `requested` certs - it sounds like Justin was saying we should still return `{ "requested" => [] }` in that case, as opposed to an empty JSON object? 
 
I think I understand now haha, though not totally sure if I agree with that behavior, since the text version just says "No certificates to list".

spec/puppetserver/ca/action/list_spec.rb
Tu2607: heh like how local variables declared in `if` block over in C/C++ land work

lib/puppetserver/ca/action/list.rb
Tu2607: All of your bullet points are correct! At least that's what I understand from Justin's suggestions.

Magisus: I could go either way. I always find myself a little confused about what the behavior is when no flag is passed, and I think this is actually clearer than what the text version does, so I'm kind of in favor of it.

mwaggett: ah yeah, that's fair. I'm good with leaving it, but agree that the comment could be clarified to describe exactly how we get here.

mwaggett: oh I see you did that already 😅

Tu2607: I got y'all lol

Magisus (approved):

mwaggett (approved):
fuck yeah!!!

PR #83, opened 2021-07-09
(SERVER-2740) Pruning CRL(s) through Puppetserver-ca CLI

Tu2607:
Early stages of a command for Puppetserver-ca CLI to prune CRL(s) on disk. Please send halp!

### DIFF
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -1,0 +1,116 @@
+require 'optparse'
+require 'openssl'
+require 'puppetserver/ca/errors'
+require 'puppetserver/ca/utils/cli_parsing'
+require 'puppetserver/ca/utils/file_system'
+require 'puppetserver/ca/utils/config'
+require 'puppetserver/ca/x509_loader'
+
+module Puppetserver
+  module Ca
+    module Action
+      class Prune
+        include Puppetserver::Ca::Utils
+
+        SUMMARY = "Prune the local CRL on disk to remove any duplicated certificates"
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca prune [--help]
+  puppetserver ca prune [--config]
+
+Description:
+  Prune the list of revoked certificates of any duplication within it.  This command
+  will only prune the CRL issued by Puppet's CA cert.
+
+Options:
+BANNER
+
+        def initialize(logger)
+          @logger = logger
+        end
+
+        def run(inputs)
+          config_path = inputs['config']
+
+          # Validate the config path.
+          if config_path
+            errors = FileSystem.validate_file_paths(config_path)
+            return 1 if Errors.handle_with_usage(@logger, errors)
+          end
+
+          # Validate puppet config setting.
+          puppet = Config::Puppet.new(config_path)
+          puppet.load(logger: @logger)
+          return 1 if Errors.handle_with_usage(@logger, puppet.errors)
+
+          # Validate that we are offline
+          return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+
+          # Getting the CRL(s)
+          loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
+
+          puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
+          prune_CRLs(puppet_crl)
+          update_pruned_CRL(puppet_crl, loader.key)
+          FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+
+          @logger.inform("Finished pruning Puppet's CRL")
+          return 0
+        end
+
+        def prune_CRLs(crl_list)
+          crl_list.each do |crl|
+            existed_serial_number = Set.new()
+            revoked_list = crl.revoked
+            @logger.debug("Pruning duplicate entries in CRL for issuer " \
+              "#{crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)}") if @logger.debug?
+
+            revoked_list.delete_if do |revoked|
+              if existed_serial_number.add?(revoked.serial)
+                false
+              else
+                @logger.debug("Removing duplicate of #{revoked.serial}, " \
+                  "revoked on #{revoked.time}\n") if @logger.debug?
+                true
+              end
+            end
+            crl.revoked=(revoked_list)
+          end
+        end
+
+        def update_pruned_CRL(crl_list, pkey)
+          crl_list.each do |crl|
+            crl.version=(crl.version + 1)
+            crl.sign(pkey, OpenSSL::Digest::SHA256.new)
+          end
+        end
+
+        def self.parser(parsed = {})
+          OptionParser.new do |opts|
+            opts.banner = BANNER
+            opts.on('--help', 'Display this command-specific help output') do |help|
+              parsed['help'] = true
+            end
+            opts.on('--config CONF', 'Path to the puppet.conf file on disk') do |conf|
+              parsed['config'] = conf
+            end
+          end
+        end
+
+        def parse(args)
+          results = {}
+          parser = self.class.parser(results)
+          errors = CliParsing.parse_with_errors(parser, args)
+          errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
+
+          if errors_were_handled
+            exit_code = 1
+          else
+            exit_code = nil
+          end
+          return results, exit_code
+        end
+      end
+    end
+  end
+end
--- a/lib/puppetserver/ca/cli.rb
+++ b/lib/puppetserver/ca/cli.rb
@@ -26,11 +26,12 @@ Manage the Private Key Infrastructure for
 Puppet Server's built-in Certificate Authority
 BANNER
 
-      INIT_ACTIONS = {
+      ADMIN_ACTIONS = {
         'import'   => Action::Import,
         'setup'    => Action::Setup,
         'enable'   => Action::Enable,
-        'migrate'  => Action::Migrate
+        'migrate'  => Action::Migrate,
+        'prune'    => Action::Prune
       }
 
       MAINT_ACTIONS = {
@@ -38,19 +39,18 @@ BANNER
         'generate' => Action::Generate,
         'list'     => Action::List,
         'revoke'   => Action::Revoke,
-        'sign'     => Action::Sign,
-        'prune'    => Action::Prune
+        'sign'     => Action::Sign
       }
 
-      VALID_ACTIONS = INIT_ACTIONS.merge(MAINT_ACTIONS).sort.to_h
+      VALID_ACTIONS = ADMIN_ACTIONS.merge(MAINT_ACTIONS).sort.to_h
 
       ACTION_LIST = "\nAvailable Actions:\n\n" +
         "  Certificate Actions (requires a running Puppet Server):\n\n" +
         MAINT_ACTIONS.map do |action, cls|
           "    #{action}\t#{cls::SUMMARY}"
         end.join("\n") + "\n\n" +
-        "  Initialization Actions (requires Puppet Server to be stopped):\n\n" +
-        INIT_ACTIONS.map do |action, cls|
+        "  Administrative Actions (requires Puppet Server to be stopped):\n\n" +
+        ADMIN_ACTIONS.map do |action, cls|
           "    #{action}\t#{cls::SUMMARY}"
         end.join("\n")
 
--- a/lib/puppetserver/ca/cli.rb
+++ b/lib/puppetserver/ca/cli.rb
@@ -8,6 +8,7 @@ require 'puppetserver/ca/action/list'
 require 'puppetserver/ca/action/revoke'
 require 'puppetserver/ca/action/setup'
 require 'puppetserver/ca/action/sign'
+require 'puppetserver/ca/action/prune'
 require 'puppetserver/ca/action/migrate'
 require 'puppetserver/ca/errors'
 require 'puppetserver/ca/logger'
@@ -28,8 +29,8 @@ BANNER
       INIT_ACTIONS = {
         'import'   => Action::Import,
         'setup'    => Action::Setup,
-        'enable' => Action::Enable,
-        'migrate' => Action::Migrate,
+        'enable'   => Action::Enable,
+        'migrate'  => Action::Migrate
       }
 
       MAINT_ACTIONS = {
@@ -37,7 +38,8 @@ BANNER
         'generate' => Action::Generate,
         'list'     => Action::List,
         'revoke'   => Action::Revoke,
-        'sign'     => Action::Sign
+        'sign'     => Action::Sign,
+        'prune'    => Action::Prune
       }
 
       VALID_ACTIONS = INIT_ACTIONS.merge(MAINT_ACTIONS).sort.to_h
--- a/lib/puppetserver/ca/logger.rb
+++ b/lib/puppetserver/ca/logger.rb
@@ -17,8 +17,12 @@ module Puppetserver
         @level
       end
 
+      def debug?
+        return @level >= LEVELS[:debug]
+      end
+
       def debug(text)
-        if @level >= LEVELS[:debug]
+        if debug?
           @out.puts(text)
         end
       end
--- a/spec/puppetserver/ca/action/prune_spec.rb
+++ b/spec/puppetserver/ca/action/prune_spec.rb
@@ -1,0 +1,107 @@
+require 'spec_helper'
+require 'utils/ssl'
+require 'utils/http'
+
+require 'tmpdir'
+
+require 'puppetserver/ca/logger'
+require 'puppetserver/ca/action/prune'
+require 'puppetserver/ca/utils/http_client'
+
+RSpec.describe Puppetserver::Ca::Action::Prune do
+  include Utils::SSL
+
+  let(:stdout) { StringIO.new }
+  let(:stderr) { StringIO.new }
+  let(:logger) { Puppetserver::Ca::Logger.new(:info, stdout, stderr) }
+
+  subject { Puppetserver::Ca::Action::Prune.new(logger) }
+
+  describe 'flags' do
+    it 'take a custom path to the puppet.conf' do
+      result, maybe_exit_code = subject.parse(['--config', '/dev/tcp/example.com'])
+      expect(maybe_exit_code).to be(nil)
+      expect(result['config']).to eq('/dev/tcp/example.com')
+    end
+  end
+
+  describe 'prune' do
+    let(:connection) { double }
+    let(:online) { Utils::Http::Result.new('200', 'running') }
+    let(:offline) { Utils::Http::Result.new('503', 'offline') }
+
+    before do
+      allow_any_instance_of(Puppetserver::Ca::Utils::HttpClient).
+        to receive(:with_connection).and_yield(connection)
+      allow_any_instance_of(Puppetserver::Ca::Utils::HttpClient).
+        to receive(:make_store)
+    end
+
+    it 'errors when validating path to puppet.conf' do
+      exit_code = subject.run({'config' => "fake/faker/puppet.conf"})
+      expect(exit_code).to eq(1)
+      expect(stderr.string).to eq("Error:\nCould not read file 'fake/faker/puppet.conf'\n")
+    end
+
+    it 'refuses to prune when server is running' do
+      allow(connection).to receive(:get).and_return(online)
+      Dir.mktmpdir do |tmpdir|
+        with_temp_dirs tmpdir do |config|
+          exit_code = subject.run({'config' => config})
+          expect(exit_code).to eq(1)
+          expect(stderr.string).to include('server service is running')
+        end
+      end
+    end
+
+    it 'errors when fails to validate puppet settings' do
+      allow(connection).to receive(:get).and_return(offline)
+      Dir.mktmpdir do |dir|
+        puppet_conf = File.join(dir, 'puppet.conf')
+        File.open puppet_conf, 'w' do |f|
+          f.puts(<<-INI)
+            [main]
+              certname = $cadir
+          INI
+        end
+        exit_code = subject.run({ 'config' => puppet_conf })
+        expect(exit_code).to eq(1)
+        expect(stderr.string).to include('Could not parse')
+      end
+    end
+
+    it 'reduces a CRL with duplicate revoked certs down to 1 cert' do
+      ca_key = OpenSSL::PKey::RSA.new(512)
+      ca_cert = create_cert(ca_key, "You-Shall-Not-Pass")
+      revoked_cert = create_cert(ca_key, 'revoked')
+      ca_crl = create_crl(ca_cert, ca_key, Array.new(5, revoked_cert))
+
+      subject.prune_CRLs([ca_crl])
+      expect(ca_crl.revoked.length).to eq(1)
+    end
+
+    it 'deduplicates a CRL with multiple certs that have duplicate of themselves' do
+      ca_key = OpenSSL::PKey::RSA.new(512)
+      ca_cert = create_cert(ca_key, "You-Shall-Not-Pass")
+      first_cert = Array.new(5, create_cert(ca_key, 'first'))
+      second_cert = Array.new(10, create_cert(ca_key, 'second'))
+      third_cert = Array.new(3, create_cert(ca_key, 'third'))
+
+      ca_crl = create_crl(ca_cert, ca_key, first_cert + second_cert + third_cert)
+
+      subject.prune_CRLs([ca_crl])
+      expect(ca_crl.revoked.length).to eq(3)
+    end
+  end
+
+  describe 'update' do
+    it 'bumps CRL version up by 1' do
+      ca_key = OpenSSL::PKey::RSA.new(512)
+      ca_cert = create_cert(ca_key, "Bazzup")
+      ca_crl = create_crl(ca_cert, ca_key)
+
+      subject.update_pruned_CRL([ca_crl], ca_key)
+      expect(ca_crl.version).to eq(2)
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -55,6 +55,11 @@ To create a new keypair and certificate for a certname:
 puppetserver ca generate --certname foo.example.com
 ```
 
+To remove duplicated entries from Puppet's CRL:
+```
+puppetserver ca prune
+```
+
 To enable verbose mode:
 ```
 puppetserver ca --verbose <action>
--- a/spec/puppetserver/ca/cli_spec.rb
+++ b/spec/puppetserver/ca/cli_spec.rb
@@ -58,8 +58,13 @@ RSpec.describe Puppetserver::Ca::Cli do
       /.*Usage.* puppetserver ca sign.*Display this command-specific help output.*/m
   end
 
+  describe 'the prune action' do
+    include_examples 'basic cli args',
+      'prune',
+      /.*Usage.* puppetserver ca prune.*Display this command-specific help output.*/m
+  end
   # This test is a representation of what to expect when the verbose flag
-  # is raised with an action. We're using the 'clean' action as an example 
+  # is raised with an action. We're using the 'clean' action as an example
   it 'raise the verbose flag' do
     args = ['--verbose', 'clean'].compact
     action_class = Puppetserver::Ca::Cli::VALID_ACTIONS[args[1]]


Tu2607:
Here's an example of the `revoked_lists_by_issuer`:
`{"/CN=Puppet CA: TuVu-C02FJ00XML87.local"=>[#<OpenSSL::X509::Revoked:0x00007fa1d2ba5600>, #<OpenSSL::X509::Revoked:0x00007fa1d2ba55d8>, #<OpenSSL::X509::Revoked:0x00007fa1d2ba55b0>], "/CN=Puppet Root CA: a8730dd27576b6"=>[]}`

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,116 @@
...
+          crl_list = loader.crls    # A reference to the CRL list
+          prune_per_issuer(crl_list)
+          return 0                  # Place holder return value for now
+        end
+
+        # Given that we have an array of CRL, each CRL contain a list of revoked
+        # certs signed by a specific issuer.  Thus by iterating through each
+        # item of the CRL array, we can invoke each item's list of revoked cert.
+        # Since it is another array, we will have to iterate through that as well
+        # to prune.  Currently pruning by serial number ?
```
justinstoller, line 80: serial number is the correct thing to prune by though @tvpartytonight or @mwaggett might want to weigh in.

justinstoller:
I skipped any grammar kind of comments, but the overall approach seems good.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,116 @@
...
+        def prune_per_issuer(crl_list)
+          crl_list.each do |list|
+            existed_serial_number = Set.new()
```
justinstoller, line 83: fwiw, you can also use a hash here

Tu2607: That was my original approach.  Did some research on set vs hash and found out that the background work of a set is essentially a hash.  Thus why I ended up using a set (mainly because of the `include?` function)

lib/puppetserver/ca/cli.rb
```diff
@@ -28,16 +29,17 @@ class Cli
...
       }
 
       MAINT_ACTIONS = {
         'clean'    => Action::Clean,
         'generate' => Action::Generate,
         'list'     => Action::List,
         'revoke'   => Action::Revoke,
-        'sign'     => Action::Sign
+        'sign'     => Action::Sign,
+        'prune'    => Action::Prune
```
mwaggett, line 42: hmm. it makes sense semantically to include `prune` in the `MAINT_ACTIONS`, but it then the `ACTION_LIST` below will say that this action requires a running Puppet Server, which is not true. Not sure how we want to address that..

Tu2607: we can add in a tidbit like `except for prune`. Or the create a new category but that seems a bit extreme for just one action.

mwaggett: yeah, I'm not loving either of those. we can think about it some more and wait to see if anyone else has any good ideas.

lib/puppetserver/ca/action/prune.rb
mwaggett: serial number seems right to me

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,123 @@
...
+        end
+
+        # Output for debug
+        #  - Name of issuer
+        #  - Reason for revocation
+        #  - Time of revocation
+        def debug_output(curr_crl, curr_revoked_cert)
+          issuer_name = curr_crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)
+          reason = curr_revoked_cert.extensions.join(', ') # Potentially ugly format.  Need more testing
+          time = curr_revoked_cert.time
```
mwaggett, line 103: I don't think we need these. The cert will have already been revoked, so that info was probably logged elsewhere at the time of revocation. This is just removing duplicate entries from the CRL, so the issuer and cert subject are probably all you need for debug output?

Tu2607: I tried to get the name of the cert but couldn't find a way to do so.  Apparently the `X509::Revoked` class does not have a getter for the cert name.  If we need the cert name, I will need to come up with a way to get that.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,123 @@
...
+          # Getting the CRL(s)
+          loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
+
+          crl_list = loader.crls
+          prune_per_issuer(crl_list)  # Prune the list
+          return 0                  # Place holder return value for now
+        end
+
+        # Further note: The class Revoked override the == method.  We could make
+        # use of this by using array.uniq method to dedup ?
```
mwaggett, line 78: Not sure what you're referring to here..

Tu2607: Ooh, the documentation for the class X509::Revoked has the method `==` override for comparison between 2 Revoked object. Spent sometime researching and found out that the Array class method `uniq` utilize the `==` method so I thought it could be another way of pruning the array.  Just some note I left in case performance aren't good 😄

justinstoller: that'd be a pretty easy implementation ( :

Tu2607: I'll spend sometime digging around on that.  I'm still learning the behind the scene of Ruby :)

lib/puppetserver/ca/action/prune.rb
justinstoller: fwiw, the equivalent implementation with a Hash would look something like this:
```rb
hash[serial] = revoked unless hash[serial]
```
or
```rb
hash[serial] = revoked unless hash.has_key?(serial)
```
I think there's a different thing to look at though, all of these methods dedupe by keeping the _first_ instance they find.
Which Hash you could also just do:
```rb
hash[serial] = revoked
```
which will end up dropping references to all but the _last_ revoked instance with a given serial.

Do you know, do Revoked instances contain any data where it would matter if we saved the first or the last (or the oldest or the newest) instance?

lib/puppetserver/ca/action/prune.rb
mwaggett: hmm, based on the output you included, we could probably just use the alt names extension for that purpose.

Tu2607: The `X509::Revoked.extensions` method doesn't return any name as far as I have tested.  It returns an array and the only thing in that array was the reason of revocation.  Maybe I've misunderstood what you mean by alt names extension?

mwaggett: oh, I was looking at the alt names from
```
Revoked Certificates:
    tu             (SHA256)  18:52:07:6E:6F:E0:42:40:20:75:B8:80:0F:5D:26:94:0A:22:DC:2A:DD:67:81:90:BA:B3:13:8D:AD:B3:B5:95    alt names: ["DNS:tu"]
    batman         (SHA256)  8B:C6:E1:62:8B:E4:90:69:8D:0D:35:A4:36:CC:D8:EE:4B:74:89:06:80:49:21:F2:AD:AA:F9:63:DF:0D:20:42    alt names: ["DNS:batman"]
    superman       (SHA256)  52:3A:D2:06:26:AA:0C:EE:BA:B5:AB:5D:21:B5:12:4C:91:98:98:9A:23:B2:ED:EE:1C:B2:64:3D:67:39:27:AA    alt names: ["DNS:superman"]
```
but maybe that uses a different object

lib/puppetserver/ca/action/prune.rb
Tu2607: Revoked instances do indeed contain the time of revocation which will definitely be different per instance.  The reason of revocation is another thing that could be different between the oldest and the newest.  There is an `X509::Extension` class which I'm looking into right now to see if I can pull anything out of that.

lib/puppetserver/ca/action/prune.rb
Tu2607: So I got this to work for pruning:
```
revoked_list.uniq! do |rev|
  rev.serial
end
```
But haven't figure out how to slide the debugging function for the logger into that block for only the removed cert.  Any good idea on how to do it?

justinstoller: I forgot about that requirement. I doubt you can log for the removed revocation with that method. Unless you iterated over the list twice to compare what did exist to what no-longer does.

Tu2607: Yeah I got stuck on trying to slip the debug function in, couldn't figure out how to do it without having to output info for the whole array.  I'll stick with the current implementation for now and maybe compare their performance.

Tu2607:
I benchmarked the pruning performance in two different ways. One is through using Ruby built-in library `benchmark` and the other is through the `time` command.  The `ca_crl.pem` file I used to test with consisted of 3 CRLs with each CRL contains a revoked list with 50,000 duplicated certs.  Anyway, here is the output of running `time puppetserver-ca prune --config /path/to/puppet.conf`:
```
Finished pruning CRL of CN=tu
Finished pruning CRL of CN=cat
Finished pruning CRL of CN=dog
Finished pruning your local CRLs

real    0m1.412s
user    0m1.198s
sys     0m0.195s
```
As for benchmarking with Ruby's `benchmark` library, I decided to focus solely on the performance of the pruning function thus I wrapped the function in a block like this:
```
time = Benchmark.measure {
  prune_CRLs(crl_list)
}
puts time
```
The result of that benchmark (in second) is as below:
```
User CPU time: 0.491376   
System CPU time: 0.014048   
Sum of system and user CPU time: 0.505424 
Elapsed real time: (  0.506021)
```

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,130 @@
...
+
+        def prune_CRLs(crl_list)
+          crl_list.each do |list|
+            existed_serial_number = Set.new()
+            revoked_list = list.revoked
+            revoked_list.delete_if do |revoked|
+              if existed_serial_number.add?(revoked.serial)
+                false
+              else
+                debug_output(list, revoked) unless @logger.level < 4
```
justinstoller, line 82: This is a good reason to add methods to the logger like `debug?`

Tu2607: Yeah for a cleaner look.  I'll implement that.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,130 @@
...
+            file.close unless file.nil?
+          end
+        end
+
+        # Output for debug
+        #  - Name of issuer
+        #  - Reason for revocation
+        def debug_output(curr_crl, curr_revoked_cert)
+          issuer_name = curr_crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)
+          extensions = curr_revoked_cert.extensions.join(', ')
```
justinstoller, line 110: Honestly I think the serial number is probably the only really interesting thing that needs to be logged at debug (certname would be nice, but that's hard/slow/unreliable)

justinstoller: I think the debug would be more like: "Removing duplicate of #{serial}" or "Removing duplicate of #{serial} from #{revocation_time}"

justinstoller: If we did include the issuer, it'd be nice to get the issuer name at the top of the calling loop and reuse if for every deduped revocation.

Tu2607: Ooh good idea. That's one way we can include the issuer name without it being slow.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,130 @@
...
+          crl_list = loader.crls
+          prune_CRLs(crl_list)
+          write_crl_to_file(crl_list, puppet.settings[:cacrl])
+
+          @logger.inform("Finished pruning your local CRLs")
+          return 0
+        end
+
+        def prune_CRLs(crl_list)
+          crl_list.each do |list|
```
justinstoller, line 75: So, thinking about this a bit more, we don't want to prune CRLs that we don't own. After having deduplicated a CRL we need to increment its version number and re-sign it before writing it to disk. To sign it we need the private key of the CRL issuer, which will only be for our leaf CA.

Tu2607: So I should implement a check to see if a CRL's owner is us and then only prune those selected CRLs.

mwaggett: yeah, I believe that should just be the first CRL from the `ca_crl.pem` file, but would probably be good to check. But I don't _think_ we need to loop through the rest of the CRLs once we have ours.

justinstoller:
fwiw, I think a CRL with 100,000 _legitimate_ revocations, each one with 15 duplicated revocations (ie a CRL with 1.6 million entries) would be the absolute highest we have to deal with. Typically a CRL becomes prohibitively expensive to load into memory for a server after several hundred thousand entries.

lib/puppetserver/ca/action/prune.rb
Tu2607: My understand on CRL are a bit limited beside looking at Ruby classes.  Is `ca_crl.pem` contains only the CRLS that are ours?  I thought it was that way since it's in the puppetlabs directory on my local computer.

justinstoller: There's a few different scenarios you can run Puppet Server's CA under:
1) We operate a single self-signed certificate for our CA which signs every other puppet cert. (Deprecated, no longer default, but still common with folks upgrading)
2) We create a long lived, self-signed root cert that is stored on the server but not used for anything but signing the shorter lived "child" or "leaf" cert. The root in this sense is "offline" and the leaf is "active". This allows a shorter CA cert for security purposes, you can revoke your existing cert and use the root to create a new one that your agent's will still trust. (Current default out of the box).
3) There's a corporate root CA cert which then has a chain of children, one of which being the Puppet CA. This is common in large enterprises.

So, it's possible for the CRL chain to one or more certs, and when there's only two certs we may "own" the root and have the private key on disk, but that root can be considered "offline" and doesn't need maintenance. The other common case when there's 2+ CRLs is that we only own the leaf CRL and should only act on it.

Sorry, that wasn't spelled out clearly in the ticket.

mwaggett: No, the `ca_crl.pem` file contains a CRL for each certificate in the certificate chain. We allow people to use their own external root CA certificates, so there may be certs, and therefore corresponding CRLs, that we don't own.

This page has more detail, including what to put in each of the files: https://puppet.com/docs/puppet/7/server/intermediate_ca.html

Tu2607: Interesting.  Based on that website from @mwaggett , the first CRL is probably all that I need then (still gonna check for their name tho).  Thanks for explaining it further guys!

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+
+        SUMMARY = "Prune the local CRL on disk to remove any duplicated certificates"
+        BANNER = <<-BANNER
+Usage:
+  puppetserver ca prune [--help]
+  puppetserver ca prune [--config]
+
+Description:
+  Prune the list of revoked certificates of any duplication within it.  This command
+  will only prune the CRLs that was signed by Puppet.
```
mwaggett, line 23:
```suggestion
  will only prune the CRL issued by Puppet's CA cert.
```

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+          # Update 'em
+          update_pruned_CRL(puppet_crl, loader.key)
+          # Write 'em out
+          write_crl_to_file(loader.crls, puppet.settings[:cacrl])
+
+          @logger.inform("Finished pruning Puppet's CRL")
+          return 0
+        end
+
+        def select_puppet_CRL(crl_array,key)
```
mwaggett, line 78: (nit)
```suggestion
        def select_puppet_CRL(crl_array, key)
```

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+          puppet_crl = Array.new
+          crl_array.each do |crl|
+            if crl.verify(key)
+              puppet_crl << crl
+            end
+          end
+          return puppet_crl
```
mwaggett, line 85: You should be able to use the `select` method to do this: https://ruby-doc.org/core-2.7.0/Array.html#method-i-select

We _might_ want to confirm that only one CRL was found, but not totally sure.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+        end
+
+        def prune_CRLs(crl_list)
+          crl_list.each do |list|
```
mwaggett, line 89: (nit) I think it might be clearer if we use `crl` instead of `list` here

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+            existed_serial_number = Set.new()
+            revoked_list = list.revoked
+            puts("Issuer #{list.issuer.to_s(OpenSSL::X509::Name::RFC2253)}") if @logger.debug?
```
mwaggett, line 92: We probably want to say something about pruning here, instead of just randomly logging an issuer 😅

mwaggett: Also should use `@logger.debug` instead of `puts`. And then I don't think you need `@logger.debug?` at all..

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+            @logger.err("Error when write out to file on disk")
+            @logger.err("Error: " + e.message)
+          ensure
+            file.close unless file.nil?
+          end
+        end
+
+        # Output for debug
+        #  - Serial number and time of revocation
+        def debug_output(serial,time)
```
mwaggett, line 129: Not sure this needs to be its own method, since it's only used once and it's literally just logging a message.
It's easier to read the `prune_CRLs` method and see what it's logging if you don't have to scroll to a different method.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+        end
+
+        def update_pruned_CRL(crl_list, pkey)
+          crl_list.each do |crl|
+            crl.version=(crl.version + 1)
+            crl.sign(pkey, OpenSSL::Digest::SHA256.new)
+          end
+        end
+
+        def write_crl_to_file(crl_to_write, file_path)
```
mwaggett, line 113: I'm wondering if there's already a method in this repo for writing out CRLs.. I would imagine there must be.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,149 @@
...
+          end
+        end
+
+        # Output for debug
+        #  - Serial number and time of revocation
+        def debug_output(serial,time)
+          @logger.debug("Remove duplicate of #{serial}, revoked on #{time}\n")
+        end
+
+        def parse(args)
```
mwaggett, line 133: (nit) Might be nice to organize the `parse` and `parser` methods to be next to each other.

lib/puppetserver/ca/logger.rb
```diff
@@ -17,6 +17,12 @@ def level
         @level
       end
 
+      def debug?
```
mwaggett, line 20: I don't think this method is necessary, since the `debug` method already checks for the debug level before logging.

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  let(:stdout) { StringIO.new }
+  let(:stderr) { StringIO.new }
+  let(:logger) { Puppetserver::Ca::Logger.new(:info, stdout, stderr) }
+
+  subject { Puppetserver::Ca::Action::Prune.new(logger) }
+
+  describe 'flags' do
+    it 'take a custom path to the puppet.conf' do
+      result, maybe_code = subject.parse(['--config', '/dev/tcp/example.com'])
```
mwaggett, line 22: (nit) I think it would be clearer to call this `maybe_exit_code` (or even just `exit_code`) - at first I was like, what code?? 😅

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    before do
+      allow_any_instance_of(Puppetserver::Ca::Utils::HttpClient).
+        to receive(:with_connection).and_yield(connection)
+      allow_any_instance_of(Puppetserver::Ca::Utils::HttpClient).
+        to receive(:make_store)
+    end
+
+    it 'errors when validating path to puppet.conf' do
+      code = subject.run({'config' => "fake/faker/puppet.conf"})
```
mwaggett, line 41: (nit)
```suggestion
      exit_code = subject.run({'config' => "fake/faker/puppet.conf"})
```

mwaggett (commented):
This is looking really good!

Most of my comments are nitpicky at this point, but we probably also want to add some unit tests for the functions that do the actual pruning.

lib/puppetserver/ca/action/prune.rb
Tu2607: That's one of the thing that I was a bit unsure about.  When I prune with the `verbose` flag, the Puppet Root CA also shown up so there were 2 CRL that was found.  However, Puppet Root CA revoked list is empty.  I guess I could implement another check for empty revoked list.

lib/puppetserver/ca/logger.rb
Tu2607: Yeah. I implemented because of how I used to implement debug mode for this command.  I can remove it.

spec/puppetserver/ca/action/prune_spec.rb
Tu2607: Lol this was a copy and paste from the other test file.

lib/puppetserver/ca/action/prune.rb
Tu2607: Nvm, I'm tripping.  That was when I didn't verify the key...

mwaggett: heh yeah, at first I thought it would include the root CA, but then I remembered that that uses a different key.

spec/puppetserver/ca/action/prune_spec.rb
mwaggett: ahh, if we use that elsewhere, then I'm fine leaving it as-is

Tu2607: but I do agree with what you said about `maybe_code`, I was confused too.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,133 @@
...
+          config_path = inputs['config']
+
+          # Validate the config path.
+          if config_path
+            errors = FileSystem.validate_file_paths(config_path)
+            return 1 if Errors.handle_with_usage(@logger, errors)
+          end
+
+          # Validate puppet config setting.
+          settings_overrides = {}
```
mwaggett, line 42: Don't need this now.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,133 @@
...
+
+          @logger.inform("Finished pruning Puppet's CRL")
+          return 0
+        end
+
+        def prune_CRLs(crl_list)
+          crl_list.each do |crl|
+            existed_serial_number = Set.new()
+            revoked_list = crl.revoked
+            @logger.debug("Pruning issuer #{crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)} CRL")
```
mwaggett, line 70:
```suggestion
            @logger.debug("Pruning duplicate entries in CRL for issuer #{crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)}")
```

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    it 'reduces a CRL with duplicate revoked certs down to 1 cert' do
+      baz_key = OpenSSL::PKey::RSA.new(512)
+      baz_cert = create_cert(baz_key, "Bazzup")
+      revoked_cert = create_cert(OpenSSL::PKey::RSA.new(512), 'revoked')
+      baz_crl = create_crl(baz_cert, baz_key, Array.new(5,revoked_cert))
+
+      crl_array = []
+      crl_array << baz_crl
+
+      subject.prune_CRLs(crl_array)
```
mwaggett, line 82:
```suggestion
      subject.prune_CRLs([baz_crl])
```

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        end
+        exit_code = subject.run({ 'config' => puppet_conf })
+        expect(exit_code).to eq(1)
+        expect(stderr.string).to include('Could not parse')
+      end
+    end
+
+    it 'reduces a CRL with duplicate revoked certs down to 1 cert' do
+      baz_key = OpenSSL::PKey::RSA.new(512)
+      baz_cert = create_cert(baz_key, "Bazzup")
```
mwaggett, line 75: might be clearer to call this `ca_key`, `ca_cert`, etc.

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      revoked_cert = create_cert(OpenSSL::PKey::RSA.new(512), 'revoked')
```
mwaggett, line 76: I think you probably want to use the same key as above for this? though it probably doesn't really matter for testing purposes.

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+              certname = $cadir
+          INI
+        end
+        exit_code = subject.run({ 'config' => puppet_conf })
+        expect(exit_code).to eq(1)
+        expect(stderr.string).to include('Could not parse')
+      end
+    end
+
+    it 'reduces a CRL with duplicate revoked certs down to 1 cert' do
```
mwaggett, line 73: I think it would be good to also test a CRL with multiple certs and maybe multiple duplicated certs, since the resulting CRL won't always (and probably usually won't) contain just one cert.

spec/puppetserver/ca/action/prune_spec.rb
Tu2607: heh, looks like I still need to learn Ruby more 😄

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,134 @@
...
+            revoked_list = crl.revoked
+            @logger.debug("Pruning duplicate entries in CRL for issuer " \
+              "#{crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)}") if @logger.debug?
+
+            revoked_list.delete_if do |revoked|
+              if existed_serial_number.add?(revoked.serial)
+                false
+              else
+                @logger.debug("Remove duplicate of #{revoked.serial}, " \
+                  "revoked on #{revoked.time}\n") if @logger.debug?
```
Tu2607, line 77: Found out that calling the getter function for 150k times will indeed slow down runtime.  Thus I reimplemented a check for verbose mode status.  That way we don't need to call the getter functions all the time and get rejected because we're not in verbose mode.  The difference in run time was from 0.5 - 1 second.  It's not a huge impact on time, but it's noticeable.

lib/puppetserver/ca/cli.rb
Tu2607: Almost forgot about this.  Any ideas on what I should do for `prune`?

mwaggett: oh yeah, hmm. I think I prefer putting `prune` into `INIT_ACTIONS` over calling out `prune` as an exception in the message for `MAINT_ACTIONS`, but I'm wondering if we should call the `INIT_ACTIONS` something else that would more-obviously include `prune`.. `OFFLINE_ACTIONS`? Not really sure, would love to hear others' opinions on this.

lib/puppetserver/ca/logger.rb
```diff
@@ -17,6 +17,10 @@ def level
...
+        return @level >= LEVELS[:debug] ? true : false
```
mwaggett, line 21:
```suggestion
        return @level >= LEVELS[:debug]
```

lib/puppetserver/ca/logger.rb
```diff
@@ -17,6 +17,10 @@ def level
...
+      end
+
       def debug(text)
         if @level >= LEVELS[:debug]
```
mwaggett, line 25:
```suggestion
        if debug?
```

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,134 @@
...
+          # Getting the CRL(s)
+          loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
+
+          # Take 'em
+          puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
+          # Prune 'em
+          prune_CRLs(puppet_crl)
+          # Update 'em
+          update_pruned_CRL(puppet_crl, loader.key)
+          # Write 'em out
```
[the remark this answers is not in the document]
Tu2607: Yeah, I used comment to map out my steps.  Gonna removes these when I finalize the code.

lib/puppetserver/ca/logger.rb
Tu2607: Facepalm moment haha.

lib/puppetserver/ca/action/prune.rb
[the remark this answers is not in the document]
Tu2607: Also, the getter functions are basically the `revoked.time` and `revoked.serial`.

lib/puppetserver/ca/cli.rb
Tu2607: I'm absolutely fine with moving it into `INIT_ACTIONS`.

lib/puppetserver/ca/action/prune.rb
Tu2607: It's the `time` and `serial` function.  They are essentially getter functions for the X509::Revoked class.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -0,0 +1,117 @@
...
+            existed_serial_number = Set.new()
+            revoked_list = crl.revoked
+            @logger.debug("Pruning duplicate entries in CRL for issuer " \
+              "#{crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)}") if @logger.debug?
+
+            revoked_list.delete_if do |revoked|
+              if existed_serial_number.add?(revoked.serial)
+                false
+              else
+                @logger.debug("Remove duplicate of #{revoked.serial}, " \
```
mwaggett, line 73:
```suggestion
                @logger.debug("Removing duplicate of #{revoked.serial}, " \
```

lib/puppetserver/ca/action/prune.rb
Magisus: I think the time is valuable to have, so you could get an understanding of when duplicates are appearing.

lib/puppetserver/ca/cli.rb
```diff
@@ -28,8 +29,9 @@ class Cli
       INIT_ACTIONS = {
```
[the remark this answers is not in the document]
Magisus: How about "Administration", particularly in the help text below.

lib/puppetserver/ca/action/prune.rb
mwaggett: fair

mwaggett (approved):
woo this looks really good!
bidoof

Do you wanna squash into logical commits? Let me know if you need help with any of that, including determining which changes warrant their own commit.

PR #85, opened 2021-07-27
(SERVER-2740) Fix how CRL number is updated

Tu2607:
This commit contains a fix to the prune action so that the CRL number is properly updated after pruning.  Before, the code would update the protocol version which is incorrect.

### DIFF
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -50,12 +50,12 @@ BANNER
           loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
 
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          update_flag = prune_CRLs(puppet_crl)
+          number_of_removed_duplicates = prune_CRLs(puppet_crl)
 
-          if update_flag
+          if number_of_removed_duplicates > 0
             update_pruned_CRL(puppet_crl, loader.key)
             FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
-            @logger.inform("Finished pruning Puppet's CRL.")
+            @logger.inform("Removed #{number_of_removed_duplicates} duplicated certs from Puppet's CRL.")
           else
             @logger.inform("No duplicate revocations found in the CRL.")
           end
@@ -64,7 +64,7 @@ BANNER
         end
 
         def prune_CRLs(crl_list)
-          update_flag = false
+          number_of_removed_duplicates = 0
 
           crl_list.each do |crl|
             existed_serial_number = Set.new()
@@ -76,7 +76,7 @@ BANNER
               if existed_serial_number.add?(revoked.serial)
                 false
               else
-                update_flag = true
+                number_of_removed_duplicates += 1
                 @logger.debug("Removing duplicate of #{revoked.serial}, " \
                   "revoked on #{revoked.time}\n") if @logger.debug?
                 true
@@ -85,14 +85,15 @@ BANNER
             crl.revoked=(revoked_list)
           end
 
-          return update_flag
+          return number_of_removed_duplicates
         end
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            number_ext, other_ext = crl.extensions.partition{ |ext| ext.oid == "crlNumber"}
+            number_ext, other_ext = crl.extensions.partition{ |ext| ext.oid == "crlNumber" }
             number_ext.each do |crl_number|
-              crl_number.value=(OpenSSL::ASN1::Integer(crl_number.value.to_i + 1))
+              updated_crl_number = OpenSSL::BN.new(crl_number.value) + OpenSSL::BN.new(1)
+              crl_number.value=(OpenSSL::ASN1::Integer(updated_crl_number))
             end
             crl.extensions=(number_ext + other_ext)
             crl.sign(pkey, OpenSSL::Digest::SHA256.new)
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -80,7 +80,12 @@ BANNER
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            crl.version=(crl.version + 1)
+            extension_key = crl.extensions.select { |ext| ext.oid != "crlNumber" }
+            extension_number = crl.extensions.select { |ext| ext.oid == "crlNumber" }
+            extension_number.each do |crl_number|
+              crl_number.value=((crl_number.value.to_i + 1).to_s)
+            end
+            crl.extensions=(extension_number + extension_key)
             crl.sign(pkey, OpenSSL::Digest::SHA256.new)
           end
         end
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -50,15 +50,22 @@ BANNER
           loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
 
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          prune_CRLs(puppet_crl)
-          update_pruned_CRL(puppet_crl, loader.key)
-          FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+          update_flag = prune_CRLs(puppet_crl)
 
-          @logger.inform("Finished pruning Puppet's CRL")
+          if update_flag
+            update_pruned_CRL(puppet_crl, loader.key)
+            FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+            @logger.inform("Finished pruning Puppet's CRL.")
+          else
+            @logger.inform("No duplicate revocations found in the CRL.")
+          end
+
           return 0
         end
 
         def prune_CRLs(crl_list)
+          update_flag = false
+
           crl_list.each do |crl|
             existed_serial_number = Set.new()
             revoked_list = crl.revoked
@@ -69,6 +76,7 @@ BANNER
               if existed_serial_number.add?(revoked.serial)
                 false
               else
+                update_flag = true
                 @logger.debug("Removing duplicate of #{revoked.serial}, " \
                   "revoked on #{revoked.time}\n") if @logger.debug?
                 true
@@ -76,16 +84,17 @@ BANNER
             end
             crl.revoked=(revoked_list)
           end
+
+          return update_flag
         end
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            extension_key = crl.extensions.select { |ext| ext.oid != "crlNumber" }
-            extension_number = crl.extensions.select { |ext| ext.oid == "crlNumber" }
-            extension_number.each do |crl_number|
-              crl_number.value=((crl_number.value.to_i + 1).to_s)
+            number_ext, other_ext = crl.extensions.partition{ |ext| ext.oid == "crlNumber"}
+            number_ext.each do |crl_number|
+              crl_number.value=(OpenSSL::ASN1::Integer(crl_number.value.to_i + 1))
             end
-            crl.extensions=(extension_number + extension_key)
+            crl.extensions=(number_ext + other_ext)
             crl.sign(pkey, OpenSSL::Digest::SHA256.new)
           end
         end
--- a/spec/puppetserver/ca/action/prune_spec.rb
+++ b/spec/puppetserver/ca/action/prune_spec.rb
@@ -76,8 +76,9 @@ RSpec.describe Puppetserver::Ca::Action::Prune do
       revoked_cert = create_cert(ca_key, 'revoked')
       ca_crl = create_crl(ca_cert, ca_key, Array.new(5, revoked_cert))
 
-      subject.prune_CRLs([ca_crl])
+      number_of_removed_duplicates = subject.prune_CRLs([ca_crl])
       expect(ca_crl.revoked.length).to eq(1)
+      expect(number_of_removed_duplicates).to eq(4)
     end
 
     it 'deduplicates a CRL with multiple certs that have duplicate of themselves' do
@@ -89,8 +90,9 @@ RSpec.describe Puppetserver::Ca::Action::Prune do
 
       ca_crl = create_crl(ca_cert, ca_key, first_cert + second_cert + third_cert)
 
-      subject.prune_CRLs([ca_crl])
+      number_of_removed_duplicates = subject.prune_CRLs([ca_crl])
       expect(ca_crl.revoked.length).to eq(3)
+      expect(number_of_removed_duplicates).to eq(15)
     end
   end
 
--- a/spec/puppetserver/ca/action/prune_spec.rb
+++ b/spec/puppetserver/ca/action/prune_spec.rb
@@ -104,7 +104,7 @@ RSpec.describe Puppetserver::Ca::Action::Prune do
 
       extensions = ca_crl.extensions.select { |ext| ext.oid == "crlNumber"}
       extensions.each do |ext|
-        expect(ext.value).to match("1")
+        expect(ext.value).to eq("1")
       end
     end
   end
--- a/spec/puppetserver/ca/action/prune_spec.rb
+++ b/spec/puppetserver/ca/action/prune_spec.rb
@@ -95,13 +95,17 @@ RSpec.describe Puppetserver::Ca::Action::Prune do
   end
 
   describe 'update' do
-    it 'bumps CRL version up by 1' do
+    it 'bumps CRL number up by 1' do
       ca_key = OpenSSL::PKey::RSA.new(512)
       ca_cert = create_cert(ca_key, "Bazzup")
       ca_crl = create_crl(ca_cert, ca_key)
 
       subject.update_pruned_CRL([ca_crl], ca_key)
-      expect(ca_crl.version).to eq(2)
+
+      extensions = ca_crl.extensions.select { |ext| ext.oid == "crlNumber"}
+      extensions.each do |ext|
+        expect(ext.value).to match("1")
+      end
     end
   end
 end


spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -95,13 +95,21 @@
...
-    it 'bumps CRL version up by 1' do
+    it 'bumps CRL number up by 1' do
       ca_key = OpenSSL::PKey::RSA.new(512)
       ca_cert = create_cert(ca_key, "Bazzup")
       ca_crl = create_crl(ca_cert, ca_key)
 
       subject.update_pruned_CRL([ca_crl], ca_key)
-      expect(ca_crl.version).to eq(2)
+
+      # crlNumber is based on the amount of certs in revoked list
```
mwaggett, line 105: I don't think that's true? It gets updated every time the CRL is updated, so _could_ happen to be the same as the number of revoked certs, but I think we allow batch revocations and also this pruning would mean that the CRL Number and number of revoked certs are not aligned.

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -95,13 +95,21 @@
...
+      # in our case, that number is 0. So crlNumber is 0.
+      ca_crl.extensions.each do |ext|
+        if ext.oid == "crlNumber"
+          ext.value=((ext.value.to_i + 1).to_s)
```
Magisus, line 109: I don't think you want to set this here. Shouldn't this already be getting updated to 1 when you call `updated_pruned_CRL` above?

mwaggett, line 109: This is the same thing you're doing in `update_pruned_CRL`, so not sure how this is testing that the method does the right thing.. In fact, if we're starting with 0 as the CRL number, running `update_pruned_CRL`, and then running this, I would expect the result to be 2???

lib/puppetserver/ca/action/prune.rb
```diff
@@ -80,7 +80,11 @@ def prune_CRLs(crl_list)
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            crl.version=(crl.version + 1)
+            crl.extensions.each do |ext|
```
mwaggett, line 83: You should be able to use `select` rather than iterating over each extension.

spec/puppetserver/ca/action/prune_spec.rb
Tu2607: Oh darn copy and paste.

spec/puppetserver/ca/action/prune_spec.rb
Tu2607: Ah, this is because of the util function `create_crl` that I used to build the crl.  That function uses the length of the revoked list to set the crl number and in the case of the spec test, I didn't pass in any array of revoked list so the length is 0.  Sorry if that comment is misleading.

Tu2607:
Got a new error now relate to parsing the extensions
```
at java.base/sun.security.x509.CRLExtensions.parseExtension(CRLExtensions.java:130)
	at java.base/sun.security.x509.CRLExtensions.init(CRLExtensions.java:102)
	at java.base/sun.security.x509.CRLExtensions.<init>(CRLExtensions.java:83)
	at java.base/sun.security.x509.X509CRLImpl.parse(X509CRLImpl.java:1203)
	at java.base/sun.security.x509.X509CRLImpl.<init>(X509CRLImpl.java:137)
	at java.base/sun.security.provider.X509Factory.engineGenerateCRL(X509Factory.java:395)
	at java.base/java.security.cert.CertificateFactory.generateCRL(CertificateFactory.java:513)
	at org.bouncycastle.cert.jcajce.JcaX509CRLConverter.getCRL(Unknown Source)
	at com.puppetlabs.ssl_utils.SSLUtils.pemToCRLs(SSLUtils.java:542)
	at puppetlabs.ssl_utils.core$eval3278$pem__GT_crls__3283$fn__3284.invoke(core.clj:624)
	at puppetlabs.ssl_utils.core$eval3278$pem__GT_crls__3283.invoke(core.clj:619)
	at puppetlabs.ssl_utils.core$eval3326$pem__GT_ca_crl__3331$fn__3332.invoke(core.clj:640)
	at puppetlabs.ssl_utils.core$eval3326$pem__GT_ca_crl__3331.invoke(core.clj:634)
	at puppetlabs.puppetserver.certificate_authority$eval40327$get_certificate_statuses__40332$fn__40336.invoke(certificate_authority.clj:1607)
```

Tu2607:
The openssl output after a prune
```
Certificate Revocation List (CRL):
        Version 2 (0x1)
    Signature Algorithm: sha256WithRSAEncryption
        Issuer: /CN=Puppet CA: TuVu-C02FJ00XML87.local
        Last Update: Jul 12 23:59:48 2021 GMT
        Next Update: Jul 12 23:59:49 2026 GMT
        CRL extensions:
            X509v3 CRL Number: 
                7
            X509v3 Authority Key Identifier: 
                keyid:57:62:1C:98:A2:AE:8C:D5:57:10:80:F2:B4:63:41:5E:B8:C8:8B:E1

Revoked Certificates:
    Serial Number: 03
        Revocation Date: Jun 29 19:20:17 2021 GMT
        CRL entry extensions:
            X509v3 CRL Reason Code: 
                Key Compromise
    Serial Number: 04
        Revocation Date: Jul  9 20:40:34 2021 GMT
        CRL entry extensions:
            X509v3 CRL Reason Code: 
                Key Compromise
    Serial Number: 05
        Revocation Date: Jul  9 22:16:47 2021 GMT
        CRL entry extensions:
            X509v3 CRL Reason Code: 
                Key Compromise
    Serial Number: 06
        Revocation Date: Jul 12 23:59:43 2021 GMT
        CRL entry extensions:
            X509v3 CRL Reason Code: 
                Key Compromise
    Signature Algorithm: sha256WithRSAEncryption
         95:09:a9:51:71:dc:24:07:39:01:97:ba:fe:cb:69:b7:46:f2:
         4d:9c:2d:e1:e8:cb:e5:e3:32:10:1d:7d:18:97:1b:1f:14:b2:
         46:d2:66:75:3c:ec:13:8f:c2:78:87:26:cf:ec:db:7c:b6:27:
         43:4f:9b:d1:a8:91:a2:5f:10:a8:4c:12:67:1a:1c:36:ac:62:
         98:87:02:79:3c:2e:ce:b8:ba:e5:24:16:5a:df:00:d5:79:49:
         57:05:ab:0d:fa:3c:f4:09:ef:ff:61:bc:6f:93:f4:c0:85:13:
         b1:30:70:18:26:82:bd:9b:f7:fe:ac:c3:bd:3a:60:c9:5a:a2:
         b6:a8:4c:0e:a7:1f:e9:03:ab:46:4c:ec:5e:08:8a:3e:31:92:
         61:6f:fe:ba:4d:95:48:b1:d3:9e:c7:df:fc:e6:9d:cf:6a:fc:
         e9:0d:02:05:8a:a0:86:67:29:b6:0a:42:b2:ea:1d:fc:47:ee:
         3f:0d:67:0b:8c:a7:9d:af:49:ae:a0:89:2d:78:0c:1a:bf:71:
         d6:ca:c1:8f:7e:bc:ed:11:6d:2e:b6:4a:a7:7c:45:6e:a1:d7:
         45:4c:ae:f2:61:25:0f:58:15:b3:99:1c:5f:e9:d5:5e:6e:0d:
         a5:93:4b:d8:e4:27:e6:a2:06:3e:d4:d5:3c:67:75:ae:67:ee:
         17:eb:8b:c7:83:fc:ef:37:9f:6b:d2:68:ac:cd:23:ed:bd:ad:
         d4:7b:48:3b:4b:fa:ed:0f:c2:f5:37:35:94:ea:7d:4c:8d:9d:
         93:99:75:51:b8:dd:1c:a1:29:a3:18:0e:bf:59:11:b2:8a:3f:
         a5:1f:10:ec:23:76:04:12:bb:08:44:b0:79:b5:40:06:02:83:
         14:31:ee:3a:2e:d6:d4:a1:86:c2:da:f0:40:5f:3e:49:3d:11:
         d6:cd:eb:61:e0:97:8d:fa:35:eb:54:2f:b1:d5:32:ea:75:3b:
         2f:ea:60:03:6c:cb:2a:ce:12:55:3c:a7:83:80:43:8e:7f:91:
         e9:d0:1e:e3:56:bc:fd:1e:cd:d5:95:57:bf:40:b8:46:da:4c:
         6a:c5:64:bf:f4:92:34:3d:54:5f:20:3a:10:2e:04:23:18:35:
         6c:b8:87:11:d9:40:f7:7f:f0:aa:8d:45:52:b0:aa:1c:59:c6:
         cf:36:ef:9a:d4:b6:2d:a4:5e:77:0c:59:8d:55:45:a5:18:54:
         bd:84:e9:e8:9d:2d:75:5c:eb:0a:42:9f:25:eb:76:84:d2:5e:
         07:a4:43:66:9e:fe:d1:11:3d:2d:8d:24:f3:e4:51:58:65:70:
         85:06:93:11:a6:50:d8:ad:f1:b6:0f:6e:13:49:ab:ca:c8:89:
         81:ef:3e:a1:19:26:ae:d4
```
The only thing that remain constant after a prune is the Authority Key Identifier

lib/puppetserver/ca/action/prune.rb
justinstoller: Not to get into golfing territory but [partition](https://ruby-doc.org/core-2.5.1/Enumerable.html#method-i-partition) is what you want.

```
irb(main):001:0> a = [ 1, 2, 3, 4, 5, 6]
irb(main):003:0> even, odd = a.partition(&:even?)
irb(main):004:0> even
=> [2, 4, 6]
irb(main):005:0> odd
=> [1, 3, 5]
```

lib/puppetserver/ca/action/prune.rb
```diff
@@ -80,7 +80,12 @@ def prune_CRLs(crl_list)
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            crl.version=(crl.version + 1)
+            extension_key = crl.extensions.select { |ext| ext.oid != "crlNumber" }
+            extension_number = crl.extensions.select { |ext| ext.oid == "crlNumber" }
```
justinstoller, line 84: I'd write this as `number_extension` or `number_ext` instead of `extension_number`, I'd also call the other array, `other_extensions` since the kind and quantity of other extensions isn't important to this function.

justinstoller:
I'm seeing those failures locally as well. I'm not sure what's causing them though. I'm wondering if there's a bug in OpenSSL and we need to be calling `CRL#add_extension` instead of `CRL#extensions=` (it would seem like the bytesize of an extension isn't being computed correctly but OpenSSL can infer the correct length).

Tu2607:
> I'm seeing those failures locally as well. I'm not sure what's causing them though. I'm wondering if there's a bug in OpenSSL and we need to be calling `CRL#add_extension` instead of `CRL#extensions=` (it would seem like the bytesize of an extension isn't being computed correctly but OpenSSL can infer the correct length).

When I switched to using `partition` instead of `select`, somehow those failures went away.  Not really sure what byte was loss.  And wouldn't `CRL#add_extension` create a duplicate of the same `CRL::Extensions` with the OID being `crlNumber`?   From what I interpret from the doc, `CRL#extensions=` delete the existing extensions and use the array of `CRL::Extensions` that I supply to repopulate the CRL's extensions.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,15 +50,22 @@ def run(inputs)
           loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
 
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          prune_CRLs(puppet_crl)
-          update_pruned_CRL(puppet_crl, loader.key)
-          FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+          update_flag = prune_CRLs(puppet_crl)
+
+          if update_flag == 1
```
justinstoller, line 55: just make this a boolean

lib/puppetserver/ca/action/prune.rb
```diff
@@ -69,18 +76,25 @@ def prune_CRLs(crl_list)
...
+
+          return update_flag
         end
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            crl.version=(crl.version + 1)
+            number_ext, other_ext = crl.extensions.partition{ |ext| ext.oid == "crlNumber"}
+            number_ext.each do |crl_number|
+              crl_number.value=(OpenSSL::ASN1::Integer(crl_number.value.to_i + 1))
```
justinstoller, line 95: I assume this is was fixed the issue you were seeing?

Tu2607: Yeah. Apparently I lose out some bytes when I used `select` as compared to `partition` which is pretty interesting.  Also, not sure if using `OpenSSL::ASN1::Integer` really helped since the `CRL#value=` try to convert the parameter to ASN1 format anywat.

justinstoller: iirc, prior to this you were casting to a string via something like
```ruby
(extension.value.to_i + 1).to_s
```
maybe that was it?

Tu2607: Yep. Switch back to that and lost 4 bytes and the error also came back.  Those conversions might have drop something valuable.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,15 +50,22 @@ def run(inputs)
...
 
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          prune_CRLs(puppet_crl)
-          update_pruned_CRL(puppet_crl, loader.key)
-          FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+          update_flag = prune_CRLs(puppet_crl)
+
+          if update_flag == true
+            update_pruned_CRL(puppet_crl, loader.key)
+            FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0640)
```
mwaggett, line 57: Is there a reason you changed the file permissions here? I'm not sure off the top of my head whether 'other' needs read permissions.. If not, then this change is good.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,15 +50,22 @@ def run(inputs)
...
+            @logger.inform("Finished pruning Puppet's CRL.")
+          else
+            @logger.inform("No pruning needed.")
```
mwaggett, line 60: This might be a little vague without context. Maybe something like
```suggestion
            @logger.inform("No duplicate revocations found in Puppet's CRL.")
```

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -95,13 +95,17 @@
...
       ca_key = OpenSSL::PKey::RSA.new(512)
       ca_cert = create_cert(ca_key, "Bazzup")
       ca_crl = create_crl(ca_cert, ca_key)
 
       subject.update_pruned_CRL([ca_crl], ca_key)
-      expect(ca_crl.version).to eq(2)
+
+      extensions = ca_crl.extensions.select { |ext| ext.oid == "crlNumber"}
+      extensions.each do |ext|
+        expect(ext.value).to match("1")
```
mwaggett, line 107: It looks like `match` is typically used when comparing against a regex: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/match-matcher
We might just want `eq` for this?

lib/puppetserver/ca/action/prune.rb
Tu2607: Ah, I was testing out why the error out parsing was happening and did some test with how the file was written out. I forgot to change it back to 0644.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,15 +50,24 @@ def run(inputs)
           loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
 
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          prune_CRLs(puppet_crl)
-          update_pruned_CRL(puppet_crl, loader.key)
-          FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+          update_flag, dedupe_counter = prune_CRLs(puppet_crl)
+
+          if update_flag
```
mwaggett, line 55: You should be able to get rid of the `update_flag` now and just check whether `dedupe_counter` is greater than 0.

mwaggett: I also might call it something like `number_of_duplicate_revocations` or something.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,15 +50,24 @@ def run(inputs)
...
+            update_pruned_CRL(puppet_crl, loader.key)
+            FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+            @logger.inform("Removed #{dedupe_counter} duplicated certs from Puppet's CRL.")
+            @logger.inform("Finished pruning Puppet's CRL.")
```
mwaggett, line 59: Probably don't need this second log message now?

Tu2607: Yeah I'm fine with removing it

lib/puppetserver/ca/action/prune.rb
```diff
@@ -69,18 +76,26 @@ def prune_CRLs(crl_list)
...
+
+          return number_of_removed_duplicates
         end
 
         def update_pruned_CRL(crl_list, pkey)
           crl_list.each do |crl|
-            crl.version=(crl.version + 1)
+            number_ext, other_ext = crl.extensions.partition{ |ext| ext.oid == "crlNumber" }
+            number_ext.each do |crl_number|
+              updated_crl_number = OpenSSL::BN.new(crl_number.value) + OpenSSL::BN.new(1)
```
mwaggett, line 95: Why do we need `OpenSSL::BN`s now?

Tu2607: It was one those thing that the doc wasn't being really clear about what `OpenSSL::ASN1::Integer` take for parameter. I found out that `OpenSSL::ASN1::Integer()` take a `OpenSSL::BN` object so I figured I stay as close as I can to the doc and used `OpenSSL::BN`.  I was following how we generated the `crlNumber` in the utils file and may have went over the edge 😅

mwaggett: yeah, the docs are definitely confusing. I think this is fine, but based on https://ruby-doc.org/stdlib-2.7.0/libdoc/openssl/rdoc/OpenSSL/ASN1.html it looks like just `OpenSSL::ASN1::Integer.new(number)` would work.

Tu2607: Oh yeah. The main reason create the `updated_crl_number` variable is because of how the `extension.value` function return a string.  Luckily `OpenSSL::BN.new` can accept a string so that why I opted to use `OpenSSL::BN` so that it is uniform everywhere and doesn't cause lossage (is that a word?) of bytes.

mwaggett: ahh, gotcha! that seems like a good approach to me, thanks for explaining!

mwaggett (approved):

PR #88, opened 2021-08-13
(SERVER-3060) Improve speed when querying for CSR only

Tu2607:
This commits modified the `puppetserver ca list` command to only walk the directory of certificate requests when the user
want to get certificate requests. This should improve the speed when running the command `puppetserver ca list` without any option flags.

### DIFF
--- a/lib/puppetserver/ca/action/list.rb
+++ b/lib/puppetserver/ca/action/list.rb
@@ -62,6 +62,7 @@ Options:
           certnames = input['certname'] || []
           all = input['all']
           output_format = input['format'] || "text"
+          missing = []
 
           unless VALID_FORMAT.include?(output_format)
             Errors.handle_with_usage(@logger, ["Unknown format flag '#{output_format}'. Valid formats are '#{VALID_FORMAT.join("', '")}'."])
@@ -87,14 +88,14 @@ Options:
             filter_names = lambda { |x| true }
           end
 
-          all_certs = get_all_certs(puppet.settings).select { |cert| filter_names.call(cert) }
-          requested, signed, revoked = separate_certs(all_certs)
-          missing = certnames - all_certs.map { |cert| cert['name'] }
-
           if (all || certnames.any?)
+            all_certs = get_certs_or_csrs(puppet.settings).select { |cert| filter_names.call(cert) }
+            requested, signed, revoked = separate_certs(all_certs)
+            missing = certnames - all_certs.map { |cert| cert['name'] }
             output_certs_by_state(all, output_format, requested, signed, revoked, missing)
           else
-            output_certs_by_state(all, output_format, requested)
+            all_csrs = get_certs_or_csrs(puppet.settings, "requested")
+            output_certs_by_state(all, output_format, all_csrs)
           end
 
           return missing.any? ? 1 : 0
@@ -209,8 +210,9 @@ Options:
           return requested, signed, revoked
         end
 
-        def get_all_certs(settings)
-          result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses
+        def get_certs_or_csrs(settings, queried_state = nil)
+          query = queried_state ? { :state => queried_state } : {}
+          result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses(query)
 
           if result
             return JSON.parse(result.body)
--- a/lib/puppetserver/ca/certificate_authority.rb
+++ b/lib/puppetserver/ca/certificate_authority.rb
@@ -41,8 +41,8 @@ module Puppetserver
       end
 
       # Returns a URI-like wrapper around CA specific urls
-      def make_ca_url(resource_type = nil, certname = nil)
-        HttpClient::URL.new('https', @ca_server, @ca_port, 'puppet-ca', 'v1', resource_type, certname)
+      def make_ca_url(resource_type = nil, certname = nil, query = {})
+        HttpClient::URL.new('https', @ca_server, @ca_port, 'puppet-ca', 'v1', resource_type, certname, query)
       end
 
       def process_ttl_input(ttl)
@@ -250,8 +250,8 @@ module Puppetserver
       end
 
       # Returns nil for errors, else the result of the GET request
-      def get_certificate_statuses
-        result = get('certificate_statuses', 'any_key')
+      def get_certificate_statuses(query = {})
+        result = get('certificate_statuses', 'any_key', query)
 
         unless result.code == '200'
           @logger.err 'Error:'
@@ -287,8 +287,8 @@ module Puppetserver
       # @param resource_type [String] the resource type of url
       # @param resource_name [String] the resource name of url
       # @return [Struct] an instance of the Result struct with :code, :body
-      def get(resource_type, resource_name)
-        url = make_ca_url(resource_type, resource_name)
+      def get(resource_type, resource_name, query = {})
+        url = make_ca_url(resource_type, resource_name, query)
         @client.with_connection(url) do |connection|
           connection.get(url)
         end
--- a/lib/puppetserver/ca/utils/http_client.rb
+++ b/lib/puppetserver/ca/utils/http_client.rb
@@ -1,5 +1,6 @@
 require 'net/https'
 require 'openssl'
+require 'uri'
 
 require 'puppetserver/ca/errors'
 
@@ -114,7 +115,6 @@ module Puppetserver
             request.body = body
             result = @conn.request(request)
 
-
             Result.new(result.code, result.body)
           end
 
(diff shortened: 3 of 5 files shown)


lib/puppetserver/ca/action/list.rb
```diff
@@ -209,8 +210,8 @@ def separate_certs(all_certs)
           return requested, signed, revoked
         end
 
-        def get_all_certs(settings)
-          result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses
+        def get_all_certs(settings, queried_state = nil)
```
justinstoller, line 213: seems like a weird name now that you can make it explicitly not get all certs.

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -136,10 +136,15 @@ def delete(url_override = nil, headers = {})
...
         URL = Struct.new(:protocol, :host, :port,
                          :endpoint, :version,
-                         :resource_type, :resource_name) do
+                         :resource_type, :resource_name, :queried_state) do
                 def full_url
-                  protocol + '://' + host + ':' + port + '/' +
-                  [endpoint, version, resource_type, resource_name].join('/')
+                  if queried_state
+                    return protocol + '://' + host + ':' + port + '/' +
+                         [endpoint, version, resource_type, resource_name].join('/') + "?state=" + queried_state
```
justinstoller, line 143: Is the `state` parameter something that all endpoints accept?

If not I don't think we should be hardcoding like this. I know of other http apis that let you pass a Hash of query parameters and then convert that into the appropriate string. Something like that might work if `state` isn't a universal parameter to the CA.

Tu2607: I took inspiration from this `https://localhost:8140/puppet-ca/v1/certificate_statuses/fake?state=signed` since it was how I use the certificate status endpoint.

lib/puppetserver/ca/action/list.rb
```diff
@@ -87,14 +87,15 @@ def run(input)
...
-
           if (all || certnames.any?)
+            all_certs = get_all_certs(puppet.settings).select { |cert| filter_names.call(cert) }
+            requested, signed, revoked = separate_certs(all_certs)
+            missing = certnames - all_certs.map { |cert| cert['name'] }
             output_certs_by_state(all, output_format, requested, signed, revoked, missing)
           else
-            output_certs_by_state(all, output_format, requested)
+            all_csrs = get_all_certs(puppet.settings, "requested")
+            missing = []
```
justinstoller, line 97: is this extra

lib/puppetserver/ca/utils/http_client.rb
mwaggett: I think @justinstoller is right - this is a generic URL constructor, so probably shouldn't have `state` hardcoded. I think it would make more sense to accept arbitrary query params.

lib/puppetserver/ca/action/list.rb
Tu2607: Sort of, it's so that the `return missing.any?` doesn't return a 1 since we're not asking for a specific cert.

lib/puppetserver/ca/action/list.rb
```diff
@@ -209,8 +210,14 @@ def separate_certs(all_certs)
...
         end
 
-        def get_all_certs(settings)
-          result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses
+        def get_all_certs_or_csr(settings, queried_state = nil)
+          if queried_state
+            query = { :state => queried_state }
+          else
+            query = {}
+          end
```
Tu2607, line 218: Interested in feedback on making this if/else into a single line through the use of a ternary operator.

lib/puppetserver/ca/certificate_authority.rb
```diff
@@ -1,4 +1,5 @@
 require 'json'
+require 'uri'
```
justinstoller, line 2: I don't see why this is needed here?

Tu2607: ah I was testing out where to put the query hash and forgot to remove that.  Thanks for the catch!

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -136,10 +137,16 @@ def delete(url_override = nil, headers = {})
...
                 def full_url
-                  protocol + '://' + host + ':' + port + '/' +
-                  [endpoint, version, resource_type, resource_name].join('/')
+                  if query.empty?
+                    return  protocol + '://' + host + ':' + port + '/' +
+                         [endpoint, version, resource_type, resource_name].join('/')
+                  else
+                    return protocol + '://' + host + ':' + port + '/' +
+                         [endpoint, version, resource_type, resource_name].join('/') +
+                         "?" + URI.encode_www_form(query)
```
justinstoller, line 148: Can we conditionally add the '?' and the params only if params isn't empty?

lib/puppetserver/ca/action/list.rb
justinstoller:
```rb
query = queried_state ? { state: queried_state } : {}
```
Seems fine, this might be a little more idiomatic? Not sure..
```rb
query = {}
query[:state] = queried_state if queried_state
```

lib/puppetserver/ca/utils/http_client.rb
Tu2607: Isn't the `if query.empty?` already check for whether params aren't empty in this if/else block?

mwaggett: yeah, I think it does..
Thought this could be simplified to put the base url into a variable and then conditionally append the query part if it's not empty.

lib/puppetserver/ca/action/list.rb
```diff
@@ -209,8 +210,9 @@ def separate_certs(all_certs)
           return requested, signed, revoked
         end
 
-        def get_all_certs(settings)
-          result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses
+        def get_all_certs_or_csr(settings, queried_state = nil)
```
mwaggett, line 213: `all` still seems a little weird, since you can filter by state. maybe just `get_certs_or_csrs`?

lib/puppetserver/ca/action/list.rb
mwaggett: yeah I kinda like the latter one better, but don't feel super strongly.

lib/puppetserver/ca/utils/http_client.rb
justinstoller: yes, what molly said is what I meant, sorry.

lib/puppetserver/ca/action/list.rb
```diff
@@ -209,8 +210,9 @@ def separate_certs(all_certs)
           return requested, signed, revoked
         end
 
-        def get_all_certs(settings)
-          result = Puppetserver::Ca::CertificateAuthority.new(@logger, settings).get_certificate_statuses
+        def get_certs_or_csrs(settings, queried_state = nil)
+          query = queried_state ? { :state => "requested" } : {}
```
mwaggett, line 214:
```suggestion
          query = queried_state ? { :state => queried_state } : {}
```

justinstoller (approved):

mwaggett (commented):
Did we want to write any tests to make sure we actually get back just CSRs when running the CLI command or do we think manual testing is sufficient for this?

Otherwise, this looks good!

Tu2607:
In term of manual testing, it's working as intended and the time to retrieve the CSRs is much faster than before.  As for the tests, I'm pretty sure that the test `logs requested certs` in our test suite cover the case of getting back just CSRs.

mwaggett:
sorry, which test are you referring to by `logs requested certs`? I agree that we probably don't need extensive testing, since most of this functionality is tested in the puppetserver repo, but wanted to check whether we think it's worthwhile to just have a test of this behavior from the CLI. but maybe there's already a test for this functionality in this repo so we don't need to add anything?

Tu2607:
> sorry, which test are you referring to by `logs requested certs`? I agree that we probably don't need extensive testing, since most of this functionality is tested in the puppetserver repo, but wanted to check whether we think it's worthwhile to just have a test of this behavior from the CLI. but maybe there's already a test for this functionality in this repo so we don't need to add anything?

Sorry I wasn't being clear. It's a test case in `list_spec.rb` that basically test the `puppetserver-ca list` command without any arguments.

mwaggett:
ahh gotcha, thanks! yeah, that looks sufficient to me 👍🏼

mwaggett (approved):

PR #90, opened 2021-08-18
(maint) Bug fixes and update for `prune` command

Tu2607:
This PR includes:
- Bug fixes related to the recently added functionality of accepting a query parameter when constructing URL.
- Update output information for `prune` command to include the number of certificates found in Puppet's CRL.
- Update `prune` command to only take Puppet's CA CRL, no longer taking a list.

### DIFF
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -50,7 +50,8 @@ BANNER
           loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
 
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          number_of_removed_duplicates = prune_CRLs(puppet_crl)
+          number_of_removed_duplicates, number_of_certificates = prune_CRLs(puppet_crl)
+          @logger.inform("Total number of certificates found in Puppet's CRL is: #{number_of_certificates}.")
 
           if number_of_removed_duplicates > 0
             update_pruned_CRL(puppet_crl, loader.key)
@@ -65,10 +66,12 @@ BANNER
 
         def prune_CRLs(crl_list)
           number_of_removed_duplicates = 0
+          number_of_certificates = 0
 
           crl_list.each do |crl|
             existed_serial_number = Set.new()
             revoked_list = crl.revoked
+            number_of_certificates += revoked_list.length
             @logger.debug("Pruning duplicate entries in CRL for issuer " \
               "#{crl.issuer.to_s(OpenSSL::X509::Name::RFC2253)}") if @logger.debug?
 
@@ -85,7 +88,7 @@ BANNER
             crl.revoked=(revoked_list)
           end
 
-          return number_of_removed_duplicates
+          return number_of_removed_duplicates, number_of_certificates
         end
 
         def update_pruned_CRL(crl_list, pkey)
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -31,6 +31,7 @@ BANNER
 
         def run(inputs)
           config_path = inputs['config']
+          exit_code = 0
 
           # Validate the config path.
           if config_path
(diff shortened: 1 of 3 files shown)


lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -176,7 +176,7 @@ def make_store(bundle, crl_usage, crls = nil)
         # Returns true if it receives back a response of "running", and false if
         # no connection can be made, or a different response is received.
         def self.check_server_online(settings, logger)
-          status_url = URL.new('https', settings[:ca_server], settings[:ca_port], 'status', 'v1', 'simple', 'ca')
+          status_url = URL.new('https', settings[:ca_server], settings[:ca_port], 'status', 'v1', 'simple', 'ca', {})
```
mwaggett, line 179: What exactly was the error you were seeing that lead to this? I think `:query` should be an optional param, so you don't need this, but we may need to update the URL constructor to check if `query` is nil, in addition to checking if it's empty.

Tu2607: Since this line called `URL.new` directly instead of going through the `make_ca_url` function in `certificate_authority.rb`, the `:query` is automatically set to `nil` if you call `URL.new` without passing in a hashmap.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -65,10 +66,12 @@ def run(inputs)
 
         def prune_CRLs(crl_list)
           number_of_removed_duplicates = 0
+          number_of_certificates = 0
 
           crl_list.each do |crl|
```
mwaggett, line 71: So this is potentially looping through multiple CRLs, right? I think we might want to do the logging within this method, rather than in `run`, so that we can log separately for each CRL we prune - I'm not sure it's valuable to add together all the CRL entries from all CRLs.

mwaggett: hmm though I guess this does mirror what we do for the duplicates.. Maybe it makes sense to adjust `prune_CRLs` to just prune a single CRL?

Tu2607: I believe that `prune_CRLs` only operate on Puppet's CRL which there should be only one (please correct me if I am wrong here) so I opted to do it the way that I did.

Tu2607: The variable `puppet_crl` should be Puppet's CRL and that should be a single CRL.

lib/puppetserver/ca/utils/http_client.rb
mwaggett: right, that's why I think we should adjust the check in `full_url` to be 
```
url = url + "?" + URI.encode_www_form(query) unless (query.nil? || query.empty?)
```
(not positive if the parentheses are needed, but added for clarity)

lib/puppetserver/ca/action/prune.rb
mwaggett: The way we're calling it, yes, we only pass a list of one CRL, but this method currently allows for a list of CRLs, right? Since we don't need that functionality, we could just update `prune_CRLs` to be `prune_crl` singular and just take a singular CRL.

lib/puppetserver/ca/utils/http_client.rb
Tu2607: That is a better solution! Thanks Molly!

lib/puppetserver/ca/action/prune.rb
Tu2607: I see what you meant, I think I had that thought in mind when I was coding before and since `select` return an array even if it's just one item hence why I went with a "safe" option of using a list of CRLs just in case that there might be more than one Puppet's CRL (even though that should not happen).

mwaggett: yeah, we could add a check that the returned array from `select` only contains one element and then just use `first` to pass the single CRL.

Tu2607: Since now we're just pruning one CRL, do you think it's a good idea to log the number of certificate in `prune_CRL`? I sort of reserved debug logging for `prune_CRL`.

mwaggett: I think it's probably okay to keep it in `run` if we'll definitely only be operating on a single CRL.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,40 +50,46 @@ def run(inputs)
...
           puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          number_of_removed_duplicates = prune_CRLs(puppet_crl)
 
-          if number_of_removed_duplicates > 0
-            update_pruned_CRL(puppet_crl, loader.key)
-            FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
-            @logger.inform("Removed #{number_of_removed_duplicates} duplicated certs from Puppet's CRL.")
+          if puppet_crl.length == 1
+            number_of_removed_duplicates = prune_CRL(puppet_crl.first)
+            @logger.inform("Total number of certificates found in Puppet's CRL is: #{puppet_crl.first.revoked.length}.")
```
Tu2607, line 56: Curious about the `puppet_crl.first.revoked.length`, too much?

lib/puppetserver/ca/action/prune.rb
```diff
@@ -50,40 +50,46 @@ def run(inputs)
...
+              FileSystem.write_file(puppet.settings[:cacrl], loader.crls, 0644)
+              @logger.inform("Removed #{number_of_removed_duplicates} duplicated certs from Puppet's CRL.")
+            else
+              @logger.inform("No duplicate revocations found in the CRL.")
+            end
+
+            return 0
           else
-            @logger.inform("No duplicate revocations found in the CRL.")
+            @logger.err("Multiple Puppet's CRL were found. Aborting process.")
```
mwaggett, line 68: Might want to make this more generic, in case we find no CRLs. Something like
> Could not identify Puppet's CRL, aborting prune action.

lib/puppetserver/ca/action/prune.rb
mwaggett: I don't think it's too much, but if you wanted, you could change `puppet_crl` to something like `verified_crls` and then set a new variable `puppet_crl = verified_crls.first`, since you pass that to `prune_CRL` anyway.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -49,55 +50,59 @@ def run(inputs)
           # Getting the CRL(s)
           loader = X509Loader.new(puppet.settings[:cacert], puppet.settings[:cakey], puppet.settings[:cacrl])
 
-          puppet_crl = loader.crls.select { |crl| crl.verify(loader.key) }
-          number_of_removed_duplicates = prune_CRLs(puppet_crl)
+          verified_crl = loader.crls.select { |crl| crl.verify(loader.key) }
```
mwaggett, line 53: (nit) I would pluralize this, just so it's clear that this returns an array
```suggestion
          verified_crls = loader.crls.select { |crl| crl.verify(loader.key) }
```

mwaggett (approved):
Have you done any manual testing?

Tu2607:
Yes, spent the last hour or two testing manually with a small crl.  Gonna benchmark speed again.

mwaggett (approved):
fabulous bulldog

PR #94, opened 2021-09-01
(SERVER-2842) Ensures `generate --ca-client` success when puppetserve…

Tu2607:
…r is offline

This commits ensures that `puppetserver ca generate --ca-client` will not fail due to an error being raised by adding a `--force` flag that suppress the error. The new `--force` flag is meant to be only used with `--ca-client`.

### DIFF
--- a/lib/puppetserver/ca/action/generate.rb
+++ b/lib/puppetserver/ca/action/generate.rb
@@ -26,7 +26,7 @@ Usage:
   puppetserver ca generate [--help]
   puppetserver ca generate --certname NAME[,NAME] [--config PATH]
                            [--subject-alt-names NAME[,NAME]]
-                           [--ca-client]
+                           [--ca-client] [--force]
 
 Description:
   Generates a new certificate signed by the intermediate CA
@@ -74,6 +74,10 @@ BANNER
                     'Whether this cert will be used to request CA actions.',
                     'Causes the cert to be generated offline.') do |ca_client|
               parsed['ca-client'] = true
+            end
+            opts.on('--force', 'Suppress errors when signing cert offline.',
+                    "To be used with '--ca-client'") do |force|
+              parsed['force'] = true
             end
             opts.on('--ttl TTL', 'The time-to-live for each cert generated and signed') do |ttl|
               parsed['ttl'] = ttl
@@ -140,7 +144,18 @@ BANNER
           # Generate and save certs and associated keys
           if input['ca-client']
             # Refused to generate certs offfline if the CA service is running
-            return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+            begin
+              return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+            rescue Puppetserver::Ca::ConnectionFailed => e
+              if input['force']
+                @logger.inform("Could not determine that Puppet Server is offline, " \
+                  "connection check failed with error: #{e.wrapped}.\nContinuing with certificate signing.")
+              else
+                @logger.inform("Could not determine if Puppet Server is online. Please ensure it is offline, " \
+                  "then run this command again with the '--force' flag")
+                raise e
+              end
+            end
             all_passed = generate_authorized_certs(certnames, alt_names, puppet.settings, signer.digest)
           else
             all_passed = generate_certs(certnames, alt_names, puppet.settings, signer.digest, input['ttl'])
--- a/spec/puppetserver/ca/action/generate_spec.rb
+++ b/spec/puppetserver/ca/action/generate_spec.rb
@@ -499,4 +499,43 @@ RSpec.describe Puppetserver::Ca::Action::Generate do
       end
     end
   end
+
+  describe "force flag" do
+    it 'errors when generate offline without --force flag when error is not CONNREFUSED' do
+
+      Dir.mktmpdir do |tmpdir|
+        with_temp_dirs tmpdir do |config|
+          allow(subject).to receive(:generate_authorized_certs) { true }
+          allow(httpclient).to receive(:check_server_online).and_raise(
+            Puppetserver::Ca::ConnectionFailed.create(
+              StandardError.new,
+                'UnknownException'))
+
+          expect { subject.run({'certname' => ['foo'],
+                              'config' => config,
+                              'subject-alt-names' => '',
+                              'ca-client' => true})}.to raise_error(Puppetserver::Ca::ConnectionFailed)
+        end
+      end
+    end
+
+    it 'signs when generate offline whilst there is a ConnectionFailed error' do
+      Dir.mktmpdir do |tmpdir|
+        with_temp_dirs tmpdir do |config|
+          allow(subject).to receive(:generate_authorized_certs) { true }
+          allow(httpclient).to receive(:check_server_online).and_raise(
+            Puppetserver::Ca::ConnectionFailed.create(
+             StandardError.new,
+                'UnknownException'))
+
+          expect { subject.run({'certname' => ['foo'],
+                                'config' => config,
+                                'subject-alt-names' => '',
+                                'ca-client' => true,
+                                'force' => true})}.not_to raise_error
+          expect(stdout.string).to include("Continuing with certificate signing")
+        end
+      end
+    end
+  end
 end


lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -175,7 +175,7 @@ def make_store(bundle, crl_usage, crls = nil)
         # Queries the simple status endpoint for the status of the CA service.
         # Returns true if it receives back a response of "running", and false if
         # no connection can be made, or a different response is received.
-        def self.check_server_online(settings, logger)
+        def self.check_server_online(settings, logger, suppress_error = nil)
```
Magisus, line 178: Can we default this to `false`, since we expect it to be a boolean?

Tu2607: yeah, definitely!

Magisus:
@Sharpie will this work for you?

Magisus:
If someone passed `force`, do we want to do the offline check at all? Or should we skip it?

If we do the check, I think it'd be valuable to add some logging, that we couldn't determine if the server was offline, but will be executing anyway because the `--force` flag was passed.

Tu2607:
I think still need to ensure that the server is offline even if we're passing `--force` just so that in case the server is online, there would be a text output telling users that the server is online.  And yeah, I'll add in logging at debug level?

Magisus:
Ah I see the other logging we have, yeah that makes sense. I would log the error suppression at info, I think, particularly if we can report what error we did get besides "Connection Refused". Seems useful for auditing and diagnosing the issue. But open to other opinions.

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -196,7 +196,11 @@ def self.check_server_online(settings, logger)
             if e.wrapped.is_a? Errno::ECONNREFUSED
               return false
             else
-              raise e
+              if suppress_error
+                logger.inform("Logging a suppressed error: #{e.wrapped}")
```
Magisus, line 200: Perhaps
```
Could not determine that puppetserver is offline, connection check failed with #{e.wrapped}. Proceeding with cert signing.
```

Tu2607: Aren't we looking for lunch spot, why are you reviewing my PR  🤣

lib/puppetserver/ca/action/generate.rb
```diff
@@ -78,6 +78,10 @@ def self.parser(parsed = {})
             opts.on('--ttl TTL', 'The time-to-live for each cert generated and signed') do |ttl|
               parsed['ttl'] = ttl
             end
+            opts.on('--force', 'Suppress error when signing cert offline.',
+                    'To be used with ca-client') do |force|
+              parsed['force'] = true
+            end
```
Magisus, line 84: Can we move this block up right under `ca-client`, since they're tied together?

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -196,7 +196,12 @@ def self.check_server_online(settings, logger)
             if e.wrapped.is_a? Errno::ECONNREFUSED
               return false
             else
-              raise e
+              if suppress_error
+                logger.inform("Could not determine that puppetserver is offline, " \
+                  "connection check failed with #{e.wrapped}. Proceeding with cert signing.")
```
mwaggett, line 201: Since this check is used by various actions, I don't think this log message belongs here. It might make more sense for the caller to rescue this error if `--force` is used.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -78,6 +78,10 @@ def self.parser(parsed = {})
             opts.on('--ttl TTL', 'The time-to-live for each cert generated and signed') do |ttl|
               parsed['ttl'] = ttl
             end
+            opts.on('--force', 'Suppress error when signing cert offline.',
+                    'To be used with ca-client') do |force|
```
mwaggett, line 82: What do we want the behavior to be if `--force` is passed without `--ca-client`? The options I see are:
- just ignore the `--force` flag and proceed with the action
- log that `--force` is being ignored and proceed with the action
- fail fast and log that `--force` can't be used without `--ca-client`

mwaggett, line 82:
```suggestion
                    "To be used with '--ca-client'") do |force|
```

Magisus: I was inclined to do the first, since it's not harmful, but am fine with either of the others too 🤷‍♀️

lib/puppetserver/ca/utils/http_client.rb
Magisus: I was actually wondering if we wanted the `force` flag to be usable with those other actions also, but assumed it was out of scope for now. Since currently that flag can only come from the `sign` command, it seemed okay to have the logging be specific to that.

mwaggett: yeah idk it just doesn't seem great to me, conceptually. I would be okay with leaving _some_ logging here, but I think the "Proceeding with cert signing." bit doesn't belong here.

mwaggett: also, I think this conditional needs to `return false`?

lib/puppetserver/ca/action/generate.rb
Tu2607: Right now, it's the first option. The `--force` flag is not really use anywhere outside of the the `run` function. I can definitely add a check.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -75,6 +75,10 @@ def self.parser(parsed = {})
                     'Causes the cert to be generated offline.') do |ca_client|
               parsed['ca-client'] = true
             end
+            opts.on('--force', 'Suppress error when signing cert offline.',
```
justinstoller, line 78: I think it should be `errors` rather than `error`?

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -196,7 +196,13 @@ def self.check_server_online(settings, logger)
             if e.wrapped.is_a? Errno::ECONNREFUSED
               return false
             else
-              raise e
+              if suppress_error
+                logger.inform("Could not determine that puppetserver is offline, " \
```
justinstoller, line 200: I think it should be "Puppet Server" rather than "puppetserver"

lib/puppetserver/ca/action/generate.rb
```diff
@@ -140,7 +144,11 @@ def run(input)
           # Generate and save certs and associated keys
           if input['ca-client']
             # Refused to generate certs offfline if the CA service is running
-            return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+            if HttpClient.check_server_online(puppet.settings, @logger, input['force'])
+              return 1
+            else
+              @logger.inform("Continue with certificate signing.") if input['force']
```
justinstoller, line 150: I think you want "Continuing" rather than "Continue"

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -196,7 +196,13 @@ def self.check_server_online(settings, logger)
             if e.wrapped.is_a? Errno::ECONNREFUSED
               return false
             else
-              raise e
+              if suppress_error
```
mwaggett, line 199: might want to smush this into an `elsif`? I find that easier to read than nested conditionals.

mwaggett (approved):
This generally looks good to me.

Not sure if other folks wanted to weigh in on
- whether `force` is an appropriate name for this flag, given its behavior
- whether doing nothing is the right behavior when passing `--force` without `--ca-client`

lib/puppetserver/ca/action/generate.rb
```diff
@@ -140,7 +144,11 @@ def run(input)
...
           if input['ca-client']
             # Refused to generate certs offfline if the CA service is running
-            return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+            if HttpClient.check_server_online(puppet.settings, @logger, input['force'])
+              return 1
+            else
+              @logger.inform("Continuing with certificate signing.") if input['force']
+            end
             all_passed = generate_authorized_certs(certnames, alt_names, puppet.settings, signer.digest)
           else
```
Tu2607, line 153: Perhaps I could add a logging statement in this `else` block  that says "--force flag is ineffective without --ca-client flag, proceed with signing certificate while Puppet Server is online."

mwaggett: Lemme think out loud about this for a sec.. 

It looks like we only `check_server_online` when `--ca-client` is used, i.e. the `generate` action can be run while the server is online, except for when `--ca-client` is used. 
So in that case, I can't think of what someone would expect from using the `--force` flag without `--ca-client`.. so maybe it is best to leave it as-is, just doing nothing.. 

Does that make sense @Magisus @justinstoller ?

Magisus: Yeah agreed, if we were going to warn about the misused flag, I would do it during args parsing at the very beginning, and not down here.

Magisus:
So here's my thinking:
* `force` is fine as a name, but we should document that it should be used to force special cert signing _when the status of the server cannot be determined_.
* I still like doing nothing when it's passed without the `ca-client` flag.

And my reasoning for both of these things is that people are only going to reach for it when they've been thwarted by the ambiguous error message. Maybe we should catch non `CONNREFUSED` errors and, when the flag hasn't been passed, log that they should check on their server and if it's legitimately not running, run the command again with `force`?

Tu2607:
I believe we're already catching non `CONNREFUSED` by doing `raise e`.  That should stop whatever people are trying to do and inform them of the error.  So now before we `raise e` and halt everything, we're logging a message like "You should check if Puppet Server is running or not then run the command again with --force flag" right?

mwaggett:
@Magisus That all makes sense to me. Good call on docs - @Tu2607 would you mind opening a new ticket for documenting this new flag? I think SERVER is the right project and we'll just add the `DOCS` label.

@Tu2607 I think @Magisus is suggesting that we re-catch the errors in the `run` method and then log and re-throw if `--force` isn't used?

lib/puppetserver/ca/action/generate.rb
Sharpie: Warning or erroring early during argument validation (e.g. "Error: --force must be used with --ca-client") makes sense to me too.

Sharpie (approved):
Looks good to me with `--force` causing the logic to catch and swallow any error,  not just TCP connection refused.

Magisus:
I just think we should log before we ultimately raise, and tell people about force to avoid the issue if they feel confident that it's really okay.

Tu2607:
Don't merge yet.  I'm not too fond with the logging before raising. Currently trying to find a better solution for that.

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -195,7 +195,13 @@ def self.check_server_online(settings, logger)
           rescue Puppetserver::Ca::ConnectionFailed => e
             if e.wrapped.is_a? Errno::ECONNREFUSED
               return false
+            elsif suppress_error
+              logger.inform("Could not determine that Puppet Server is offline, " \
+                "connection check failed with error: #{e.wrapped}.")
+              return false
             else
+              logger.inform("An error has happened. Please check if Puppet Server is running.\n" \
+                "If '--ca-client' flag was used, try again with the '--force' flag to avoid the issue.")
```
Magisus, line 204: Okay I changed my mind, I think this _should_ go in the outer class, so we can know for sure if they used `ca-client` or not.
 As for wording, let's be more specific as to the issue and the remedy, something like:
"Could not determine if Puppet Server is online. Please ensure it is offline, then run this command again with `--force`."

lib/puppetserver/ca/utils/http_client.rb
```diff
@@ -195,6 +195,10 @@ def self.check_server_online(settings, logger)
           rescue Puppetserver::Ca::ConnectionFailed => e
             if e.wrapped.is_a? Errno::ECONNREFUSED
               return false
+            elsif suppress_error
+              logger.inform("Could not determine that Puppet Server is offline, " \
+                "connection check failed with error: #{e.wrapped}.")
+              return false
```
Magisus, line 201: Back to one of Molly's original questions:
Since we're now catching this error out in the generate action, and doing something different if `force` is true vs false, maybe we can move this out there too. As in, leave this whole file alone and do all the special logic in the action. With the extra logic and special logging we've added, I think I now agree that it makes more sense to do at the call site.

lib/puppetserver/ca/action/generate.rb
```diff
@@ -140,7 +144,17 @@ def run(input)
           # Generate and save certs and associated keys
           if input['ca-client']
             # Refused to generate certs offfline if the CA service is running
-            return 1 if HttpClient.check_server_online(puppet.settings, @logger)
+            begin
+              if HttpClient.check_server_online(puppet.settings, @logger, input['force'])
+                return 1
+              else
+                @logger.inform("Continuing with certificate signing.") if input['force']
```
Magisus, line 151: Does this mean we will proceed even if the server is definitely online (`check_server_online` returned false)? I think that's not what we wanted...

lib/puppetserver/ca/action/generate.rb
```diff
@@ -140,7 +144,17 @@ def run(input)
...
+              end
+            rescue Puppetserver::Ca::ConnectionFailed => e
+              @logger.inform("Could not determine if Puppet Server is online. Please ensure it is offline, " \
+                "then run this command again with '--force' flag")
```
mwaggett, line 155: (nit)
```suggestion
                "then run this command again with the '--force' flag.")
```

lib/puppetserver/ca/utils/http_client.rb
Tu2607: Yeah, gonna restore this file to what it was before any changes.  Gonna have to rewrite the test cases as well. Might take a bit longer 😅

mwaggett (approved):
I think this looks good, other than some small wording things that we can clean up in a follow-on PR.

Magisus (approved):

mwaggett:
Followup: https://github.com/puppetlabs/puppetserver-ca-cli/pull/95

PR #96, opened 2021-09-07
Merge 1.x -> main

mwaggett:


lib/puppetserver/ca/action/generate.rb
```diff
@@ -26,7 +26,7 @@ class Generate
   puppetserver ca generate [--help]
   puppetserver ca generate --certname NAME[,NAME] [--config PATH]
                            [--subject-alt-names NAME[,NAME]]
-                           [--ca-client]
+                           [--ca-client] [--force]
```
justinstoller, line 29: fwiw, this should be `[--ca-client [--force]]`

justinstoller (approved):

lib/puppetserver/ca/action/generate.rb
mwaggett: ooh, would you mind opening a PR to 1.x with that? Then I can release again and incorporate that into the mergeup.

justinstoller: np, https://github.com/puppetlabs/puppetserver-ca-cli/pull/97

mwaggett: thank you!

PR #98, opened 2022-03-09
Add failure exit code to puppetserver ca list, SERVER-2797

gcoxmoz:
(SERVER-2797) Add failure exit code when `puppetserver ca list` hits errors.

When running a `puppetserver ca` command on a non-CA server, the command exits with a 0 = ok, when the command has obviously failed.  The failure is easily detected by a human, but less so for a computer.

This seeks to distinguish "failed at looking up" from "found no certificates" and exit accordingly.

### DIFF
--- a/lib/puppetserver/ca/action/list.rb
+++ b/lib/puppetserver/ca/action/list.rb
@@ -92,7 +92,7 @@ Options:
             found_certs = get_certs_or_csrs(puppet.settings)
             if found_certs.nil?
                # nil is different from no certs found
-               @logger.inform('Error while getting certificates')
+               @logger.err('Error while getting certificates')
                return 1
             end
             all_certs = found_certs.select { |cert| filter_names.call(cert) }
@@ -103,7 +103,7 @@ Options:
             all_csrs = get_certs_or_csrs(puppet.settings, "requested")
             if all_csrs.nil?
                # nil is different from no certs found
-               @logger.inform('Error while getting certificates')
+               @logger.err('Error while getting certificate requests')
                return 1
             end
             output_certs_by_state(all, output_format, all_csrs)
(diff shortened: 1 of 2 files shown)


lib/puppetserver/ca/action/list.rb
```diff
@@ -214,9 +225,14 @@ def get_certs_or_csrs(settings, queried_state = nil)
...
 
-          if result
+          if result.nil?
+            # nil returned from an errored lookup
+            return nil
+          elsif result
+            # values from a legit lookup
             return JSON.parse(result.body)
           else
+            # empty result from a legit lookup
```
Magisus, line 235: I think this will never be hit, and I think that's okay. In Ruby, everything besides `nil` or `false` is truthy, so I think all that was needed was to change the `nil` path to return `nil` instead of `empty`, and we don't need the third branch.

I think you should be able to go back to the conditional we had before and just change the `[]` to `nil`.

spec/puppetserver/ca/action/list_spec.rb
```diff
@@ -19,6 +19,20 @@
...
+      expect(err.string).to include('Error while getting certificates')
+    end
+
+    it 'exit-1 when run on a non-CA, certnames' do
+      allow(action).to receive(:get_certs_or_csrs).and_return(nil)
+      exit_code = action.run({'certname' => ['foo','baz']})
+      expect(exit_code).to eq(1)
+      expect(err.string).to include('Error while getting certificates')
+    end
+
```
mwaggett, line 35: Could you add a case for getting just the certificate requests?

gcoxmoz: Submitted.  I am not a ruby person so I only covered what mine introduced.  You had+have an existing coverage gap on that piece of the function, but I'm not qualified to be the fixer.

mwaggett (approved):
This looks great, thanks so much for your work!

PR #100, opened 2022-04-19
(maint) Add snyk action

justinstoller:
(copied from
https://github.com/puppetlabs/puppet/blob/main/.github/workflows/snyk_monitor.yaml)

### DIFF
--- a/.github/workflows/snyk.yaml
+++ b/.github/workflows/snyk.yaml
@@ -1,0 +1,31 @@
+---
+name: Snyk Monitor
+on:
+  push:
+    branches:
+      - main
+jobs:
+  snyk_monitor:
+    if: ${{ github.repository_owner == 'puppetlabs' }}
+    runs-on: ubuntu-latest
+    name: Snyk Monitor
+    steps:
+      - name: Checkout current PR
+        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Magisus (approved):

.github/workflows/snyk.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install dependencies
+        run: bundle install --jobs 3 --retry 3
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+        id: extract_branch
+      - name: Run Snyk to check for vulnerabilities
+        uses: snyk/actions/ruby@master
+        env:
+          SNYK_TOKEN: ${{ secrets.SNYK_FOSS_KEY }}
```
Magisus, line 28: We don't have a `SNYK_FOSS_KEY` in this repo. We have `SNYK_GENERAL_KEY` though, maybe that's what we want?

justinstoller: I checked with security and they said we should use the FOSS key and they have added it to this repo, should be good to go now.

mwaggett (approved):

PR #106, opened 2023-03-01
(PE-34952) Add inventory library

nmburgan:
This adds a module for parsing the inventory.txt file. This will handle any errors found in the file, and return a hash with certname as the key and the rest of the info has another hash.  Because the inventory will likely have old entries for a given certname, this records the serials for the old certnames, but the serial value will be the latest cert found.

Also fixes a couple issues found, mostly around spec files not getting cleaned up due to the way some of the tests were written.


shaigy:
I don't think this is needed for my current PR.  But LGTM.

shaigy (approved):

PR #107, opened 2023-03-02
(PE-34952) Add delete action with --expired flag

nmburgan:
This adds the `puppetserver ca delete` action, which deletes signed certs on disk without revoking them. While all the flags are defined here, only --expired is implemented so far (maybe not ideal, but the others are coming soon). This searches the inventory file for any expired certs, deletes them, then finds any certs that are not in the inventory file, checks if they are expired, and deletes those.

### DIFF
--- a/lib/puppetserver/ca/action/delete.rb
+++ b/lib/puppetserver/ca/action/delete.rb
@@ -1,0 +1,187 @@
+require 'openssl'
+require 'optparse'
+require 'time'
+require 'puppetserver/ca/certificate_authority'
+require 'puppetserver/ca/config/puppet'
+require 'puppetserver/ca/errors'
+require 'puppetserver/ca/utils/cli_parsing'
+require 'puppetserver/ca/utils/file_system'
+require 'puppetserver/ca/utils/inventory'
+require 'puppetserver/ca/x509_loader'
+
+module Puppetserver
+  module Ca
+    module Action
+      class Delete
+
+        include Puppetserver::Ca::Utils
+
+        CERTNAME_BLOCKLIST = %w{--config --expired --revoked --all}
+
+        SUMMARY = 'Delete certificate(s)'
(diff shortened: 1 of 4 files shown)


lib/puppetserver/ca/action/delete.rb
```diff
@@ -0,0 +1,187 @@
...
+              next
+            end
+            begin
+              cert = OpenSSL::X509::Certificate.new(File.read(f))
+            rescue OpenSSL::X509::CertificateError
+              @logger.err("Error reading certificate at #{f}")
+              errored = true
+              next
+            end
+            if cert.not_after < Time.now
```
shaigy, line 176: Should this be converted to utc?

nmburgan: I believe it's converted to a plain `Time` object so it can be directly compared.  I'll double check, though.

nmburgan: Yeah, confirmed, it evaluates it correctly.  In the case below, I created a localized time at the +8 timezone that on its face looks later than `cert.not_after`, but when converted to UTC, is before `cert.not_after`.  It is evaluated correctly.

```
[54] pry(#<Puppetserver::Ca::Action::Delete>)> cert.not_after
=> 2038-02-20 19:11:48 UTC
[55] pry(#<Puppetserver::Ca::Action::Delete>)> t=Time.parse('2038-02-21 00:00:00.0000000 +0800')
=> 2038-02-21 00:00:00 +0800
[56] pry(#<Puppetserver::Ca::Action::Delete>)> t.utc
=> 2038-02-20 16:00:00 UTC
[57] pry(#<Puppetserver::Ca::Action::Delete>)> cert.not_after < t
=> false```

nmburgan:
Justin says 24 is okay, so this should be G2G.

shaigy (approved):

PR #108, opened 2023-03-02
PE-35372 add --remove-expired option to prune command

shaigy:
Add --remove-expired option to prune command which will check for expired certificates in inventory.txt and/or in the signed directory and remove them from CRL.

### DIFF
--- a/lib/puppetserver/ca/action/prune.rb
+++ b/lib/puppetserver/ca/action/prune.rb
@@ -20,11 +20,13 @@ Usage:
   puppetserver ca prune [--help]
   puppetserver ca prune [--config]
   puppetserver ca prune [--config] [--remove-duplicates]
+  puppetserver ca prune [--config] [--remove-expired]
   puppetserver ca prune [--config] [--remove-entries] [--serial NUMBER[,NUMBER]] [--certname NAME[,NAME]]
 
 Description:
   Prune the list of revoked certificates. If no options are provided or
   --remove-duplicates is specified, prune CRL of any duplicate entries.
+  If --remove-expired is specified, remove expired entries from CRL.
   If --remove-entries is specified, remove matching entries provided by
   --serial and/or --certname values. This command will only prune the CRL
   issued by Puppet's CA cert.
@@ -39,6 +41,7 @@ BANNER
         def run(inputs)
           config_path = inputs['config']
           remove_duplicates = inputs['remove-duplicates']
+          remove_expired = inputs['remove-expired']
           remove_entries = inputs['remove-entries']
           serialnumbers = inputs['serial']
           certnames = inputs['certname']
@@ -84,9 +87,15 @@ BANNER
                 number_of_removed_crl_entries += prune_using_certname(puppet_crl, loader.key, inventory_file, cadir, certnames)
               end
             end
-            if (remove_duplicates || (!remove_entries))
+
+            if remove_expired
+              number_of_removed_crl_entries += prune_expired(puppet_crl, loader.key, inventory_file, cadir)
+            end
+
+            if (remove_duplicates || (!remove_entries && !remove_expired))
                number_of_removed_duplicates += prune_CRL(puppet_crl)
             end
+
 
             if (number_of_removed_duplicates > 0 || number_of_removed_crl_entries > 0)
               update_pruned_CRL(puppet_crl, loader.key)
@@ -170,7 +179,7 @@ BANNER
               end
             end
           else
-            @logger.inform "Reading inventory file at #{inventory_file} failed with error #{errors}"
+            @logger.warn "Reading inventory file at #{inventory_file} failed with error #{errors}"
           end
           if certnames
             @logger.debug("Checking CA dir #{cadir} for matching cert names")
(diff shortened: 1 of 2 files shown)


shaigy:
Validated manually by adding expired certs to inventory.txt and signed directory.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -187,6 +196,34 @@ def prune_using_certname(crl, key, inventory_file, cadir, certnames)
...
+        def prune_expired (crl, key, inventory_file, cadir)
+          serialnumbers = []
+          signed_dir = "#{cadir}/signed"
+          @logger.debug("Checking inventory file #{inventory_file} for expired entries") if @logger.debug?
+          errors = FileSystem.validate_file_paths(inventory_file)
+          if errors.empty?
+            File.open(inventory_file).each_line do |line|
+              if line.match(/\/CN=.*$/) && line.split.length == 4
+                not_after = line.split[2]
+                if Time.strptime(not_after,"%Y-%m-%dT%H:%M:%SUTC") && Time.parse(not_after) < Time.now
```
nmburgan, line 208: I thing strptime is part of DateTime, not Time, but I'm also not sure we really need to do this?  What was this check trying to accomplish?

nmburgan: `Time.parse` will throw an exception if the string is not a parseable format, so we may want to handle that here and skip it if it doesn't parse correctly.

lib/puppetserver/ca/action/prune.rb
```diff
@@ -187,6 +196,34 @@ def prune_using_certname(crl, key, inventory_file, cadir, certnames)
...
+                  serialnumbers.push(line.split.first)
+                end
+              end
+            end
+          else
+            @logger.warn "Reading inventory file at #{inventory_file} failed with error #{errors}"
+          end
+          @logger.debug("Checking CA dir #{cadir} for expired certs")
+          Dir.foreach(signed_dir) do |filename|
```
nmburgan, line 217: Might be able to compact this into `Dir.glob("#{signed_dir}/*.pem").each do |filename|`

lib/puppetserver/ca/action/prune.rb
```diff
@@ -187,6 +196,34 @@ def prune_using_certname(crl, key, inventory_file, cadir, certnames)
           end
         end
 
+        def prune_expired (crl, key, inventory_file, cadir)
+          serialnumbers = []
+          signed_dir = "#{cadir}/signed"
+          @logger.debug("Checking inventory file #{inventory_file} for expired entries") if @logger.debug?
+          errors = FileSystem.validate_file_paths(inventory_file)
+          if errors.empty?
+            File.open(inventory_file).each_line do |line|
```
nmburgan, line 205: If you're open to it, perhaps use the Inventory library added recently?  This will parse it for you and return a hash you can act on, handling any errors that arise.

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -147,6 +153,41 @@
       expect(number_of_removed_duplicates).to eq(15)
     end
 
+    it 'reduces a CRL with duplicate revoked certs when called without any arguments' do
+      allow(connection).to receive(:get).and_return(offline)
+      Dir.mktmpdir do |tmpdir|
+        with_files_in tmpdir do |bundle, key, chain, conf|
+          puppet_conf = File.join(tmpdir, 'puppet.conf')
+          File.open puppet_conf, 'w' do |f|
```
nmburgan, line 161: If you like, I think you can use `with_ca_in` and it passes you a `config` that you then pass into the call to `subject.run` that has all the right paths set up.

spec/puppetserver/ca/action/prune_spec.rb
```diff
@@ -183,6 +224,105 @@
...
+      ca_cert = create_cert(ca_key, "You-Shall-Not-Pass")
+      revoked_cert = create_cert(ca_key, 'foo')
+      revoked_cert.not_before = Time.now - (60 * 60 * 24 * 365 * 2)
+      revoked_cert.not_after = Time.now  - (60 * 60 * 24)
+      ca_crl = create_crl(ca_cert, ca_key, [revoked_cert])
+      inventory_entries = ca.inventory_entry(revoked_cert)
+      File.write("#{settings[:cadir]}/signed/foo.pem", revoked_cert)
+      File.write("#{settings[:cadir]}/signed/hello", "hello")
+      File.write("#{settings[:cadir]}/signed/test.txt", "testing")
+      allow(File).to receive(:exist?).with(settings[:cert_inventory]).and_return(false)
```
nmburgan, line 317: Instead of stubbing out a bunch of this stuff, perhaps we can actually write the files?

lib/puppetserver/ca/action/prune.rb
shaigy: strptime was added to make sure it confirms to the correct pattern, otherwise skip it. When I was testing it, probably the line was skipped for some other reason i need to go back and check.

lib/puppetserver/ca/action/prune.rb
shaigy: I read somewhere Dir.foreach is preferred over Dir.glob for performance if there are huge number of files.

shaigy: Here https://stackoverflow.com/questions/2512254/iterate-through-every-file-in-one-directory

lib/puppetserver/ca/action/prune.rb
nmburgan: I think it should be okay not to worry about this.  `Time.parse` can handle a variety of formats, and it results in a `Time` object that doesn't really matter what original form the string was in.  As long as it can parse, it can compare against `Time.now`.

lib/puppetserver/ca/action/prune.rb
nmburgan: Ah, TIL!  Makes sense, though.

lib/puppetserver/ca/action/prune.rb
shaigy: I looked at the library. The serial number check anyway will be done [here.](https://github.com/puppetlabs/puppetserver-ca-cli/pull/108/files#diff-c3becff886d5f2e548cbd5bdf8233fc0156da22cbd17ed00c127baff5c8a98c6R154). With the library it will do that check twice and also the serial number check has to be done against older-serial and new-serial where as it is just a line match here.

lib/puppetserver/ca/action/prune.rb
shaigy: Removed strptime and updated to catch the error. Also updated spec test.

lib/puppetserver/ca/action/prune.rb
shaigy: hmm...that didn't work for me. With that default config, the crl is not getting set correctly. So I am going to leave it as it is.

spec/puppetserver/ca/action/prune_spec.rb
shaigy: hmm...that didn't work for me. With that default config, the crl is not getting set correctly. So I am going to leave it as it is.

spec/puppetserver/ca/action/prune_spec.rb
shaigy: In that case the puppet config needs to be written. So I am going to leave it as it is.

shaigy:
@nmburgan Can you review the changes?

nmburgan (approved):

PR #110, opened 2023-03-08
(PE-34952) Add --revoked action

nmburgan:
This inspects the CRL for a list of revoked serials. It then searches inventory.txt for those serials. If the given serial is the latest one for the particular certname, it puts it on a list for deletion (although it may in fact already be deleted). If it's an old serial for a particular certname, it verifies the cert on disk is a later serial. If for some reason it isn't, it removes the cert on disk as it matches the serial in the CRL. If it isn't in inventory.txt at all, it searches all certs on disk for the given serial. If it's found, it adds that cert to the list for deletion. It then deletes all relevant certs that exist.

### DIFF
--- a/lib/puppetserver/ca/action/delete.rb
+++ b/lib/puppetserver/ca/action/delete.rb
@@ -13,7 +13,6 @@ module Puppetserver
   module Ca
     module Action
       class Delete
-
         include Puppetserver::Ca::Utils
 
         CERTNAME_BLOCKLIST = %w{--config --expired --revoked --all}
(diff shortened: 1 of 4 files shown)


lib/puppetserver/ca/action/delete.rb
```diff
@@ -107,7 +106,70 @@ def run(args)
...
+          if args['revoked']
+            loader = X509Loader.new(settings[:cacert], settings[:cakey], settings[:cacrl])
+            verified_crls = loader.crls.select { |crl| crl.verify(loader.key) }
+            unless verified_crls.length == 1
+              @logger.err("Could not identify Puppet's CRL. Aborting delete action.")
+              return 1
+            end
+            crl = verified_crls.first
+
+            # First, search the inventory for the given serial.
```
shaigy, line 123: Isn't it 'revoked serial' instead of 'given serial'?

nmburgan: Yeah, suppose it is :)

shaigy (approved):

PR #111, opened 2023-03-10
(PE-35382) Add --all action

nmburgan:
This deletes all certs in the signed directory, and must be used by itself.

Additionally, this tweaks the behavior of puppetserver ca delete --expired to not print an error when a cert listed in the inventory is not found on disk, since this will happen when you run the command multiple times and we don't really want people messing with inventory.txt manually to get rid of this error.

### DIFF
--- a/lib/puppetserver/ca/action/delete.rb
+++ b/lib/puppetserver/ca/action/delete.rb
@@ -273,7 +273,7 @@ BANNER
           files.each do |f|
             begin
               s = get_cert_serial(f)
-              return File.basename(f)[0...-4] if s == serial # Remove .pem
+              return File.basename(f, '.pem') if s == serial # Remove .pem
             rescue Exception => e
               @logger.debug("Error reading certificate at #{f} with exception #{e}. Skipping this file.")
             end
--- a/lib/puppetserver/ca/action/delete.rb
+++ b/lib/puppetserver/ca/action/delete.rb
@@ -80,6 +80,10 @@ BANNER
             errors << '  Must pass one of the valid flags to determine which certs to delete'
           end
 
+          if results['all'] && (results['expired'] || results['revoked'] || results['certname'])
+            errors << '  The --all flag must not be used with --expired, --revoked, or --certname'
+          end
+
           errors_were_handled = Errors.handle_with_usage(@logger, errors, parser.help)
 
           exit_code = errors_were_handled ? 1 : nil
@@ -190,6 +194,12 @@ BANNER
           if args['certname']
             count, errored = delete_certs(cadir, args['certname'])
             deleted_count += count
+          end
+
+          if args['all']
+            certnames = Dir.glob("#{cadir}/signed/*.pem").map{ |c| File.basename(c, '.pem') }
+            # Since we don't run this with any other flags, we can set these variables directly
+            deleted_count, errored = delete_certs(cadir, certnames)
           end
 
           plural = deleted_count == 1 ? "" : "s"
--- a/lib/puppetserver/ca/action/delete.rb
+++ b/lib/puppetserver/ca/action/delete.rb
@@ -182,7 +182,9 @@ BANNER
             inventory, err = Inventory.parse_inventory_file(inventory_file_path, @logger)
             errored ||= err
             expired_in_inventory = inventory.select { |k,v| v[:not_after] < Time.now }.map(&:first)
-            count, err = delete_certs(cadir, expired_in_inventory)
+            # Don't print errors if the cert is not found, since the inventory
+            # file can contain old entries that have already been deleted.
+            count, err = delete_certs(cadir, expired_in_inventory, false)
             deleted_count += count
             errored ||= err
             other_certs_to_check = find_certs_not_in_inventory(cadir, inventory.map(&:first))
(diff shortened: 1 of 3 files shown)


lib/puppetserver/ca/action/delete.rb
```diff
@@ -192,6 +198,12 @@ def run(args)
             deleted_count += count
           end
 
+          if args['all']
+            certnames = Dir.glob("#{cadir}/signed/*.pem").map{ |c| File.basename(c, '.pem') }
```
shaigy, line 202: This will walk the directory twice, once for getting filenames, and then in delete_certs to delete the files. Probably not too expensive.

lib/puppetserver/ca/action/delete.rb
```diff
@@ -192,6 +198,12 @@ def run(args)
...
+            # Since we don't run this with any other flags, we can set these variables directly
+            deleted_count, errored = delete_certs(cadir, certnames)
```
shaigy, line 204: Should error_on_not_found be set to false here? Its just walking the directory and deleting them, so I am thinking that error message may be misleading. What do you think?

rileynewton:
FYI, has nothing to do with this PR but `delete` is listed under the wrong action type in the cli doc, should be admin action  if you want to update it real quick here.

```
[root@wan-naturalism ca]# puppetserver ca
Unknown action: 
Usage: puppetserver ca <action> [options]

Manage the Private Key Infrastructure for
Puppet Server's built-in Certificate Authority

Available Actions:

  Certificate Actions (requires a running Puppet Server):

    clean	Revoke cert(s) and remove related files from CA
    delete	Delete certificate(s)
    generate	Generate a new certificate signed by the CA
    list	List certificates and CSRs
    revoke	Revoke certificate(s)
    sign	Sign certificate request(s)

  Administrative Actions (requires Puppet Server to be stopped):

    import	Import an external CA chain and generate server PKI
    setup	Setup a self-signed CA chain for Puppet Server
    enable	Setup infrastructure CRL based on a node inventory.
    migrate	Migrate the existing CA directory to /etc/puppetlabs/puppetserver/ca
    prune	Prune the local CRL on disk to remove certificate entries

```

rileynewton:
Also it looks like the "(default)" part of the line below `--remove-expired             Remove expired  entries from CRL(default)` should be removed if the description above that is accurate.

```
[root@wan-naturalism ~]# puppetserver ca prune --help
Usage:
  puppetserver ca prune [--help]
  puppetserver ca prune [--config]
  puppetserver ca prune [--config] [--remove-duplicates]
  puppetserver ca prune [--config] [--remove-expired]
  puppetserver ca prune [--config] [--remove-entries] [--serial NUMBER[,NUMBER]] [--certname NAME[,NAME]]

Description:
  Prune the list of revoked certificates. If no options are provided or
  --remove-duplicates is specified, prune CRL of any duplicate entries.
  If --remove-expired is specified, remove expired entries from CRL.
  If --remove-entries is specified, remove matching entries provided by
  --serial and/or --certname values. This command will only prune the CRL
  issued by Puppet's CA cert.

Options:
        --help                       Display this command-specific help output
        --config CONF                Path to the puppet.conf file on disk
        --remove-duplicates          Remove duplicate entries from CRL(default)
        --remove-expired             Remove expired  entries from CRL(default)
        --remove-entries             Remove entries from CRL
        --serial NUMBER[,NUMBER]     Serial numbers(s) in HEX to be removed from CRL
        --certname NAME[,NAME]       Name(s) of the cert(s) to be removed from CRL
```

lib/puppetserver/ca/action/delete.rb
nmburgan: It theoretically shouldn't ever pass in a certname it can't find, since it's just picking up files on disk, so I don't think the error will ever actually show in the first place.

lib/puppetserver/ca/action/delete.rb
nmburgan: I don't think it's actually walking the directory inside `delete_certs`.  `delete_certs` assumes the file is there and constructs a file path from the certname.  However, you're right in that it might be slightly more efficient to have the function delete the file as it walks the directory.  But I think it's pretty negligible.

nmburgan:
> FYI, has nothing to do with this PR but `delete` is listed under the wrong action type in the cli doc, should be admin action if you want to update it real quick here.
> 
> ```
> [root@wan-naturalism ca]# puppetserver ca
> Unknown action: 
> Usage: puppetserver ca <action> [options]
> 
> Manage the Private Key Infrastructure for
> Puppet Server's built-in Certificate Authority
> 
> Available Actions:
> 
>   Certificate Actions (requires a running Puppet Server):
> 
>     clean	Revoke cert(s) and remove related files from CA
>     delete	Delete certificate(s)
>     generate	Generate a new certificate signed by the CA
>     list	List certificates and CSRs
>     revoke	Revoke certificate(s)
>     sign	Sign certificate request(s)
> 
>   Administrative Actions (requires Puppet Server to be stopped):
> 
>     import	Import an external CA chain and generate server PKI
>     setup	Setup a self-signed CA chain for Puppet Server
>     enable	Setup infrastructure CRL based on a node inventory.
>     migrate	Migrate the existing CA directory to /etc/puppetlabs/puppetserver/ca
>     prune	Prune the local CRL on disk to remove certificate entries
> ```

Good catch! Fixed, and updated the summary text a bit.

shaigy (approved):

PR #116, opened 2023-12-13
(SERVER-2691) configurable root ca name

h0tw1r3:
Adds command line parameter for the root ca name, and supports spaces in the ca_name and root_ca_name values when parsed from the config file.


justinstoller (approved):

justinstoller:
Thanks, @h0tw1r3 !

We're in the midst of release work (clean up for the recent FOSS release and prep for the upcoming PE release) so this will be released as a gem shortly after that has wrapped up and will be in the product releases after that.

You can, of course, do a manual `puppetserver gem install` to upgrade your puppetserver-ca-cli version without waiting for the next release. Once we've released the gem, of course.

PR #121, opened 2024-06-25
Do not rely on legacy facts

bastelfreak:


lib/puppetserver/ca/config/puppet.rb
```diff
@@ -77,8 +77,9 @@ def load(cli_overrides: {}, logger:, ca_dir_warn: true)
         end
 
         def default_certname
-          hostname = Facter.value(:hostname)
-          domain = Facter.value(:domain)
+          networking = Facter.value(:networking)
+          hostname = networking[:hostname]
```
bastelfreak: I'm not sure if that's actually a symbol or if Facter returns a string, but I think it's all symbols.

lib/puppetserver/ca/config/puppet.rb
```diff
@@ -77,8 +77,9 @@ def load(cli_overrides: {}, logger:, ca_dir_warn: true)
...
+          domain = networking[:domain]
           if domain and domain != ''
             fqdn = [hostname, domain].join('.')
```
bastelfreak, line 84: wouldn't it be easier to always use the networking.fqdn fact? I think when there's no domain, facter will set the fqdn to the hostname?

lib/puppetserver/ca/config/puppet.rb
bastelfreak: all strings:
```
$ bundle exec irb
irb(main):001:0> require 'facter'
=> true
irb(main):002:0> networking = Facter.value(:networking).keys
=> ["domain", "fqdn", "interfaces", "ip6", "mac", "mtu", "netmask", "ip", "network", "netmask6", "primary", "network6", "scope6", "hostname"]
irb(main):003:0>
```

lib/puppetserver/ca/config/puppet.rb
smortex: That seems reasonable.

smortex (approved):
Moving away of legacy facts seems a good move.  As you say, this can probably use the `networking.fqdn` fact directly.

lib/puppetserver/ca/config/puppet.rb
joshcooper: Im assuming this was copied from puppet https://github.com/puppetlabs/puppet/blob/7739378f6fa840d6dbee237132754c64f94cc38a/lib/puppet/settings.rb#L72

joshcooper (approved):

PR #122, opened 2024-08-09
Only ship the minimally required files

ekohl:
This avoids shipping files like .github and .travis.yml.

### DIFF
--- a/puppetserver-ca.gemspec
+++ b/puppetserver-ca.gemspec
@@ -13,11 +13,11 @@ Gem::Specification.new do |spec|
   spec.summary       = %q{A simple CLI tool for interacting with Puppet Server's Certificate Authority}
   spec.homepage      = "https://github.com/puppetlabs/puppetserver-ca-cli/"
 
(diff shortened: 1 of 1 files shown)


puppetserver-ca.gemspec
```diff
@@ -13,11 +13,11 @@ Gem::Specification.new do |spec|
   spec.summary       = %q{A simple CLI tool for interacting with Puppet Server's Certificate Authority}
   spec.homepage      = "https://github.com/puppetlabs/puppetserver-ca-cli/"
 
-  spec.files         = `git ls-files -z`.split("\x0").reject do |f|
-    f.match(%r{^(test|spec|features)/})
+  spec.files         = `git ls-files -z`.split("\x0").select do |f|
+    f.start_with?('LICENSE', 'README.md', 'lib/')
```
smortex, line 17: I think the `exe` directory is missing here :thinking:.

ekohl: I used `gem build` and they show up. There was no need to list it twice (other is via executables)

PR #123, opened 2024-08-09
Stop activating Bundler in puppetserver-ca

ekohl:
This fails on current Bundler versions where require is a private method. It also makes it useless outside of a bundler context.


smortex (approved):

jcharaoui:
Debian has been carrying this very patch for a while: https://salsa.debian.org/puppet-team/ruby-puppetserver-ca-cli/-/blob/debian/main/debian/patches/0003-Disable-Bundler-during-test-execution.patch?ref_type=heads

